PR #2, opened 2017-10-25
Reworked parts of repair_graph to detect modifications

pckroon:
It's been extended to find and identify modifications compared to the reference
graphs. This includes, but is not limited to PTMs and termini.

Note that it's still WIP, but I could do with some review on the entire file.


pckroon:
I think this is ready for review and merging.

martinize2/processors/repair_graph.py
```diff
@@ -80,40 +74,131 @@ def make_reference(mol):
...
+    while missing and added:
+        added = False
+        for ref_idx in missing:
+            # See if the atom we want to add has a known neighbour. Otherwise,
+            # continue to the next.
+            if all(ref_neighbour in missing for ref_neighbour in reference[ref_idx]):
+                continue
+            added = True
+            missing.pop(missing.index(ref_idx))
+            res_idx = max(molecule) + 1  # Alternative: find the first unused number
```
jbarnoud: Finding the first unused number would be dangerous. It could lead to some outdated references leading somewhere when they should not. The current approach is much safer.

martinize2/processors/repair_graph.py
```diff
@@ -131,12 +133,87 @@ def make_reference(mol):
...
+                    molecule.add_edge(neighbour_res_idx, res_idx)
+                    neighbours += 1
+            assert neighbours != 0
+    if missing:
+        for ref_idx in missing:
+            print('Could not reconstruct atom {}{}:{}'.format(reference.nodes[ref_idx]['resname'],
+                  reference.nodes[ref_idx]['resid'], reference.nodes[ref_idx]['atomname']))
+
+
+def repair_graph(molecule, reference_graph):
```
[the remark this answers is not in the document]
jbarnoud: The docstring mentions a copy, but I do not see where it happens. Also, since the "molecule" parameter is a `Molecule`, then the return should be a `Molecule` too rather than just a `Graph`.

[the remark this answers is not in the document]
jbarnoud: And the docstring does not mention the PTMs.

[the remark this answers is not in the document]
jbarnoud: I just found the copy. It seems to be in the processor and not in this function.

martinize2/processors/canonize_PTMs.py
```diff
@@ -0,0 +1,268 @@
...
+                    to_replace = ptm_node.get('replace', dict())
+                    for attr_name, val in to_replace.items():
+                        if attr_name == 'atomname' and val is None:
+                            # DEBUG output
+                            print('Removing node {}, {}'.format(mol_idx), mol_node['atomname'])
+                            mol_node.remove_node(mol_idx)
+                            n_idxs.remove(mol_idx)
+                            break
+                        # DEBUG output
+                        print('Changing attribute {} from {} to {}'.format(attr_name, mol_node[attr_name], val))
```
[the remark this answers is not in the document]
pckroon: At some point these should become logging calls. I'm ok with postponing that a little longer though.

martinize2/processors/repair_graph.py
pckroon: Good point.

jbarnoud:
Could you add the format of a PTM in a docstring? It would help me to write tests which I need to rebase without breaking everything.

martinize2/processors/canonize_PTMs.py
```diff
@@ -0,0 +1,268 @@
...
+    Parameters
+    ----------
+    residue : networkx.Graph
+        The residues involved with these PTMs. Need not be connected.
+
+    residue_ptms : list of tuples of two sets of node indices
+        As returned by ``find_PTM_atoms``, but only those relevant for
+        ``residue``.
+
+    known_PTMs : sequence of tuples of (networkx.Graph, PTMGraphMatcher)
```
pckroon: Is this not yet clear enough?

jbarnoud: I was expecting PTMs to gave more. Aren't them named? Do they only need/use the "atomname" attribute? There is at least the 'replace' attribute to document somewhere. An example would not hurt.

pckroon: Fair enough. Which attributes of the Graph are used is not really described. I'll get on it.

jbarnoud: Thanks. Now I can write PTM graphs for the tests.

jbarnoud:
Can a PTM change the resname of a full residue? This would make things easier for mppings and some block definition. For instance, in the case of histidine protonation, changing the resname to the expected version of histidine makes it trivial to work with down the pipeline.

pckroon:
I'm fairly certain I've thought of that at some point, but I'm not sure it made it into the final code. What happens if there's multiple PTMs for a single residue?
IIRC I currently decorate all atoms in the residue with which PTMs are found for that residue. It might be a better idea to get a separate processor that takes all that info and translates it into near-canonical residuenames.

pckroon:
Note that when you'll try to rebase it the renaming will confuse git to no end. Using `git rebase --strategy-option="find-renames=10"` will help, but there might be a better way.

martinize2/processors/rename_modified_residues.py
```diff
@@ -0,0 +1,62 @@
...
+    ----------
+    mol : Molecule
+        The molecule whose residue names should be changed. Is modified
+        in-place.
+
+    Returns
+    -------
+    None
+    """
+    rename_map_ff = mol.force_field.renamed_residues
```
jbarnoud: The attribute should be added to the `ForceField` class.

martinize2/processors/rename_modified_residues.py
```diff
@@ -0,0 +1,62 @@
...
+        modifications = node.get('modifications', [])
+        resname = node['resname']
+        if not modifications:
+            continue
+        modifications = tuple(sorted(modifications))
+        try:
+            new_name = rename_map[(resname, modifications)]
+        except KeyError:
+            continue
+        node['resname'] = new_name
```
jbarnoud: This should be in a `else`, else `new_name` is not defined when `KeyError` is raised above.

martinize2/processors/rename_modified_residues.py
```diff
@@ -0,0 +1,62 @@
...
+
+from .processor import Processor
+
+
+def rename_modified_residues(mol):
+    """
+    Renames residue names based on the current residue name, and the found
+    modifications. The new names are found in
+    :attr:`force_field.renamed_residues`, which should be a mapping of
+    ``{(rename, [modifications, ...]): new_name}``.
```
jbarnoud: What is `modification`? Is it a name or the graph directly?

martinize2/processors/rename_modified_residues.py
```diff
@@ -0,0 +1,62 @@
...
+    # Sort the list of modifications, so that the order does not matter
+    # This should probably be done as the list is parsed. Also, as we parse it,
+    # make sure we actually recognize all the modification names.
+    for (resname, mods), new_name in rename_map_ff.items():
+        rename_map[resname, tuple(sorted(mods))] = new_name
+
+    for node_key in mol:
+        node = mol.nodes[node_key]
+        modifications = node.get('modifications', [])
+        resname = node['resname']
```
jbarnoud: Is there a reason to enforce a node to have a 'resname' here? Could this be `node.get('resname')`? This would mean that we do not change the name of residues that are not named.

martinize2/processors/rename_modified_residues.py
```diff
@@ -0,0 +1,62 @@
...
+    -------
+    None
+    """
+    rename_map_ff = mol.force_field.renamed_residues
+    rename_map = {}
+    # Sort the list of modifications, so that the order does not matter
+    # This should probably be done as the list is parsed. Also, as we parse it,
+    # make sure we actually recognize all the modification names.
+    for (resname, mods), new_name in rename_map_ff.items():
+        rename_map[resname, tuple(sorted(mods))] = new_name
```
jbarnoud: Depending on the type of the element of `mods` (see my question above), could the tuple be a set instead?

jbarnoud:
I am a bit dreading the rebase. I am not sure I understand every single bit of the PR, and I am not sure it all works as intended which makes it difficult to identify what I may break.

martinize2/processors/repair_graph.py
```diff
@@ -80,40 +74,131 @@ def make_reference(mol):
...
+    match = ref_residue['match']
+
+    resid = ref_residue['resid']
+    resname = ref_residue['resname']
+
+    for ref_idx in reference:
+        if ref_idx in match:
+            res_idx = match[ref_idx]
+            node = molecule.nodes[res_idx]
+            # Copy, because it's references everywhere.
```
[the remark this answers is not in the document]
jbarnoud: Then shouldn't it be "because there are references"?

martinize2/processors/rename_modified_residues.py
jbarnoud: Raahh. Me dumb! It is a continue, not a pass. Would be clearer with a `pass` and a `else`, though; but it is not wrong.

martinize2/processors/rename_modified_residues.py
pckroon: Mistake on my side. It's the modification's name now.

pckroon:
> What happens when only some of the nodes in the residue receive a modification, but not all the nodes from the residue? Couln't that lead to a mismatch in residue name within a residue?

Yes and no. Yes, that is what woud happen. But no, it never will, because the PTM code labels every atom in the respective residues. I could check that that happened here, but that seems needlessly expensive. In addition, I'm not convinced it's such a bad thing that modifications can cause a residue to be cut into pieces.

martinize2/processors/canonize_PTMs.py
```diff
@@ -0,0 +1,278 @@
...
+                        # DEBUG output
+                        print('Changing attribute {} from {} to {}'.format(attr_name, mol_node[attr_name], val))
+                        mol_node[attr_name] = val
+            for n_idx in n_idxs:
+                molecule.nodes[n_idx]['modifications'] = molecule.nodes[n_idx].get('modifications', [])
+                molecule.nodes[n_idx]['modifications'].append(ptm)
+    return None
+
+
+class CanonizePTMs(Processor):
```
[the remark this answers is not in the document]
pckroon: Woopsie...

pckroon:
One more thing about renaming residues. Currently modifications can have a replace attribute for every node, which may replace the resname attribute of only part of the residue. Something to keep in mind.

jbarnoud:
I extended my test from #77 with the PTM to make it work, but I get an error. What am I doing wrong? https://gist.github.com/jbarnoud/16ba64c211c953e6cbcff477ba9765af

pckroon:
I found the problem. The fix is not so easy though.
The problem is that in repair_graph OE1 and OE2 get exchanged (which is fair, because oxygen is oxygen). However, that means that the PTM suddenly needs to look for OE2-H instead of OE1-H. 

If you change the `PTMGraphMatcher` in canonicalize_modifications to this it will at least identify the PTMs properly, but at the moment I daren't give any guarantees about the atomnames further down the line.

```python
class PTMGraphMatcher(nx.isomorphism.GraphMatcher):
    # G1 >= G2; G1 is the found residue; G2 the PTM reference
    def semantic_feasibility(self, node1, node2):
        node1 = self.G1.nodes[node1]
        node2 = self.G2.nodes[node2]
        return node1.get('PTM_atom', False) == node2['PTM_atom'] and node1['element'] == node2['element']
```

Issue #6, opened 2017-10-26
User input


pckroon:
There is no option for user input, other than hardcoded paths.

jbarnoud:
From the list above, the options that are still missing are:
* `-his`
* `-link`
* `-name`
* Multiscaling

pckroon:
Status?

jbarnoud:
From the options in the list just above, only `-cys` was added.

fgrunewald:
this seems like a stale issue. If specific CLI flags are still missing I'd suggest opening a specific issue for each.

PR #41, opened 2018-02-03
[WIP] Secondary structures with DSSP

### LINKED ISSUE #22: Secondary structure dependent modifications of the topology
* [ ] Define how to describe the modifications
* [ ] Be able to apply the modifications
* [ ] Read secondary structures assignment from the user input
* [ ] Get secondary structure assignments by running DSSP

### LINKED ISSUE #43: Links cannot alter an existing atom
Links can add interactions, but they cannot change atom propertied such as the atom type or the charge. Changing the atom type is required to apply secondary structure specific bead types (see issue #22 and PR #41).

jbarnoud:
**This is a work in progress! It depends on #33.**

Add support for the secondary structure dependant topology in martini. Fixes #22. This is a step required in the progression of #6.

* [X] Run DSSP and parse its output
* [X] Annotate a structure with DSSP assignation
* [x] Describe links with secondary structure specificity
* [x] Apply links with secondary structure specificity
* [ ] Handle cyclic proteins
* [ ] Handle proteins with non-standard residues
* [ ] Handle proteins with PTMs

@Tsjerk do you have PDBs that I can use as test cases for cyclic proteins?

EDIT:
Fixes #43


jbarnoud:
Rebased against master after library renaming.

vermouth/processors/do_links.py
```diff
@@ -28,26 +29,79 @@
...
-            return True
+    return True
+
 
+def _pattern_match(molecule, atoms, raw_match):
+    for link_key, template_attr in atoms:
+        try:
+            molecule_key = raw_match[link_key]
+        except KeyError as err:
+            print(raw_match)
```
pckroon: Prettier message?

vermouth/selectors.py
```diff
@@ -0,0 +1,100 @@
...
+# limitations under the License.
+
+import numpy as np
+from .molecule import Molecule
+
+
+PROTEIN_RESIDUES = ('ALA', 'ARG', 'ASP', 'ASN', 'CYS',
+                    'GLU', 'GLN', 'GLY', 'HIS', 'ILE',
+                    'LEU', 'LYS', 'MET', 'PHE', 'PRO',
+                    'SER', 'THR', 'TRP', 'TYR', 'VAL')
```
pckroon: Do we want to move this to a FF specific datafile? I think we do.
Either way, make it a set instead of a tuple.

vermouth/selectors.py
```diff
@@ -0,0 +1,100 @@
...
+        for n_idx in molecule
+    )
+
+
+def select_all(node):
+    return True
+
+
+def select_backbone(node):
+    return node.get('atomname') == 'BB'
```
pckroon: Again, FF specific.

vermouth/selectors.py
```diff
@@ -0,0 +1,100 @@
...
+    position = atom.get('position')
+    return position is not None and np.all(np.isfinite(position))
+
+
+def filter_minimal(molecule, selector):
+    """
+    Create a minimal molecule which only include a selection.
+
+    The minimal molecule only has nodes. It does dot have any molecule level
+    attribute, not does it have edges.
```
pckroon: typo

vermouth/selectors.py
```diff
@@ -0,0 +1,100 @@
...
+    ``True`` for atoms to keep in the selection.
+
+    Parameters
+    ----------
+    molecule: Molecule
+    selector: callback
+
+    Returns
+    -------
+    Molecule
```
pckroon: You could argue that this function should return a list (iterator?) of node keys. That way it's more obvious you don't get edges etc.

vermouth/tests/datafiles.py
```diff
@@ -0,0 +1,38 @@
...
+PDB_PROTEIN = TEST_DATA / '1bta.pdb'
+PDB_PARTIALLY_PROTEIN = TEST_DATA / '1bta_mutated.pdb'  # LYS replaced by UNK
+PDB_NOT_PROTEIN = TEST_DATA / 'heme.pdb'
+
+DSSP_OUTPUT = TEST_DATA / 'dssp_1bta.ssd'
+
+# Clean the namespace so only the data file variables can be imported.
+# An other option would be to define __all__, but it is easy to forget to add
+# a variable in that list.
+del os
```
pckroon: os is already deleted?

vermouth/dssp/dssp.py
```diff
@@ -0,0 +1,397 @@
...
+        Parse a DSSP output.
+    """
+    process = subprocess.Popen(
+        [executable, "-i", "/dev/stdin"],
+        stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stdin=subprocess.PIPE
+    )
+    process.stdin.write(pdb.write_pdb_string(system, conect=False).encode('utf8'))
+    out, err = process.communicate()
```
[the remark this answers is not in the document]
jbarnoud: The input should not be that large. It falls in the "what could possibly go wrong" category. Fixing this has a high effort/benefit ratio, but I keep it in mind.

vermouth/dssp/dssp.py
```diff
@@ -0,0 +1,397 @@
...
+        atom attribute.
+    attribute: str
+        The name of the atom attribute in which to store the annotation.
+
+    See Also
+    --------
+    run_dssp, read_dssp2
+    """
+    if not is_protein(molecule):
+        return molecule
```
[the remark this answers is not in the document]
jbarnoud: I do not think they are. The doc says that non-proteins molecules and enpty molecules are returned unmodified. These lines do exactly that: return the molecule before it is modified.

vermouth/dssp/dssp.py
```diff
@@ -0,0 +1,397 @@
...
+        value = first_node.get(attribute, default)
+        sequence.append(value)
+    return sequence
+
+
+def annotate_residues_from_sequence(molecule, attribute, sequence):
+    residues = list(molecule.iter_residues())
+    if len(sequence) == 1:
+        sequence = sequence * len(residues)
+    elif len(sequence) != len(residues):
```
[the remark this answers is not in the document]
jbarnoud: If you go through the `if`, then `sequence` has the same length as `residues` by definition. Using `elif` avoid extra work and make seman

vermouth/dssp/dssp.py
```diff
@@ -0,0 +1,397 @@
...
+    return result
+
+
+def sequence_from_residues(molecule, attribute, default=None):
+    sequence = []
+    for residue_nodes in molecule.iter_residues():
+        first_name = residue_nodes[0]
+        first_node = molecule.nodes[first_name]
+        value = first_node.get(attribute, default)
+        sequence.append(value)
```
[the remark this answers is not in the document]
jbarnoud: Why not. Note that it becomes a list later on anyway.

vermouth/dssp/dssp.py
```diff
@@ -0,0 +1,397 @@
...
+        for node_name in residue_nodes:
+            node = molecule.nodes[node_name][attribute] = value
+
+
+def convert_dssp_annotation_to_martini(
+        molecule, from_attribute='secstruct', to_attribute='cgsecstruct'):
+    dssp_sequence = sequence_from_residues(molecule, from_attribute)
+    if None not in dssp_sequence:
+        cg_sequence = list(convert_dssp_to_martini(dssp_sequence))
+        annotate_residues_from_sequence(molecule, to_attribute, cg_sequence)
```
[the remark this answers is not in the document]
jbarnoud: Good catch. This could have caused a silent failure.

vermouth/forcefield.py
```diff
@@ -46,9 +50,17 @@ def find_force_fields(directory):
     directory = str(directory)  # Py<3.6 compliance
     for name in os.listdir(directory):
         path = os.path.join(directory, name)
-        if glob(os.path.join(path, '*.rtp')):
-                force_fields[name] = ForceField(path)
+        has_force_field_file = any(True for _ in iter_force_field_files(path))
```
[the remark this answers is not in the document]
jbarnoud: It does not read as well, but it would avoid issues with directories full of files.

vermouth/molecule.py
```diff
@@ -79,15 +89,39 @@ def add_interaction(self, type_, atoms, parameters, meta=None):
...
-            raise KeyError("Can't find interaction of type {} between atoms {}".format(type_, atoms))
+            msg = ("Can't find interaction of type {} between atoms {} "
+                   "and with version {}")
+            raise KeyError(msg.format(type_, atoms, version))
+        del self.interactions[type_][idx]
+
+    def remove_matching_interaction(self, type_, template_interaction):
+        for idx, interaction in enumerate(self.interactions[type_]):
+            if interaction_match(self, interaction, template_interaction):
+                break
```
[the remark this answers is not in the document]
jbarnoud: No particular reason. It was just fitting my mind better when I wrote it. Changed.

vermouth/processors/do_links.py
jbarnoud: This was there for debug purpose. There is no reason to catch the exception. Removed.

vermouth/selectors.py
jbarnoud: It is my plan indeed. But I would like to add arbitrary attribute to force fields, and I have some thinking to do about the API. I'll change that in an other PR if you don't mind. I made it a set, though.

vermouth/selectors.py
jbarnoud: Same as above.

jbarnoud:
@pckroon I just addressed the review comments.

PR #53, opened 2018-02-11
Add support for Martini 2.2 P

### LINKED ISSUE #47: Handle drude particles in polarizable force fields
Polarizable Martini uses dummy beads to carry some charges.

Supporting them should not be too much of a problem, but still requires an effort:
* In the topology, it is just a matter of having the dummy defined in the blocks.
* In the mapping, it may be possible to define the dummies as built from some of the atoms, We need to be careful when choosing those underlying atoms that it leads to the correct CG graph.
* An other option for the mapping would be to add a "[ dummies ]" section in the mapping file to adjust the graph.
* For the coordinates, the dummies cannot be positions by averaging the underlying coordinates as be do for regular beads. We need a processor to reproduce what the current martinize does in its [`add_dummy`](https://github.com/Tsjerk/Martinize/blob/master/martinize/IO.py#L300) function. We can easily have the block label the dummy, and have the processor apply its magic on the nodes that are labelled as dummies.

jbarnoud:
**This PR is based on top of #41. Wait until #41 is merged before merging this one.**

Fixes #47 

Martini 2.2P uses dummy beads to hold the charges. These beads are not
part of the mapping; therefore, they must be added to the mapping target
molecule without having underlying atoms, and there coordinates must be
constructed.
This commit:
* Adds an [ extra ] section in the mapping files to declare unmapped particles
* Adapt do_mapping to add the unmapped particles
* Adds the LocateChargeDummies that construct the coordinates of the
  dummies
* Adds the force field file for martini 2.2P

In addition, I added a [ from ] and a [ to ] sections to the mapping
files as a quick fix to specify what force field the files are for.

The AverageBeads processor can ignore beads that do not have an
underlying graph.

### DIFF
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -311,7 +311,7 @@ def _treat_atom_prefix(reference, attributes):
     if order_attribute < 0:
         prefix_symbol = '-'
     atom_name = reference[prefix_end:]
-    attributes['atomname'] = atom_name
+    attributes['atomname'] = attributes.get('atomname', atom_name)
     prefixed_reference = prefix_symbol * int(math.fabs(order_attribute)) + atom_name
 
     return prefixed_reference, atom_name, attributes
@@ -338,7 +338,6 @@ def _treat_link_interaction_atoms(atoms, context, section):
                                              value, context_atom[key]))
             context_atom.update(attributes)
         else:
-            attributes['atomname'] = atom_name
             context.add_node(prefixed_reference, **attributes)
 
 
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -411,7 +411,7 @@ def _parse_block_atom(tokens, context):
         attributes = {}
 
     # deque does not support slicing
-    first_six = (tokens.popleft() for i in range(6))
+    first_six = (tokens.popleft() for _ in range(6))
     _, atype, _, resname, name, charge_group = first_six
     atom = {
         'atomname': name,
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -406,6 +406,11 @@ def _base_parser(tokens, context, context_type, section, natoms=None, delete=Fal
 
 
 def _parse_block_atom(tokens, context):
+    if tokens[-1].startswith('{'):
+        attributes = _parse_atom_attributes(tokens.pop())
+    else:
+        attributes = {}
+
     # deque does not support slicing
     first_six = (tokens.popleft() for i in range(6))
     _, atype, _, resname, name, charge_group = first_six
@@ -420,7 +425,7 @@ def _parse_block_atom(tokens, context):
         atom['charge'] = float(tokens.popleft())
     if tokens:
         atom['mass'] = float(tokens.popleft())
-    context.add_atom(atom)
+    context.add_atom(dict(collections.ChainMap(attributes, atom)))
 
 
 def _parse_link_atom(tokens, context):
@@ -522,6 +527,7 @@ def read_ff(lines):
         'dihedrals': 4,
         'impropers': 4,
         'constraints': 2,
+        'virtual_sites2': 3,
     }
 
     macros = {}
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -405,15 +405,21 @@ def _base_parser(tokens, context, context_type, section, natoms=None, delete=Fal
         context.interactions[section] = interaction_list
 
 
-def _parse_block_atom(line, context):
-    _, atype, _, resname, name, charge_group, charge = line.split()
+def _parse_block_atom(tokens, context):
+    # deque does not support slicing
+    first_six = (tokens.popleft() for i in range(6))
+    _, atype, _, resname, name, charge_group = first_six
     atom = {
         'atomname': name,
         'atype': atype,
         'resname': resname,
-        'charge': float(charge),
         'charge_group': int(charge_group),
     }
+    # charge and mass are optional, but charge has to be defined for mass to be
+    if tokens:
+        atom['charge'] = float(tokens.popleft())
+    if tokens:
+        atom['mass'] = float(tokens.popleft())
     context.add_atom(atom)
 
 
@@ -563,7 +569,7 @@ def read_ff(lines):
             _parse_link_attribute(tokens, context, section)
         elif section == 'atoms':
             if context_type == 'block':
-                _parse_block_atom(cleaned, context)
+                _parse_block_atom(tokens, context)
             elif context_type == 'link':
                 _parse_link_atom(tokens, context)
         elif section == 'non-edges':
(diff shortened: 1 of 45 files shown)


jbarnoud:
Rebased against master after renaming.

pckroon:
I'll review this once #41 is merged. Otherwise there's a bunch of duplicate work...

jbarnoud:
Rebased against master.

vermouth/processors/locate_charge_dummies.py
```diff
@@ -0,0 +1,121 @@
...
+    points = np.stack([np.zeros((3, )), vector])
+    points -= points / 2
+    return points
+
+
+def find_anchor(molecule, node_key, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
+    dummy = molecule.nodes[node_key]
+    if dummy.get(attribute_tag, None) is None:
+        msg = 'Node "{}" is not a charge dummy. Check the "{}" node attribute.'
+        raise ValueError(msg.format(node_key, attribute_tag))
```
pckroon: A try/except might be better here.

vermouth/processors/locate_charge_dummies.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # There should be only one anchor, and that anchor is the (hopefully) only
+    # neighbor that is not a dummy.
+    potential_anchors = [
+        neighbor
+        for neighbor in molecule.neighbors(node_key)
+        if molecule.nodes[neighbor].get(attribute_tag, None) is None
```
pckroon: `if attribute_tag not in molecule.nodes[neighbor]`

vermouth/processors/locate_charge_dummies.py
```diff
@@ -0,0 +1,121 @@
...
+
+    for dummy_key, position in zip(dummy_keys, points):
+        molecule.nodes[dummy_key]['position'] = position + anchor_position
+
+
+def locate_all_dummies(molecule, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
+    dummies = [
+        (find_anchor(molecule, dummy_key, attribute_tag), dummy_key)
+        for dummy_key in molecule.nodes
+        if molecule.nodes[dummy_key].get(attribute_tag, None) is not None
```
pckroon: `if attribute_tag in ...`

vermouth/processors/locate_charge_dummies.py
```diff
@@ -0,0 +1,121 @@
...
+    ]
+    dummies.sort()  # Sort primarily on the anchor key as it appears first
+
+    grouped_by_anchor = itertools.groupby(dummies, key=operator.itemgetter(0))
```
pckroon: Are you sure you need the `itemgetter`?

vermouth/ffinput.py
```diff
@@ -405,16 +404,27 @@ def _base_parser(tokens, context, context_type, section, natoms=None, delete=Fal
...
+    _, atype, _, resname, name, charge_group = first_six
     atom = {
         'atomname': name,
         'atype': atype,
         'resname': resname,
-        'charge': float(charge),
         'charge_group': int(charge_group),
     }
-    context.add_atom(atom)
+    # charge and mass are optional, but charge has to be defined for mass to be
```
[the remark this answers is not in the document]
jbarnoud: Because charge is column 7 and mass is column 8. Only the 6 first columns are mandatory, but you need a column 7 to have a column 8.

I agree it is weird, but this is something for which I would like to be compatible with the ITP format.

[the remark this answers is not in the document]
pckroon: Ok, I expected something like this. I agree with your reasoning.

vermouth/processors/locate_charge_dummies.py
jbarnoud: I find the if more explicit regarding the logic.

vermouth/processors/locate_charge_dummies.py
jbarnoud: The attribute may be set to `None`.

vermouth/processors/locate_charge_dummies.py
jbarnoud: Same as above.

vermouth/processors/locate_charge_dummies.py
jbarnoud: I could use lambda. Without a key, `groupby` use the full object while I want to only use the first element of a tuple of size 2.

pckroon: That I see. Does using the full tuple produce a qualitatively different result though? I do prefer the itemgetter over a lambda.

vermouth/processors/do_mapping.py
```diff
@@ -45,7 +45,7 @@ class GraphMapping:
     #       of blocks and a mapping of {node_idx: [node_idx, ...], ...}
 
     # TODO: renumber output residues. Needs information about the entire system
-    #       or we need to at least garantue we run it all in order. Which we
+    #       or we need to at least the garanty we run it all in order. Which we
```
pckroon: Still typo'd ;)

guarantee

vermouth/processors/locate_charge_dummies.py
jbarnoud: It should. The list `dummies` has the form` [(anchor0, dummy0), (anchor0, dummy1), (anchor1, dummy2), ...]` and I want to group the dummies by their anchors. Since ll the dummies are different, grouping on the full tuple will keep all the element separated.

vermouth/processors/locate_charge_dummies.py
```diff
@@ -44,6 +61,32 @@ def colinear_pair():
 
 
 def find_anchor(molecule, node_key, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
+    """
+    Find the non-dummy bead to which a charge dummy is anchored.
+
+    Each charge dummy has to be attached to one non-dummy atom. This function
+    returns the node key for that non-dummy atom.
```
pckroon: `... attached to *exactly* one ...`?

vermouth/processors/locate_charge_dummies.py
```diff
@@ -44,6 +61,32 @@ def colinear_pair():
...
+
+    Parameters
+    ----------
+    molecule: vermouth.Molecule
+        The molecule to work on.
```
pckroon: Do you use Molecule API, or could it be just a networkx.Graph?

vermouth/processors/locate_charge_dummies.py
```diff
@@ -67,7 +110,26 @@ def find_anchor(molecule, node_key, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
 
 
 def locate_dummy(molecule, anchor_key, dummy_keys, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
+    """
+    Set the position of a group of charge dummies around anon-dummy anchor.
```
pckroon: missing space

vermouth/processors/do_mapping.py
jbarnoud: I'll blame it on the fact that it is 1AM here.

pckroon:
The typo is still there, but it's good enough for now.

PR #54, opened 2018-02-14
Elastic network

jbarnoud:
A first go at the rubber band elastic network. I have to figure out why I have discrepancies in my distance matrix compared to martinize1.

This is a step forward for #33. It requires #41 and #53 to be merged.

### DIFF
--- a/vermouth/processors/apply_rubber_band.py
+++ b/vermouth/processors/apply_rubber_band.py
@@ -1,0 +1,103 @@
+# Copyright 2018 University of Groningen
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import networkx as nx
+from .processor import Processor
+from .. import selectors
+
+
+def self_distance_matrix(coordinates):
+    return np.sqrt(((coordinates[:, np.newaxis, :] - coordinates[np.newaxis, :, :]) ** 2).sum(axis=-1))
+
+
+def compute_decay(distance_matrix, shift, rate, power):
+    return np.exp(-rate * ((distance_matrix - shift) **  power))
+
+
+def compute_force_constants(distance_matrix, lower_bound, upper_bound,
+                            decay_factor, decay_power, base_constant,
+                            minimum_force):
+    constants = compute_decay(distance_matrix, lower_bound, decay_factor, decay_power)
+    np.fill_diagonal(constants, 0)
+    constants *= base_constant
+    constants[constants < minimum_force] = 0
+    constants[distance_matrix > upper_bound] = 0
+    return constants
+
+
+def apply_rubber_band(molecule, selector,
+                      lower_bound, upper_bound,
+                      decay_factor, decay_power,
+                      base_constant, minimum_force,
+                      bond_type):
+    selection = []
+    coordinates = []
+    missing = []
+    for node_key, attributes in molecule.nodes.items():
+        if selector(attributes):
+            selection.append(node_key)
+            coordinates.append(attributes.get('position'))
+            if coordinates[-1] is None:
+                missing.append(node_key)
+    if missing:
+        raise ValueError('All atoms from the selection must have coordinates. '
+                         'The following atoms do not have some: {}.'
+                         .format(' '.join(missing)))
+    coordinates = np.stack(coordinates)
+    distance_matrix = self_distance_matrix(coordinates)
+    constants = compute_force_constants(distance_matrix, lower_bound,
+                                        upper_bound, decay_factor, decay_power,
+                                        base_constant, minimum_force)
+    connectivity = nx.to_numpy_matrix(molecule, nodelist=selection).astype(bool)
+    constants *= ~connectivity
+    for from_idx, to_idx in zip(*np.triu_indices_from(constants)):
+        from_key = selection[from_idx]
+        to_key = selection[to_idx]
+        force_constant = constants[from_idx, to_idx]
+        length = round(distance_matrix[from_idx, to_idx], 5)
+        if force_constant > minimum_force:
+            molecule.add_interaction(
+                type_='bonds',
+                atoms=(from_key, to_key),
+                parameters=[bond_type, length, force_constant],
+                meta={'group': 'Rubber band'},
+            )
+
+
+class ApplyRubberBand(Processor):
+    def __init__(self, lower_bound, upper_bound, decay_factor, decay_power,
+                 base_constant, minimum_force,
+                 bond_type=6,
+                 selector=selectors.select_backbone):
+        super().__init__()
+        self.lower_bound = lower_bound
+        self.upper_bound = upper_bound
+        self.decay_factor = decay_factor
+        self.decay_power = decay_power
+        self.base_constant = base_constant
+        self.minimum_force = minimum_force
+        self.bond_type = bond_type
+        self.selector = selector
+
+    def run_molecule(self, molecule):
+        apply_rubber_band(molecule, self.selector,
+                          lower_bound=self.lower_bound,
+                          upper_bound=self.upper_bound,
+                          decay_factor=self.decay_factor,
+                          decay_power=self.decay_power,
+                          base_constant=self.base_constant,
+                          minimum_force=self.minimum_force,
+                          bond_type=self.bond_type)
+        return molecule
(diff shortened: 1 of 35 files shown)


jbarnoud:
Rebased against master after renaming.

pckroon:
Also for this one, I'll review it once the other two are merged to prevent duplication of effort.

jbarnoud:
Rebased against master.

jbarnoud:
Rebased (again) against master. I'll be less responsive for 8 hours or so.

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+from .processor import Processor
+from .. import selectors
+
+
+def self_distance_matrix(coordinates):
+    return np.sqrt(((coordinates[:, np.newaxis, :] - coordinates[np.newaxis, :, :]) ** 2).sum(axis=-1))
+
+
+def compute_decay(distance_matrix, shift, rate, power):
+    return np.exp(-rate * ((distance_matrix - shift) **  power))
```
pckroon: I would call it just distance instead of distance_matrix, and let numpy deal with the broadcasting

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def compute_force_constants(distance_matrix, lower_bound, upper_bound,
+                            decay_factor, decay_power, base_constant,
+                            minimum_force):
+    constants = compute_decay(distance_matrix, lower_bound, decay_factor, decay_power)
+    np.fill_diagonal(constants, 0)
```
pckroon: But do note that it *should* be distance_matrix here

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+    constants *= base_constant
+    constants[constants < minimum_force] = 0
+    constants[distance_matrix > upper_bound] = 0
```
pckroon: `constants[(constants < minimun_force) | (distance_matrix > upper_bound)] = 0` is also an option. Yours is more readable. Do you think it's worth profiling?

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+    return constants
+
+
+def build_connectivity_matrix(graph, separation, selection=None):
+    if separation <= 0:
+        raise ValueError('Separation has to be strictly positive.')
+    if separation == 1:
+        return nx.to_numpy_matrix(graph, nodelist=selection).astype(bool)
```
pckroon: Could use a comment

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+    subgraph = graph.subgraph(selection)
+    connectivity = np.zeros((len(subgraph), len(subgraph)), dtype=bool)
+    for (i, key_i),  (j, key_j) in itertools.combinations(enumerate(subgraph.nodes), 2):
+        shortest_path = len(nx.shortest_path(subgraph, key_i, key_j))
```
pckroon: I would've done it the other way around: make a complete distance matrix, and do numpy magic to find the node_keys between which the distance is what you want. There's not really a convenient method in networkx for this though. Maybe `networkx.algorithms.shortest_paths.dense.floyd_warshall_numpy`, but it would require profiling.

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+        raise ValueError('All atoms from the selection must have coordinates. '
+                         'The following atoms do not have some: {}.'
+                         .format(' '.join(missing)))
+    coordinates = np.stack(coordinates)
+    distance_matrix = self_distance_matrix(coordinates)
+    constants = compute_force_constants(distance_matrix, lower_bound,
+                                        upper_bound, decay_factor, decay_power,
+                                        base_constant, minimum_force)
+    connectivity = build_connectivity_matrix(molecule, 2, selection=selection)
+    constants *= ~connectivity
```
pckroon: This one needs a comment

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+    for from_idx, to_idx in zip(*np.triu_indices_from(constants)):
+        from_key = selection[from_idx]
+        to_key = selection[to_idx]
+        force_constant = constants[from_idx, to_idx]
+        length = round(distance_matrix[from_idx, to_idx], 5)
```
pckroon: Why not do `distance_matrix = distance_matrix.round(5)` before the loop?

vermouth/processors/apply_rubber_band.py
```diff
@@ -0,0 +1,120 @@
...
+        if force_constant > minimum_force:
```
pckroon: Should this check not be part of `compute_force_constants`?

vermouth/processors/do_mapping.py
```diff
@@ -48,25 +48,32 @@ class GraphMapping:
...
+            from_idxs = list(self.block_from.find_atoms(atomname=name_from, resid=res_from))
             for res_to, name_to in to:
-                from_idxs = self.block_from.find_atoms(atomname=name_from, resid=res_from)
                 to_idxs = self.block_to.find_atoms(atomname=name_to, resid=res_to)
                 for to_idx in to_idxs:
                     self.mapping[to_idx].update(from_idxs)
+                    if from_idxs:
+                        self.weights[to_idx][from_idxs[0]] = (
+                            weights[(res_to, name_to)][(res_from, name_from)]
+                        )
```
pckroon: Why only `from_idxs[0]`? I'm lost in my own code :s

vermouth/processors/apply_rubber_band.py
jbarnoud: I don't think so. The point of that line is to make sure that the force constant for the self pair is 0, which could not be the case depending of the parameters for the decay; but may especially not be the case because of rounding.

vermouth/processors/apply_rubber_band.py
jbarnoud: It would work indeed. I'll let optimization for latter, though. We need a general profiling pass anyway.

vermouth/processors/apply_rubber_band.py
jbarnoud: `compute_force_constants` compute the constants, what we do with them is not its responsibility. Also, filtering the force constants in `compute_force_constants` will only change the condition of the test, it will not get rid of it.

vermouth/processors/do_mapping.py
jbarnoud: I am not completely sure either from the top of my head. But I'd rather have some tests in place before messing around with some refactoring.

vermouth/processors/apply_rubber_band.py
pckroon: I meant at line 33 ;)

vermouth/processors/apply_rubber_band.py
pckroon: In `make_force_constants` there's already a check that does
`constants[constants < minimum_force] = 0`
So maybe just `if force_constant`?

vermouth/processors/apply_rubber_band.py
jbarnoud: Oh! It was related to the comment above!

pckroon: Jup :) You're too fast

vermouth/processors/apply_rubber_band.py
jbarnoud: Indeed. It is possible, but then the way it is here may be more explicit.

pckroon: Hmmn. Or set it to np.nan in compute_force_constants, and check for that here. Now you're doing the same thing twice. And set a comment that it's set to nan in said function

pckroon:
By the way, is the WIP label/tag still relevant?

jbarnoud:
Hum... In a sense... everything is always a work in progress?

jbarnoud:
What is still blocking in this PR?

pckroon:
IIRC the missing docstrings.

jbarnoud:
I added docstring!

vermouth/processors/apply_rubber_band.py
```diff
@@ -20,16 +20,48 @@
...
+    return np.sqrt(
+        np.sum(
+            (coordinates[:, np.newaxis, :] - coordinates[np.newaxis, :, :]) ** 2,
+            axis=-1)
+    )
 
 
 def compute_decay(distance, shift, rate, power):
+    """
+    Compute the decay function of the force constant as function to the distance.
```
pckroon: too vague imho

vermouth/processors/apply_rubber_band.py
```diff
@@ -39,29 +71,117 @@ def compute_force_constants(distance_matrix, lower_bound, upper_bound,
...
+    lower_bound, upper_bound: float
+        The minimum and maximum length for a bond to be added, expressed in
+        nanometers.
+    decay_factor, decay_power: float
+        Parameters for the decay function.
+    base_constant: float
+        The base force constant for the bonds in kJ/mol/nm^2. If 'decay_factor'
+        or 'decay_power' is set to 0, then it will be the used force constant.
+    minimum_force: float
+        Minimum force constat in kJ/mol/nm^2 under which bonds are not kept.
```
pckroon: :math:? Or TeX formatting for the units? I'm pretty sure sphinx/rest can deal with it somehow.

jbarnoud:
I applied a compromise on the units.

PR #58, opened 2018-03-20
 Read proper and improper dihedrals separatly

jbarnoud:
While proper dihedral angles can be used to set edges, improper ones cannot. Storing them separately makes things easier, also it allows to care about the difference only in places we know about the format: the reading of the ff file, and the writing of the itp file.

Creating edges for improper dihedral in blocks leads to erroneous graphs, and therefore erroneous results. Not creating them, however, makes the writing of links painstaking and error prone as you manually need to define the edges manually while you do not with the other common interaction types.

Improper dihedrals can also be written directly in the [ impropers ] section. However, not allowing them to live in the [ dihedrals ] one in counter intuitive for gromacs user and will break the assumption that a simple ITP file can serve as a force field file.

### DIFF
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -527,6 +527,7 @@ def read_ff(lines):
         'impropers': 4,
         'constraints': 2,
         'virtual_sites2': 3,
+        'pairs': 2,
     }
 
     macros = {}
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -595,9 +595,24 @@ def read_ff(lines):
                 raise IOError('Error while reading line {} in section {}.'
                               .format(line_num, section))
 
-    # Finish the blocks and the links. We have the nodes and The
-    # interactions, but the edges are missing.
-    for name, block in blocks.items():
+    # Finish the blocks and the links.
+    # Because of hos they are described in gromacs, proper and improper
+    # dihedral angles are all under the [ dihedrals ] section. However
+    # the way they are treated differently in the library, at least on how
+    # they generate edges. Here we move the all the impropers into their own
+    # [ impropers ] section.
+    for block in blocks.values():
+        propers = []
+        impropers = []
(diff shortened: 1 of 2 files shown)


pckroon:
You can still put it under the [dihedrals] section but with a different function type. I don't mind splitting impropers and dihedrals in our internal datastructure. There we can deviate from "gromacs standards" as much as we like,

jbarnoud:
My PR message may not be clear. I already do the split on the basis of the function type. The reasoning here is that the function type only has meaning for gromacs. Rather than testing the function type here and there in the code, we test it once in a place that is format-specific.

What happens now is that the user can write the impropers gromacs-style in the [ dihedrals ] section (and therefore recycle an ITP file), or write it explicitly in the [ impropers ] section (which is a collateral effect of the way the parser is written). In any case, the FF parser filters what it reads in the [ dihedrals ] section based on the function type to move the impropers in the right place.

As a side effect, it fixes an issue with the RTP format that was already using an [ impropers ] section that ended up in the ITP by accident.

pckroon:
Hmmn. I read too quickly as well; I though tthis was an issue, not a PR.
I suggest we let @Tsjerk decide the fileformat question: Just [dihedrals] with function types, or [dihedrals] and [impropers]?
Both have merit as far as I can see.

jbarnoud:
This is for the current FF format, it should not affect @Tsjerk final format in any way. I needed the change for Ilias's molecules to work, his atomistic ITP that we want to use as block contains improper dihedrals that pollute the graph. The solution is to exclude these interactions when creating the edges. To exclude them, there are mostly 3 solutions:
* exclude all the dihedrals from the edge generation and add the edges manually
* test the function type in the code when required
* separate them using the function type once when reading the file

I think the last one is the least disruptive one. It does not break expectations, does not imply gromacs knowledge out of the IO, and therefore does not put constraints on a future file format.

Issue #61, opened 2018-03-21
Molecule.copy does not copy the interactions


jbarnoud:
As the title says.

jbarnoud:
One consequence is that `Molecule.subgraph` does not have interaction either. Though, subgraps should likely have their own logic to copy the interactions as they should also filter them.

pckroon:
Both should be relatively easy fixes though. The filtering logic should probably go in `copy` to solve both.

Issue #60, opened 2018-03-21
`Molecule.subgraph` does not preserve the atom order with networkx 2.1


jbarnoud:
See https://github.com/networkx/networkx/issues/2911

pckroon:
Closed downstream, we need to fix/implement this ourselves.

PR #62, opened 2018-03-24
Setup and fix some tests

jbarnoud:
We are not going to have everything unit tested in one go, but let's at least start and have the tests running on travis.


jbarnoud:
The PR is ready for review. Sorry if it produced a lot of notifications...

The test fails at the moment because they already account for #60 and #61.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+    - "3.4"
+    - "3.5"
+    - "3.5-dev"
+    - "3.6"
+    - "3.6-dev"
+    - "3.7-dev"
+matrix:
+    allow_failures:
+        - python: "2.7"
+        - python: "3.4"
```
pckroon: 3.4 is not an allowed failure, and it's errors should be fixed.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+    apt:
+        packages:
+            - dssp
+env:
+    - VERMOUTH_TEST_DSSP=mkdssp
+install:
+    - pip install pylint
+    - pip install --upgrade .
+script:
+    - pytest -vv vermouth
```
pckroon: -vv? Is that nescessary?

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+after_script:
+    - pylint vermouth
+    - pylint bin/martinize2
```
pckroon: You could argue that this should be part of the test, and not just a nice output to see. Either way, we'll probably need a custom .pylintrc. 
I guess pytest has a pylint plugin, to wrap it all in one.
Also, do we want to look at test coverage?

.travis.yml
jbarnoud: The main point is to display the whole difference when an assert fails. My experience is that you will likely rerun the failing tests with -vv so having it from the start saves some time.

.travis.yml
jbarnoud: There are way too many messages at the moment. We need a .pylintrc AND a lot of fixing before we can use it as a test.

pckroon: Let's keep it in mind for the future. 
And yes, if I recall correctly, PyLint is rather... How do I phrase this... Strict.

jbarnoud:
Should be fixed now.

jbarnoud:
Also, I suggest we add the coverage latter. First we have to choose what service to use to display the coverage, but also we do not have enough tests yet for the coverage to be meaningful.

pckroon:
Nice work, thanks :)

PR #64, opened 2018-03-26
Implement Elnedyn

jbarnoud:
This PR makes sure that Elnedyn works. Elnedyn has many divergences with standard Martini which necessitated modifications in the code. Among other:

* The backbone parameters are taken from the structure. This PR makes it possible to define bonds and angles with equilibrium values measured from the molecule. This is done using the new link effectors.
* The backbone bead is located on the C alpha. This is done by mapping the BB bead to the CA atom. However, to keep the connectivity of the backbone, the other backbone atoms have to be mapped to the BB bead as well, but with a weight of 0. This is expressed in the mapping file by prefixing the destination atom with a `!`.
* The bond function type for the elastic network differs between martini+EN and elnedyn. This means that the default type has to be defined in the force field. This PR adds the possibility to define arbitrary variables in a force field. These variables are stored in the `variables` attribute of the `ForceField` object.

I know, docstrings. It comes :stuck_out_tongue:.


pckroon:
WIP until the docstrings are there.

jbarnoud:
Rebased and documented.

vermouth/molecule.py
```diff
@@ -33,8 +34,218 @@
...
+    Parameters
+    ----------
+    value:
+        The value the attribute is tested not to be.
+    """
+    def match(self, node, key):
+        """
+        Apply the comparison.
+        """
+        return key not in node or node[key] != self.value
```
pckroon: What happens if self.value is None?

vermouth/molecule.py
```diff
@@ -33,8 +34,218 @@
...
+class LinkParameterEffector:
+    """
+    Rule to calculate an interaction parameter in a link.
+
+    This class allows to store dynamic parameters in link interactions. The
+    value of the parameter can be computed from the graph using the node keys
+    given when creating the instance.
+
+    An instance of this lass is first initialized with a list of node keys
+    from the import link in which it is defined. The instance is latter called
```
pckroon: typo

vermouth/molecule.py
```diff
@@ -33,8 +34,218 @@
...
+    like a function, and takes as arguments a molecule and a match dictionary
+    linking the link nodes with the molecule ones. The format of the dictionary
+    is expected to be ``{link key: molecule key}``.
+
+    This is a base class; it needs to be subclassed. A subclass must define an
+    :meth:`_apply` method that takes a molecule and a list of node keys from
```
pckroon: I'm not convinced you need to define an underscore method. My preference would be either `apply` or `apply_`

vermouth/molecule.py
```diff
@@ -388,18 +599,71 @@ def __init__(self):
...
-        if isinstance(value, Choice):
-            if attributes.get(attr) not in value:
-                return False
-        else:
-            if attributes.get(attr) != value:
+        if attr in ignore_keys:
+            continue
+        if isinstance(value, LinkPredicate):
+            match = value.match(attributes, attr)
+            if not value.match(attributes, attr):
```
pckroon: `if not match`? Now you're calling the method twice.

vermouth/molecule.py
```diff
@@ -388,18 +599,71 @@ def __init__(self):
...
+            return False
     return True
 
 
 def interaction_match(molecule, interaction, template_interaction):
+    """
+    Compare an interaction with a template interaction or interaction to delete.
+
+    An instance of :class:`Interaction` matches a template instance of the same
+    class or of :class:`DeleteInteraction` if, *a minima*, it involves the same
```
pckroon: *a minima* is confusing, since it's also valid english, and you don't get the italics in source

vermouth/map_input.py
```diff
@@ -0,0 +1,178 @@
...
+            continue
+        elif cleaned[0] == '[':
+            if cleaned[-1] != ']':
+                raise IOError('Format error at line {}.'.format(line_number))
+            context = cleaned[1:-1].strip()
+        elif context == 'molecule':
+            name = cleaned
+        elif context == 'atoms':
+            _, from_atom, *to_atoms = cleaned.split()
+            mapping[from_atom] = to_atoms
```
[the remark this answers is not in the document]
jbarnoud: backward will not support multiple atoms with the same name. It does not even support `to_atoms` to have the same name as `from_atoms`. Are you sure names are not unique?

vermouth/map_input.py
```diff
@@ -0,0 +1,178 @@
...
+    }
+    for to_atom, from_weights in null_weights.items():
+        null_keys = set(from_weights.keys())
+        non_null_keys = set(weights.get(to_atom, {}).keys())
+        redifined_keys = null_keys & non_null_keys
+        if redifined_keys:
+            msg = ('Atom(s) {} is mapped to "{}" with and without a weight '
+                   'in the molecule "{}". '
+                   'There cannot be the same target atom name with and '
+                   'without a "!" prefix on a same line.')
```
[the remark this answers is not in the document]
jbarnoud: You may have a bead that is shared for the purpose of the edges, but counts in only one bead in term of bead positioning.

vermouth/map_input.py
```diff
@@ -0,0 +1,178 @@
...
+    dict
+        A collection of mappings.
+    """
+    directory = Path(directory)
+    mappings = {}
+    for path in directory.glob('**/*.map'):
+        with open(path) as infile:
+            name, all_from_ff, all_to_ff, mapping, weights, extra = read_mapping(infile)
+        for from_ff in all_from_ff:
+            mappings[from_ff] = mappings.get(from_ff, {})
```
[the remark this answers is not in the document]
jbarnoud: Yep. A default dict of default dict!

vermouth/molecule.py
jbarnoud: Good catch. I updated the doc.

vermouth/molecule.py
jbarnoud: It was `apply` initially. But the method should not be used directly, which the underscore makes clear.

vermouth/map_input.py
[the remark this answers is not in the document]
pckroon: I have recently seen mapping files wheres the atomnames are not unique (beadnames are). Instead atoms are defined by their index (`_` here).
This sort of makes sense since atomnames do not have to be unique for gmx, but atomindices do.

vermouth/map_input.py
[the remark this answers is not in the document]
pckroon: You made a monster of a defaultdict, but you're not using it here ;)

pckroon:
When do you want to tackle the tokenizer?

jbarnoud:
Which tokenizer? The one from the force field file parser?

pckroon:
Yes. But I guess the file format is temporary anyway, so no sense spending a lot of time on it...

jbarnoud:
I'll be careful with how long temporary stuff last. What is the issue with the tokenizer?

pckroon:
The truly colossal try/except. @Tsjerk needs to come up with a definitive file format so we can fix it once and for all.

jbarnoud:
I was afraid you were alking about the actual tokenizer that is at the beginning of the module. It may be my most beautiful docstring!

Anyway, fixing the parser is out of scope for that PR.

PR #67, opened 2018-03-28
Allow calculation of dihedral angles

jbarnoud:
Add a function to calculate dihedral angles, but also dihedral angles shifted by -pi as used by scFix. The function also adds the corresponding link parameter effectors.

The function to calculate the angles is moved to the new `geometry.py` module that even has tests.

### DIFF
--- a/vermouth/geometry.py
+++ b/vermouth/geometry.py
@@ -53,6 +53,8 @@ def dihedral(vectorAB, vectorBC, vectorCD):
              \
               C - D
     
+    The angle around the BC axis is returned with values between -pi and +pi.
+    The direction of the angle follows the right hand convention.
     """
     normalABC = np.cross(vectorAB, vectorBC)
     normalBCD = np.cross(vectorBC, vectorCD)
@@ -60,3 +62,20 @@ def dihedral(vectorAB, vectorBC, vectorCD):
     pcos = np.dot(normalABC, normalBCD)
     return np.arctan2(psin, pcos)
 
+
+def dihedral_left(vectorAB, vectorBC, vectorCD):
+    """
+    Calculate a dihedral angle in radians following the left hand convention.
+
+    See Also
+    --------
+    dihedral
+        Calculate a dihedral angle in the right hand convention.
+    """
+    angle = dihedral(vectorAB, vectorBC, vectorCD)
+    angle -= np.pi
+    if angle > np.pi:
+        angle -= 2 * np.pi
+    if angle < -np.pi:
+        angle += 2 * np.pi
+    return angle
--- a/vermouth/geometry.py
+++ b/vermouth/geometry.py
@@ -63,14 +63,14 @@ def dihedral(vectorAB, vectorBC, vectorCD):
     return np.arctan2(psin, pcos)
 
 
-def dihedral_left(vectorAB, vectorBC, vectorCD):
+def dihedral_phase(vectorAB, vectorBC, vectorCD):
     """
-    Calculate a dihedral angle in radians following the left hand convention.
+    Calculate a dihedral angle in radians with a -pi phase correction.
 
     See Also
     --------
     dihedral
-        Calculate a dihedral angle in the right hand convention.
+        Calculate a dihedral angle.
     """
     angle = dihedral(vectorAB, vectorBC, vectorCD)
     angle -= np.pi
--- a/vermouth/geometry.py
+++ b/vermouth/geometry.py
@@ -36,26 +36,28 @@ def angle(vectorBA, vectorBC):
     denominator = np.linalg.norm(vectorBA) * np.linalg.norm(vectorBC)
     cosine = nominator / denominator
     # Floating errors at the limits may cause issues.
-    if cosine > 1:
-        cosine = 1
-    elif cosine < -1:
-        cosine = -1
+    cosine = np.clip(cosine, -1, 1)
     return np.arccos(cosine)
 
 
-def dihedral(vectorAB, vectorBC, vectorCD):
+def dihedral(coordinates):
     """
-    Calculate the dihedral angle in radians formed by 3 vectors.
+    Calculate the dihedral angle in radians.
 
-    The function assumes the following situation:
+    Parameters
+    ----------
+    coordinates: np.ndarray
+        The coordinates of 4 points defining the dihedral angle. Each row
+        corresponds to a point, and each column to a dimension.
 
-        A - B
-             \
-              C - D
-    
-    The angle around the BC axis is returned with values between -pi and +pi.
-    The direction of the angle follows the right hand convention.
+    Returns
+    -------
+    float
+        The calculated angle between -pi and +pi.
     """
+    vectorAB = coordinates[1, :] - coordinates[0, :]
+    vectorBC = coordinates[2, :] - coordinates[1, :]
+    vectorCD = coordinates[3, :] - coordinates[2, :]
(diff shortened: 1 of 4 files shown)


vermouth/geometry.py
```diff
@@ -0,0 +1,81 @@
...
+    if cosine > 1:
+        cosine = 1
+    elif cosine < -1:
+        cosine = -1
+    return np.arccos(cosine)
+
+
+def dihedral(vectorAB, vectorBC, vectorCD):
+    """
+    Calculate the dihedral angle in radians formed by 3 vectors.
```
pckroon: By the strict definition an (improper) dihedral is the angle between 2 planes, each spanned by 2 vectors. I would change the signature to take either two pairs of vectors, or 4 vectors.
What you're describing here is a torsion angle.

vermouth/geometry.py
```diff
@@ -0,0 +1,81 @@
...
+
+        A - B
+             \
+              C - D
+    
+    The angle around the BC axis is returned with values between -pi and +pi.
+    The direction of the angle follows the right hand convention.
+    """
+    normalABC = np.cross(vectorAB, vectorBC)
+    normalBCD = np.cross(vectorBC, vectorCD)
```
pckroon: Do these need to be normalized?
What happens if ABC or BCD is (nearly) colinear?

vermouth/geometry.py
```diff
@@ -0,0 +1,81 @@
...
+    psin = np.dot(normalABC, vectorCD) * np.linalg.norm(vectorBC)
+    pcos = np.dot(normalABC, normalBCD)
+    return np.arctan2(psin, pcos)
+
+
+def dihedral_phase(vectorAB, vectorBC, vectorCD):
+    """
+    Calculate a dihedral angle in radians with a -pi phase correction.
```
pckroon: I get the feeling the phase correction should be an argument

vermouth/geometry.py
```diff
@@ -0,0 +1,81 @@
...
+    --------
+    dihedral
+        Calculate a dihedral angle.
+    """
+    angle = dihedral(vectorAB, vectorBC, vectorCD)
+    angle -= np.pi
+    if angle > np.pi:
+        angle -= 2 * np.pi
+    if angle < -np.pi:
+        angle += 2 * np.pi
```
pckroon: And if it is an argument, these should be while statements.

vermouth/tests/test_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        The angle between the 3 points.
+    """
+    # The shift is arbitrary. Its purpose is to avoid having the angle centered
+    # at the origin as it may be a special case.
+    shift = np.array([1.1, -7.2, 9.1])
+    coordinates = np.array([
+        [2, 0, 0],
+        [0, 0, 0],
+        [0, 0, 0],  # Will be redifined
+    ]) + shift
```
pckroon: `np.full((3, 3), shift)`; `coords[0,0]+=2`?

vermouth/tests/test_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+    Yields
+    ------
+    coordinates: np.ndrarray
+        The coordinates of the points. Each row corresponds to a pointm and
+        each column corresponds to a dimension.
+    angle: float
+        The angle between the 3 points.
+    """
+    # The shift is arbitrary. Its purpose is to avoid having the angle centered
+    # at the origin as it may be a special case.
```
pckroon: So make sure we also test the special case ;)

vermouth/tests/test_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+    shift = np.array([1.1, -7.2, 9.1])
+    coordinates = np.array([
+        [2, 0, 0],
+        [0, 0, 0],
+        [0, 0, 5],
+        [0, 0, 5],  # Will be redifined
+    ]) + shift
```
pckroon: Same comments as above

vermouth/geometry.py
jbarnoud: It seems that an phase of -pi is a specific and useful case. It seems to be use by gromacs. I fail to grasp why, though. I'll add later a function that can rotate any angle and keep it in bound. In an other PR if you do not mind.

vermouth/tests/test_geometry.py
pckroon: I understand that. By the way, are you sure the broadcasting of (3, 3) + (3, ) happens as you expect it to? Not that I really matters if it doesn't.

vermouth/geometry.py
pckroon: Sure. Just don't forget ;)

vermouth/tests/test_geometry.py
```diff
@@ -101,29 +102,56 @@ def _generate_test_dihedrals(n_angles):
...
-    assert np.allclose(geometry.dihedral(vectorAB, vectorBC, vectorCD), angle)
-
-
-@pytest.mark.parametrize('points, angle', _generate_test_dihedrals(10))
+    calc_angle = geometry.dihedral(points)
+    # +pi and -pi are the same angle; we normalize them to pi
+    if np.allclose(calc_angle, -np.pi):
+        calc_angle *= -1
+    if np.allclose(angle, -np.pi):
+        angle *= -1
```
pckroon: I get the feeling this is not entirely correct. Adding 2 pi would be better I think.

vermouth/tests/test_geometry.py
```diff
@@ -101,29 +102,56 @@ def _generate_test_dihedrals(n_angles):
...
         angle_phase -= 2 * np.pi
     if angle_phase < -np.pi:
         angle_phase += 2 * np.pi
-    assert np.allclose(geometry.dihedral_phase(vectorAB, vectorBC, vectorCD), angle_phase)
+    calc_angle = geometry.dihedral_phase(points)
+    # +pi and -pi are the same angle; we normalize them to pi
+    if np.allclose(calc_angle, -np.pi):
+        calc_angle *= -1
+    if np.allclose(angle_phase, -np.pi):
+        angle_phase *= -1
```
pckroon: Same as above

vermouth/geometry.py
```diff
@@ -36,43 +36,56 @@ def angle(vectorBA, vectorBC):
...
-    if cosine > 1:
-        cosine = 1
-    elif cosine < -1:
-        cosine = -1
+    cosine = np.clip(cosine, -1, 1)
     return np.arccos(cosine)
 
 
-def dihedral(vectorAB, vectorBC, vectorCD):
+def dihedral(coordinates):
```
pckroon: I liked the previous signature better. But if you argue this is a low-level function, sure. Just make sure there's also human usable ones which calculate torsion angles and angles between planes respectively.

vermouth/tests/test_geometry.py
jbarnoud: It should be the exact same, I just want -pi and pi to compare to the same thing.

vermouth/geometry.py
jbarnoud: In practice I always used all dihedral calculation functions with 4 points. It is what gromacs, MDAnalysis, or VMD do. I could have a 6 points version, or a 2 times 3 point version as well.

pckroon: The problem with a 4pt version is that impropers and torsion angles become separate things. I do agree it's the most common thing you'll find.

PR #68, opened 2018-03-28
Add flexibility in the choice of the force fields

jbarnoud:
This commit makes it possible for a user to go beyond the universal to martini transformation by making it possible to select a origin force field in addition to the target one. It also makes possible to use user defined force fields and mappings that are crucial when developing models.

In practice, the PR adds the following options to martinize2:
* `-from`: the name of the origin force field
* `-ff-dir`: the path to a directory containing force field directories
* `-map-dir`: the path to a directory containing mapping files

In addition to these options, the PR adds the ability to generate self mappings for a force field, which martinize2 makes now by default. This makes it possible to generate topologies from a structure without force field transformation.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -245,8 +245,6 @@ def entry():
         raise ValueError('No mapping known to go from "{}" to "{}".'
                          .format(from_ff, args.to_ff))
 
-    print(known_mappings['universal']['universal'])
-
     # Reading the input structure.
     # So far, we assume we only go from atomistic to martini. We want the
     # input structure to be a clean universal system.
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -167,6 +167,14 @@ def entry():
     ff_group = parser.add_argument_group('Force field selection')
     ff_group.add_argument('-ff', dest='to_ff', default='martini22',
                           help='Which forcefield to use')
+    ff_group.add_argument('-from', dest='from_ff', default='universal',
+                          help='Force field of the original structure.')
+    ff_group.add_argument('-ff-dir', dest='extra_ff_dir', action='append',
+                          type=Path, default=[],
+                          help='Additional repository for custom force fields.')
+    ff_group.add_argument('-map-dir', dest='extra_map_dir', action='append',
+                          type=Path, default=[],
+                          help='Additional repository for mapping files.')
     posres_group = parser.add_argument_group('Position restraints')
     posres_group.add_argument('-p', dest='posres', type=lambda x: x.lower(),
                               choices=('none', 'all', 'backbone'), default='none',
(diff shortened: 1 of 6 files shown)


bin/martinize2
```diff
@@ -209,9 +222,26 @@ def entry():
     )
     known_mappings = read_mapping_directory(Path(DATA_PATH) / 'mappings')
 
-    from_ff = 'universal'
+    # Add user force fields and mappings
+    for directory in args.extra_ff_dir:
+        if not directory.is_dir():
+            raise ValueError('"{}" given to the -ff-dir option should be a directory.')
+        vermouth.forcefield.find_force_fields(directory, known_force_fields)
```
pckroon: Throw low, catch high. I'd rather put a try/except around the call to find_force_fields, and reraise it with a nicer error message than testing whether it's a directory beforehand.

bin/martinize2
```diff
@@ -209,9 +222,26 @@ def entry():
...
+    for directory in args.extra_map_dir:
+        if not directory.is_dir():
+            raise ValueError('"{}" given to the -map-dir option should be a directory.')
+        partial_mapping = read_mapping_directory(directory)
```
pckroon: Same here

bin/martinize2
```diff
@@ -273,13 +304,21 @@ def entry():
...
     if args.top_path is not None:
         write_gmx_topology(system, args.top_path,
                            deduplicate=not args.keep_duplicate_itp)
 
+    # Write a PDB file.
+    # Get rid of the charge, the PDB format is too primitive to actually write them.
+    for molecule in system.molecules:
+        for node in molecule.nodes.values():
+            try:
+                del node['charge']
```
pckroon: This is a peculiarity of the file format, and should be dealt with by write_pdb

vermouth/map_input.py
```diff
@@ -182,3 +186,104 @@ def read_mapping_directory(directory):
...
+
+    Parameters
+    ----------
+    known_mappings: dict
+        Collection of mapping to update **in-place**.
+    partial_mapping: dict
+        Collection of mappings to update from.
+    """
+    for origin, destinations in partial_mapping.items():
+        known_mappings[origin] = known_mappings.get(origin, {})
```
pckroon: This looks like a defaultdict of a defaultdict again, but I'm not sure we can use that here since it's an existing datastructure. I guess we can't :(

vermouth/molecule.py
```diff
@@ -600,16 +615,31 @@ def to_molecule(self, atom_offset=0, resid=1, offset_charge_group=1):
...
+        # what has been passed in the 'attr' argument.
+        defaults = {
+            'non_edges': [],
+            'removed_interactions': {},
+            'molecule_meta': {},
+            'patterns': [],
+        }
+        for attribute, default_value in defaults.items():
+            if not hasattr(self, attribute):
+                setattr(self, attribute, default_value)
```
pckroon: We can wrap the whole default values things in a class, and do some magical inheritance if we want to cut back on code duplication. Or make it a function, if you're not feeling adventurous.

bin/martinize2
jbarnoud: Done. I knew it but I was lazy.

bin/martinize2
jbarnoud: Done.

bin/martinize2
jbarnoud: Done.

vermouth/pdb/pdb.py
```diff
@@ -66,7 +66,7 @@ def keyfunc(graph, node_idx):
             temp_factor = get_not_none(node, 'temp_factor', 0)
             element = get_not_none(node, 'element', '')
             charge = get_not_none(node, 'charge', 0)
-            if charge:
+            if charge and not omit_charges:
```
pckroon: Are you sure this is the way to do it? This still puts file format issues on the user. Instead it might make more sense to see whether the charge is within <error> of an integer.

jbarnoud: We want to write either all the charges or none of them. This means that we would have to check if all the charges are integer. Doing this check is not very difficult, however its output is not predictable for the final user that ends up with or without charges in a non obviously predictable way. Having a full switch has the benefit of letting the user in control.

pckroon: You could argue the user made that choice by choosing to write it as a pdb file, but ok

PR #70, opened 2018-03-29
Make PDB reader more loose

jbarnoud:
Allow the resname to be 4 characters by reading the empty column after the resname as part of the resname. This does not affect well formatted PDB file as that column is not assigned to any type of data in the PDB standard, but it allows to read PDB files that make use of that empty column, such as those from the ATB.

This is a common variation around the PDB format, and it is supported at least by pymol and VMD.

Note that the writing is not modified, so 4 chars resnames are still truncated at writing to write valid PDB files.


jbarnoud:
It is comment enough that a warning seems overkill. A note maybe.

PR #71, opened 2018-03-29
Allow to define a format for link parameter effectors

jbarnoud:
Some force field interactions expect their parameters to be formatted in a specific way. For instance, the side chain dihedral correction expect the angle to be rounded to the integer. This PR adds the possibility to define a format to a link effector.

While I am fine with the way it is implemented in the `LinkParameterEffector` class, I am open to suggestion about the syntax in the ff file. For now, writing a link that defines a SC1-BB-BB-SC1 torsion looks like:

```
[ link ]
[ dihedrals ]
SC1 BB +BB +SC1 1 dihedral(SC1,BB,+BB,+SC1) 75 1
```

With the syntax introduce in this PR, it becomes possible to limit the dihedral to 2 decimal places for instance by writing:

```
[ link ]
[ dihedrals ]
SC1 BB +BB +SC1 1 dihedral(SC1,BB,+BB,+SC1|.2f) 75 1
```

This syntax has the benefit to be super easy to parse. The pipe character was chosen because it is already forbidden in atom names for other reasons. It works well enough, especially as a prototype. But it is not really satisfying.

### DIFF
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -368,8 +368,12 @@ def _parse_interaction_parameters(tokens):
             except KeyError:
                 raise IOError('{} is not a known parameter effector.'
                               .format(effector_name))
+            if '|' in effector_param_str:
+                effector_param_str, effector_format = effector_param_str.split('|')
+            else:
+                effector_format = None
             effector_param = [elem.strip() for elem in effector_param_str.split(',')]
-            parameter = effector_class(effector_param)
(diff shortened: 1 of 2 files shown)


pckroon:
```
[ link ]
[ dihedrals ]
SC1 BB +BB +SC1 1 {dihedral(SC1,BB,+BB,+SC1):.2f} 75 1
```
To make it more like Pythons formatting syntax?

PR #72, opened 2018-03-30
Allow links to declare features

jbarnoud:
With this PR, links can declare the feature they implement. This allow to check is the feature is available in a given force field. Martinize2 benefit from this feature by issuing a warning when a command line argument ask for a feature that is not declared.

This should avoid surprises for user, and ease the life of people who write force field files.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -217,7 +217,7 @@ def entry():
                           help='Comma separated list of bead names for elastic bonds')
 
     prot_group = parser.add_argument_group('Protein description')
-    prot_group.add_argument('-nt', dest='neutral_terminii',
+    prot_group.add_argument('-nt', dest='neutral_termini',
                             action='store_true', default=False,
                             help='Set neutral termini (charged is default)')
     args = parser.parse_args()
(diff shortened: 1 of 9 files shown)


pckroon:
Good idea! But I'd like to have it read all the allowed options for features from a separate file somewhere. That way you can make noisy errors if someone makes a typo in a forcefield file.

jbarnoud:
Good idea. However, how do you handle user-defined feature?

pckroon:
The same as ours. They'll need to add them to the list/file.

jbarnoud:
Then the user should be able to provide its own custom list without touching the code. Maybe we could have a tool that validate in depth the force field and mapping files, and only have a lighter validation at runtime.

pckroon:
That's a good option, but for later. I do think it'll be valuable for FF development to have a list of features available somewhere. But like you said, we don't want FF devs to touch our code; so the list of available features has to be in a separate file.

pckroon:
I suggest we add the file with allowed features later, and merge this now. Do you agree?

jbarnoud:
I am fine with it. Latter we should think about how to deal with debugging of the input files. I have a few ideas, but if we want the extensibility to be real we need the program to give all the assistance it can.

pckroon:
Agreed.

PR #73, opened 2018-03-30
Add a version number using pbr

### LINKED ISSUE #66: Neither martinize2 nor vermouth have a version number
And this does not look very serious coming from us...

Should martinize2 be versioned separately from vermouth? What version number should we start with? How do we keep martinize2 and vermouth versions in sync?

Also, in the current version of martinize, there is a version number that is never updated but that ranges from 2.4 to 2.6 depending on where you download the file. Should we rename martinize2 to martinize3 and start counting with martinize 3.0 to avoid confusion? It will also fit nicely with martini getting its version 3.

jbarnoud:
Creates the `vermouth.__version__` attribute. Also adds `--version` to martinize2.

We will have to pay attention to our commit messages, though: https://docs.openstack.org/pbr/latest/user/features.html#version

The following strings are recognized in commit messages:
* feature
* api-break
* deprecation
* bugfix

When prefixes with `Sem-Ver: ` in the commit message, they can trigger a bump in the version number.

Fixes #66

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ keywords = martini MD martinize
 packages = find:
 setup-requires =
     setuptools >= 30.3.0
+    pbr
 install-requires =  # ?? requires-dist?
     pbr
     numpy
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ packages = find:
 setup-requires =
     setuptools >= 30.3.0
 install-requires =  # ?? requires-dist?
+    pbr
     numpy
     networkx ~= 2.0
     matplotlib  # See issue #23
--- a/vermouth/__init__.py
+++ b/vermouth/__init__.py
@@ -30,4 +30,8 @@ else:
     DATA_PATH = pkg_resources.resource_filename('vermouth', 'data')
     del pkg_resources
 
+import pbr.version
+__version__ = pbr.version.VersionInfo('vermouth').release_string()
+del pbr
+
 from .vermouth import *
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -41,6 +41,8 @@ import operator
 import textwrap
 import functools
 
+VERSION = 'martinize with vermouth {}'.format(vermouth.__version__)
+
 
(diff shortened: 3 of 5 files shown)


pckroon:
It will only do magic things if there's a line like this somewhere in the commit message.
```
Sem-Ver: bugfix
```
so I don't think we'll need to worry too much. Version increments should be done by tags.

setup.cfg
```diff
@@ -28,6 +28,7 @@ packages = find:
 setup-requires =
     setuptools >= 30.3.0
```
pckroon: I get the idea pbr should be added here (as well)

vermouth/__init__.py
```diff
@@ -30,4 +30,8 @@
     DATA_PATH = pkg_resources.resource_filename('vermouth', 'data')
     del pkg_resources
 
+import pbr.version
+__version__ = pbr.version.VersionInfo('vermouth').release_string()
+del pbr
```
pckroon: del pbr.version?

jbarnoud: `del pbr.version` does not delete `pbr`:

```
>>> import pbr.version
>>> del pbr.version
>>> pbr
<module 'pbr' from '/home/jon/local/anaconda3/lib/python3.6/site-packages/pbr/__init__.py'>
```

pckroon: And does deleting pbr also delete pbr.version?

setup.cfg
jbarnoud: I have the feeling that setup-requires should not be in setup.cfg anymore. The doc of pbr consistently puts it in setup.py, and having it at the two locations seems wrong. I'll try to remove it from setup.cfg and from setup.py when I'll be done cleaning the commit I am working on on a other branch.

pckroon: PEP-518; PEP-517.
Python packaging is a *mess* :(

vermouth/__init__.py
jbarnoud: If there is no reference it should get garbage collected.

PR #78, opened 2018-04-18
Repair mod2

### LINKED ISSUE #14: Termini, PTMs and protonation states
There's a start to detect and identify PTMs, termini and protonation states in repair_graph, but it's currently not using any of that information. We need a some sort of database (dict?) of known changes, and of course use it.
And tie it in with user input

It's going to be awesome.

jbarnoud:
Superseeds #2 

Rebased against master.

Closes #2 
Fixes #14


jbarnoud:
How are PTMs supposed to work with mappings? Termini for instance are not defined as part of the amino acid mappings, as a consequence, a residue with an OXT atom cannot be mapped. How do you see this handled?

pckroon:
The appropriate mapping should be selected or modified depending on the modifications found. Or all the residuenames need to be changed depending on modifications, but that might be a little overkill.

jbarnoud:
https://gist.github.com/jbarnoud/2707ea358d1ea691fe2337b1a051084d

vermouth/graph_utils.py
```diff
@@ -198,23 +241,20 @@ def initialize(self):
...
     Analogous to networkx.blockmodel, but can deal with incomplete partitions,
-    assigns ``attrs`` to nodes, and calculates the new ``position`` attribute as
-    center of geometry.
+    and assigns ``attrs`` to nodes.
 
     Parameters
     ----------
     G: networkx.Graph
         The graph to partition
-    partitions: iterable of iterables
```
jbarnoud: Shouldn’t that remains?

pckroon: Yes it should. I'm not sure why it disappeared...

Issue #23, opened 2018-01-11
Remove matplotlib requirement


jbarnoud:
We currently require matplotlib for dev testing. This should be removed before the release.

PR #81, opened 2018-04-24
Get rid of old and unused code

pckroon:
Needs to be tested by running an integration test. Why don't we have Travis do that yet?


pckroon:
Also, fixes #23

PR #88, opened 2018-06-18
Add coverage and CodeCov

pckroon:


jbarnoud:
Could you exclude the `tests` and the `redistributed` directories from the coverage report? Except for that I think it is good to go.

pckroon:
I also want codecov PR status/check, and mute the comment bot.

PR #96, opened 2018-07-27
Graph util tests

pckroon:
There's still a problem with the MCS code, it's in the comments. I'm not convinced it's actually a problem, but I don't know how to tell Hypothesis.

Still WIP


vermouth/tests/test_graph_utils.py
```diff
@@ -0,0 +1,384 @@
...
+         ([{'atomname': 0}], [{'atomname': 1}], {}, 0),
+         ([{'atomname': 0}], [{'atomname': 0}], {}, 0),
+         ([{'atomname': 0}], [{'atomname': 0}], {0:0}, 1),
+         ([{'atomname': 0}, {'atomname': 0}], [{'atomname': 0}, {'atomname': 0}], {0:0}, 1),
+         ([{'atomname': 0}, {'atomname': 0}], [{'atomname': 0}, {'atomname': 0}], {0:0, 1:1}, 2),
+         ([{'atomname': 0}, {'atomname': 0}], [{'atomname': 0}, {'atomname': 1}], {0:0, 1:1}, 1),
+         ([{'atomname': 0}, {'atomname': 1}], [{'atomname': 0}, {'atomname': 1}], {0:0, 1:1}, 2),
+        ]
+        )
+def test_rate_match(nodes1, nodes2, match, expected):
```
jbarnoud: The `match` parameter of `rate_match` is not documented. Also, could `match` be `{0:1}`?

vermouth/graph_utils.py
```diff
@@ -197,7 +230,7 @@ def isomorphism(reference, residue):
                 if len(H_names[res_H_name]) != 1:
```
jbarnoud: The case where this case is `False` is not tested.

vermouth/tests/test_hypothesis_graph_utils.py
```diff
@@ -0,0 +1,92 @@
...
+import networkx as nx
+
+import hypothesis.strategies as st
+from hypothesis import given, note, settings
+
+from hypothesis_networkx import graph_builder
+import vermouth
+
+
+max_nodes = 10
```
jbarnoud: These variable names do not follow PEP8. Should be capitalized.

vermouth/tests/test_hypothesis_graph_utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@settings(max_examples=500)
+@given(graph1=MCS_builder, graph2=MCS_builder, attrs=attrs)
+def test_maximum_common_subgraph(graph1, graph2, attrs):
+    expected = vermouth.graph_utils.categorical_maximum_common_subgraph(graph1, graph2, attrs)
+    
+    found = vermouth.graph_utils.maximum_common_subgraph(graph1, graph2, attrs)
+    
+    note(attrs)
```
jbarnoud: It would be better with something to tell what the note is about. Same as for the prints above.

vermouth/graph_utils.py
pckroon: I'll leave this as is for now. That part of the code (and the MCS) is due for refactor anyway.

pckroon:
I'll rebase against master after the second round of review.

vermouth/tests/test_graph_utils.py
```diff
@@ -0,0 +1,783 @@
...
+    """
+    mol = basic_molecule(node_data_in)
+    vermouth.graph_utils.add_element_attr(mol)
+    expected = basic_molecule(expected_node_data)
+    assert mol.nodes(data=True) == expected.nodes(data=True)
+
+
+@pytest.mark.parametrize('node_data_in,exception', [
+    ([{'atomname': '1234'}], ValueError),
+    ([{'peanuts': '1H3'}], ValueError),
```
jbarnoud: I know now why these case raise errors, but a quick comment at the end of the lines will save some time down the road. Something like "no atomname attribute", and "no alpha character".

vermouth/tests/test_graph_utils.py
```diff
@@ -0,0 +1,783 @@
...
+    """
+    Tests to make sure the function ``add_element_attr`` raises the expected
+    errors.
+    """
+    mol = basic_molecule(node_data_in)
+    with pytest.raises(exception):
+        vermouth.graph_utils.add_element_attr(mol)
+
+
+@pytest.mark.parametrize('node_data1,node_data2,attrs,expected', [
```
jbarnoud: You can have a space after each comma in the argument string. It makes it easier to read. (ame for the other occurrences).

vermouth/tests/test_graph_utils.py
```diff
@@ -0,0 +1,783 @@
...
+
+
+def test_blockmodel_graph_attr():
+    """
+    Make sure the function ``blockmodel`` produces node attributes ``'graph'``,
+    ``'nnodes'`` and ``'nedges'`` that have the correct values.
+    """
+    graph = basic_molecule([{}, {}, {}], {(0, 1): {}, (1, 2): {}})
+    found = vermouth.graph_utils.blockmodel(graph, [[0], [1, 2]])
+    assert 1 == found.nodes[0]['nnodes'] == len(found.nodes[0]['graph'].nodes)
```
jbarnoud: Pytest reports are easier to read when the value to validate is first.

vermouth/tests/test_hypothesis_graph_utils.py
```diff
@@ -0,0 +1,141 @@
...
+NODE_DATA = st.dictionaries(keys=st.sampled_from(ATTRNAMES),
+                            values=st.integers(min_value=0, max_value=MAX_NODES))
+
+ATTRS = st.lists(st.sampled_from(ATTRNAMES), unique=True, min_size=0, max_size=2)
+
+MCS_BUILDER = graph_builder(node_data=NODE_DATA, min_nodes=0, max_nodes=MAX_NODES,
+                            node_keys=st.integers(max_value=MAX_NODES, min_value=0))
+
+
+def make_into_set(iter_of_dict):
```
jbarnoud: This function is duplicated. It should go in a utility module.

vermouth/tests/test_graph_utils.py
```diff
@@ -639,10 +632,14 @@ def test_blockmodel_graph_attr():
     """
     graph = basic_molecule([{}, {}, {}], {(0, 1): {}, (1, 2): {}})
     found = vermouth.graph_utils.blockmodel(graph, [[0], [1, 2]])
-    assert 1 == found.nodes[0]['nnodes'] == len(found.nodes[0]['graph'].nodes)
-    assert 2 == found.nodes[1]['nnodes'] == len(found.nodes[1]['graph'].nodes)
-    assert 0 == found.nodes[0]['nedges'] == len(found.nodes[0]['graph'].edges)
-    assert 1 == found.nodes[1]['nedges'] == len(found.nodes[1]['graph'].edges)
+    assert 1 == found.nodes[0]['nnodes']
```
jbarnoud: I was more thinking about how it was before:

```
assert found.nodes[0]['nnodes'] == len(found.nodes[0]['graph'].nodes) == 1
```

Then, in case of failure, pytest says that `found.nodes[0]['nnodes']` does not have the expected value instead of saying that `1` is at fault.

A minor detail, though.

PR #100, opened 2018-08-30
Make sure codecov is installed

jbarnoud:


jbarnoud:
Youhou! Codecov works now! 80% threshold is tough, though.

PR #101, opened 2018-08-31
Read more than one mapping per mapping file

jbarnoud:
This is a preliminary step toward reading specific mappings for modifications (PTMs).

### DIFF
--- a/vermouth/map_input.py
+++ b/vermouth/map_input.py
@@ -343,9 +343,7 @@ def generate_self_mappings(blocks):
     -------
     mappings: dict[str, tuple]
         A dictionary of mappings where the keys are the names of the blocks,
-        and the values are tuples like (mapping, weights, extra). The elements
-        of these tuples are formatted as the corresponding output of the
-        :func:`read_mapping` function.
+        and the values are tuples like (mapping, weights, extra).
 
     Raises
     ------
@@ -354,7 +352,7 @@ def generate_self_mappings(blocks):
 
     See Also
     --------
-    read_mapping
+    read_mapping_file
         Read a mapping from a file.
     generate_all_self_mappings
         Generate self mappings for a list of force fields.
@@ -405,8 +403,7 @@ def combine_mappings(known_mappings, partial_mapping):
     similar to the output of the :func:`read_mapping_directory` function. They
     are dictionary with 3 levels of keys: the name of the initial force field,
     the name of the target force field, and the name of the block. The values
-    in the third level dictionary are tuples of (mapping, weights, extra) as
-    described in the :func:`read_mapping`.
+    in the third level dictionary are tuples of (mapping, weights, extra).
 
     If a force field appears in 'partial_mapping' that is not in
     'known_mappings', then it is added. For existing pairs of initial and
--- a/vermouth/map_input.py
+++ b/vermouth/map_input.py
@@ -20,27 +20,92 @@ Read force field to force field mappings.
 
 from pathlib import Path
 import collections
+import itertools
 
 
-def read_mapping(lines):
+def read_mapping_file(lines):
     """
     Partial reader for modified Backward mapping files.
 
-    ..warning::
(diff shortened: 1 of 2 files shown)


vermouth/map_input.py
```diff
@@ -20,28 +20,159 @@
...
+
+    Return
+    ------
+    dict
+    """
+    lines = iter(lines)
+    mappings = collections.defaultdict(lambda: collections.defaultdict(dict))
+
+    # Throw away everything before [ molecule ]
+    line_number = 1
```
pckroon: Needs a comment: used iff lines is empty.

vermouth/map_input.py
```diff
@@ -20,28 +20,159 @@
...
+    # We do not want to return defaultdicts.
+    mappings = _default_to_dict(mappings)
+
+    return mappings
+
+
+def _default_to_dict(mappings):
+    """
+    Convert a mapping collection from a defaultdict to a dict.
+    """
```
pckroon: I think I'm in favour of making this recursive:

(untested!)
```python
def _default_to_dict(mappings):
    if isinstance(mappings, dict):
        for key, val in mappings.items():  # reduction
            mappings[key] = _default_to_dict(val)  # combination
        return dict(mappings)
    else:
        # basecase
        return mappings
```

vermouth/map_input.py
```diff
@@ -20,28 +20,159 @@
...
+    null_weights = {
+        to_atom[1:]: {from_atom: 0 for from_atom in from_atoms}
+        for to_atom, from_atoms in rev_mapping.items()
+        if to_atom.startswith('!')
+    }
+    for to_atom, from_weights in null_weights.items():
+        null_keys = set(from_weights.keys())
+        non_null_keys = set(weights.get(to_atom, {}).keys())
+        redifined_keys = null_keys & non_null_keys
+        if redifined_keys:
```
pckroon: typo: redefined?

vermouth/map_input.py
```diff
@@ -20,28 +20,159 @@
...
+            msg = ('Atom(s) {} is mapped to "{}" with and without a weight '
+                   'in the molecule "{}". '
+                   'There cannot be the same target atom name with and '
+                   'without a "!" prefix on a same line.')
+            raise IOError(msg.format(redifined_keys, to_atom, name))
+        weights[to_atom] = weights.get(to_atom, {})
```
pckroon: You do this get twice. Can you combine it?

vermouth/tests/test_map_input.py
```diff
@@ -226,21 +237,97 @@ def test_read_mapping(case):
...
+    """,
 ))
-
 def test_read_mapping_errors(content):
     """
     Test that syntax error are caught when reading a mapping.
     """
     with pytest.raises(IOError):
-        vermouth.map_input.read_mapping(content.split('\n'))
+        vermouth.map_input._read_mapping_partial(content.split('\n'), 1)  # pylint: disable=protected-access
```
pckroon: Move the pylint stanza to the top of the file. Otherwise I expect it'll be all over.

vermouth/tests/test_map_input.py
```diff
@@ -226,21 +237,97 @@ def test_read_mapping(case):
...
+
+        ; Some mess between two molecules.
+
+        [ molecule ]
+        dummy_1
+
+        [ atoms ]
+        0 X2 C
+        1 X3 D
+    """).split('\n')
```
pckroon: Comments after atom lines, comments in between atom lines, comments in the [molecule] section. Especially:
```
[ molecule ]
; Look, it's a shiny!
dummy_0
```

vermouth/map_input.py
jbarnoud: Maybe, but it would need to scramble the current logic which is very easy to get wrong.

jbarnoud:
@pckroon Your comments should be addressed.

PR #102, opened 2018-09-02
Tests for the force field class and the force field parser

jbarnoud:


vermouth/tests/test_ffinput.py
```diff
@@ -131,16 +149,494 @@ def test_get_order_and_prefix_from_prefix(prefix, ref_prefix, ref_order):
...
+    context = Link()
+    with pytest.raises(IOError):
+        ffinput._parse_features(['A', 'B'], context, 'not-link')
+
+
+@pytest.mark.parametrize('tokens, expected', (
+    (['A'], [['A', {}]]),
+    (['A', 'B'], [['A', {}], ['B', {}]]),
+    (['+A', '{"attr": 0}', 'B'], [['+A', {'attr': 0}], ['B', {}]]),
+))
```
pckroon: These testcases look similar to a previous test?

vermouth/tests/test_forcefield.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    return vermouth.forcefield.ForceField(name='empty')
+
+
+@pytest.fixture
+def force_field_with_features(empty_force_field):
+    """
+    Build a force field with links describing features.
+
+    The force field is empty expect for the links. The links are empty expect
```
pckroon: 2 typos. expect != except

vermouth/tests/test_forcefield.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_features(force_field_with_features):
+    """
+    Test that link features are found by the force field.
+    """
+    features = (
+        ('present', True),
+        ('not present', False),
+    )
```
pckroon: unused variable?

vermouth/ffinput.py
```diff
@@ -460,10 +527,9 @@ def _treat_atom_prefix(reference, attributes):
 
 def _treat_link_interaction_atoms(atoms, context, section):
     for reference, attributes in atoms:
-        if hasattr(context, '_apply_to_all_nodes'):
```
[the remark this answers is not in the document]
jbarnoud: The attribute is now part of the Block class definition.

vermouth/forcefield.py
```diff
@@ -166,4 +173,34 @@ def iter_force_field_files(directory, extensions=FORCE_FIELD_PARSERS.keys()):
...
+    # force fields and keep a cache of them. A better solution would only parse
+    # the requested force field! There would still be a need to cache the read
+    # force fields, though. Indeed, force field comparison is based on instance
+    # identity,so we want each force field to be  singleton.
+    # TODO: Implement a better way to request a force field by name.
+    global _FORCE_FIELDS
+    try:
+        return _FORCE_FIELDS[name]
+    except KeyError:
+        _FORCE_FIELDS = find_force_fields(os.path.join(DATA_PATH, 'force_fields'))
```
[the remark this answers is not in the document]
jbarnoud: This requires some deeper thinking. I keep it for later.

vermouth/tests/test_ffinput.py
```diff
@@ -131,16 +149,494 @@ def test_get_order_and_prefix_from_prefix(prefix, ref_prefix, ref_order):
...
+    """
+    Test _get_atoms works as expected.
+    """
+    tokens = collections.deque(tokens)
+    remaining = collections.deque(tokens)
+    found = ffinput._get_atoms(tokens, natoms)
+    assert found == expected
+    # Make sure '--' is consumed if needed
+    if tokens:
+        assert not tokens[0] == '--'
```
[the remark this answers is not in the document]
jbarnoud: You may have a weird case like `A B -- C D -- E F` in which case only the first `--` should be consumed.

vermouth/tests/test_ffinput.py
```diff
@@ -131,16 +149,494 @@ def test_get_order_and_prefix_from_prefix(prefix, ref_prefix, ref_order):
...
+        {
+            'A': {
+                'atomname': 'A', 'order': 0,
+                'exist': 'hello', 'new': 'world',
+                'attr': 'plop',
+            },
+            'B': {'atomname': 'B', 'order': 0, 'attr': 'plop'},
+        },
+    ),
+))
```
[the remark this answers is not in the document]
jbarnoud: I added the multiple new attributes and disregarded the rest as asked in a comment below :wink:

vermouth/forcefield.py
```diff
@@ -57,9 +57,6 @@ def __init__(self, directory=None, name=None):
             self.name = os.path.basename(directory)
         if name is not None:
             self.name = name
-        if directory is not None and name is not None:
-            msg = 'Only one of name or directory should be provided. Not both.'
-            raise TypeError(msg)
```
pckroon: Why not both? Which takes priority?

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1019 @@
...
+
+    @staticmethod
+    def test_atoms():
+        lines = """
+        [ moleculetype ]
+        GLY 1
+
+        [ atoms ]
+          1 P4 1 ALA BB 1
+        2 P3\t1\tALA   SC1 2 -3 7
```
pckroon: indent change?

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1019 @@
...
+        4 P2 1 ALA SC3 3
+        """
+        lines = textwrap.dedent(lines) + textwrap.dedent(interaction_lines)
+        lines = lines.splitlines()
+
+        ff = vermouth.forcefield.ForceField(name='test_ff')
+        vermouth.ffinput.read_ff(lines, ff)
+        block = ff.blocks['GLY']
+
+        assert set(block.edges) == set(edges)
```
pckroon: Edges are symmetric. In this case though, `{("A", "B"),} != {("B", "A")}`

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1019 @@
...
+        BB <SC1
+        """,
+        """
+        [ bonds ]
+        BB >SC1
+        """,
+        # Not enough atoms for the interaction
+        """
+        [ bonds ]
+        BB
```
pckroon: Duplicate testcase

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1019 @@
...
+        [ atoms ]
+        1 P4 1 ALA BB 1
+        ; The following atoms have names that should not be valid
+        2 P4 2 TRP +SC1 1
+        3 P4 2 TRP -SC1 1
+        4 P4 2 TRP <SC1 1
+        5 P4 2 TRP >SC1 1
+        """
+        lines = textwrap.dedent(lines) + textwrap.dedent(interaction_lines)
+        lines = lines.splitlines()
```
pckroon: Splitlines first, add after. Otherwise you need to be very very sure the first ends in a newline character.

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1019 @@
...
+            [bonds]
+            BB +SC1 1 0.3 100
+            """,
+            (
+                ('BB', {'atomname': 'BB', 'order': 0}),
+                ('+SC1', {'atomname': 'SC1', 'order': 1}),
+            ),
+            {'bonds': [
+                vermouth.molecule.Interaction(
+                    atoms=['BB', '+SC1'],
```
pckroon: Is it desirable it refers to atoms including the prefix?

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1019 @@
...
+                    meta={},
+                ),
+            ]},
+            (('BB', 'SC1'),),
+        ),
+        (  # Interaction atoms have attributes
+            """
+            [link]
+            [bonds]
+            BB {"custom": 1, "plop": "toto"} SC1 {"toto": "tata"} 1 0.3 100
```
pckroon: Is this desirable? Isn't it better to force a separate [atoms] section in these cases?

vermouth/forcefield.py
jbarnoud: It use to prevent using both; by removing this test, "name" takes priority. I need to update the documentation.

vermouth/tests/test_ff_files.py
jbarnoud: This is on purpose, the point is to make sure that trailing spaces at the beginning of the line do not cause an error.

vermouth/tests/test_ff_files.py
jbarnoud: Yes, and it breaks the tests on python 3.4 and 3.5 :cry: I need to compare sets of frozensets.

vermouth/tests/test_ff_files.py
jbarnoud: Good catch.

vermouth/tests/test_ff_files.py
jbarnoud: I am not fully convinced about it, but it is what we have been doing so far. It causes some weirdness is some edge cases (not errors, just possible traps when writing a link), but it assures that a key always refers to the same node and it makes some matching easier latter on.

Anyway, changing that is beyond the scope of the PR.

vermouth/tests/test_ff_files.py
jbarnoud: It would make writing links super über really completely unreasonably tedious. I know: I tried.

jbarnoud:
The `[ variables ]` section creates force field wide variables. They are used for instance to specify the default bond type for the elastic network, or if the mapping is COM or COG based. Latter on, I'd like to use them to define what residues are proteins, what are the radius of the bead types for the bond guessing, or what function to use to guess the element.

vermouth/tests/test_ffinput.py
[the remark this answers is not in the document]
pckroon: Add it as a testcase

vermouth/tests/test_ffinput.py
pckroon: No clue, it's 6 months ago ;)

vermouth/tests/test_forcefield.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Creates a force field by name (no directory) and make sure it has the
+    expected name.
+    """
+    name = 'the_ff_name'
+    ff = vermouth.forcefield.ForceField(name=name)
+    assert ff.name == name
+
+
+def test_create_ff_from_dir(tmpdir):
```
pckroon: tmpdir is a magical pytest fixture?

jbarnoud:
>  There might be more tests that would benefit from creating an empty FF. Could you have a look?

It may indeed simplify a bunch of tests. I'll have a look, but it is low priority.

vermouth/tests/test_ff_files.py
pckroon: I guess it depends on what the node key is. I agree that it's out of scope either way.

pckroon: And then it starts to make sense, since you can make a link `{BB, +BB}`. It's kind of convenient to have the node key BB or +BB, makes finding things easier, and it can't be the atomname, since that's not unique.

jbarnoud: I think that was the initial motivation, indeed. It bites you in rare cases where you define the order explicitly instead of by prefix:

```
[ link ]
[ atoms ]
SC1 {"order": ">", "custom": "for the demo"}
[ bonds ]
BB  >SC1
```

In the example above, one may think that we created the key `SC1` in the `[ atoms ]` section. Instead, we created the `>SC1` key. It is a niche corner case, but it exists. Shit, I'll have to document it :cry:

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1256 @@
...
+                    meta={},
+                ),
+            ]},
+            (('BB', '+SC1'),),
+        ),
+        (  # Base case with prefix
+            """
+            [link]
+            [bonds]
+            ---BB SC1 1 0.3 100
```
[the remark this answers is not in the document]
jbarnoud: The spaces are significant: `-- BB` are 2 tokens, `--BB` is only one token.

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1256 @@
...
+    """
+    [ link ]
+    [ atoms ]
+    XXX {"order": "invalid"}
+    """,
+    # Order is inconsistent with the prefix
+    """
+    [ link ]
+    [ atoms ]
+    ++XXX {"order": -3}
```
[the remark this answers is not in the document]
jbarnoud: Yes, there is. Somewhere in one of the big parametrized tests.

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1256 @@
...
+    @staticmethod
+    def test_example():
+        lines = """
+        [ modification ]
+        C-ter
+        [ atoms ]
+        CA{"element": "C"}  ; the space between the name and the attributes is optionnal
+        C {"element": "C"}
+        O {"element": "O"}
+        OXT {"element": "O", "PTM_atom": true, "replace": {"atomname": null}}
```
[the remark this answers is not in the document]
jbarnoud: Only `null` valid. The others are not valid json.

vermouth/tests/test_ff_files.py
```diff
@@ -0,0 +1,1256 @@
...
+    """
+    [ moleculetype ]
+    GLY 1
+
+    [ atoms ]
+    1 P4 1 ALA BB 1
+    2 P4 1 ALA SC1 1
+
+    [ bonds ]
+    BB SC1 -- 7 8 -- 9 10
```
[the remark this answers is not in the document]
jbarnoud: The first one is a separator hat tells the parser that their is no atoms to read anymore. This one will be removed. The second one can be your weird parameter.

PR #106, opened 2018-09-12
Make PDB writer write correct CONECT records for multiple molecules

pckroon:


vermouth/pdb/pdb.py
```diff
@@ -88,11 +88,11 @@ def keyfunc(graph, node_idx):
     # molecules in a system. Probably not a good idea
     nodeidx2atomid = {}
     atomid = 1
-    for molecule in system.molecules:
+    for idx, molecule in enumerate(system.molecules):
```
jbarnoud: `idx` could be `m_idx` for consistency with the loop bellow.

jbarnoud:
Could you add a test? It does not have to be anything complicated: just a couple of molecules with 3 or 4 node each, and making sure that the written record contains the expected indices.

pckroon:
I hate it when you keep me to my own standards ;)

jbarnoud:
I merge as soon as the tests are passing.

PR #107, opened 2018-09-12
Martinize2 writes the command and version in the ITP 

jbarnoud:
Knowing what command line generated a given file is important for reproducibility, and for overall user sanity.


jbarnoud:
Seems that #110 was not enough. Out of scope for this PR, though.

PR #110, opened 2018-09-13
Reduce the problem size in test_gro hypothesis tests

jbarnoud:
Hopefully will reduce the running time.


pckroon:
Rebase and merge.

PR #112, opened 2018-09-13
Setup logging

### LINKED ISSUE #85: The error message for missing PTM should be more explicit
Telling what residues are involved and what atoms the PTM is made from would make it much easier to identify issues.

pckroon:
First setup of the logging facilities
To Do:
- [x] setup logging
- [x] Count warnings
- [x] Count warnings per type of warning
- [ ] Log sensible output for series of errors
- [x] Different output format for different loglevels: we only want the logger name on debug.

Fixes #85

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -131,8 +131,6 @@ CONSOLE_HANDLER.setFormatter(FORMATTER)
 LOGGER.addHandler(CONSOLE_HANDLER)
 LOGGER.addHandler(COUNTER)
 
-LOGGER.setLevel(logging.DEBUG)
-
 LOGGER = StyleAdapter(LOGGER)
 
 VERSION = 'martinize with vermouth {}'.format(vermouth.__version__)
@@ -391,8 +389,14 @@ def entry():
                                    'resulting file may contain "nan" '
                                    'coordinates making it unreadable by most '
                                    'softwares.'))
+    debug_group.add_argument('-v', dest='verbosity', action='count',
+                             help='Enable debug logging output. Can be given '
+                                  'multiple times.', default=0)
 
     args = parser.parse_args()
+
+    loglevels = {0: logging.INFO, 1: logging.DEBUG, 2: 5}
+    LOGGER.setLevel(loglevels[args.verbosity])
 
     known_force_fields = vermouth.forcefield.find_force_fields(
         Path(DATA_PATH) / 'force_fields'
@@ -526,4 +530,3 @@ def entry():
 
 if __name__ == '__main__':
     entry()
-    print(COUNTER.counts)
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -19,6 +19,7 @@ High level API for Martinize2
 """
 
 import argparse
+from collections import defaultdict
 import functools
 import logging
 import itertools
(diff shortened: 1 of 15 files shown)


vermouth/processors/do_mapping.py
```diff
@@ -323,12 +329,20 @@ def do_mapping(molecule, mappings, to_ff, attribute_keep=()):
...
+                           [molecule.nodes[idx] for idx in shared_atoms],
+                           [graph_out.nodes[idx] for idx in shared_out_atoms])
     # TODO: These should be turned into warnings.
-    print('double covered:', {k: len(v) for k, v in mol_to_out.items() if len(v) > 1})
-    print('uncovered:', set(molecule.nodes.keys()) - set(mol_to_out.keys()))
-    print(len(set(molecule.nodes.keys()) - set(mol_to_out.keys())))
+    LOGGER.debug('double covered: {}', {k: v for k, v in blocks_per_atom.items() if v > 1})
+    for node_idx, v in blocks_per_atom.items():
+        if v > 1:
+            LOGGER.debug('{}', molecule.nodes[node_idx]['atomname'])
```
jbarnoud: Debug messages should ideally be standalone. This loop of messages would not be.

vermouth/utils.py
```diff
@@ -34,6 +34,10 @@ def _distance(point_1, point_2):
     distance = _distance
 
 
+def format_atom_string(node):
+    return '{}-{}{}:{}'.format(node['atomid'], node['resname'], node['resid'], node['atomname'])
```
jbarnoud: Chain?

pckroon:
I think this is ready for review. If it's ok I'll rebase it against master, make sure Travis passes, and then it can be merged. I've written a bunch of sensible warnings, but I think it's wiser to add the rest as they're found by the group

vermouth/processors/do_mapping.py
```diff
@@ -315,20 +322,53 @@ def do_mapping(molecule, mappings, to_ff, attribute_keep=()):
...
-                  " have too  particles in your output and/or erroneous bonds.")
+            LOGGER.warning("You have the following atoms that are shared"
+                           " between blocks. This may mean you have too many"
+                           " particles in your output and/or erroneous bonds."
+                           " {}. They've end up in the following output atoms:"
+                           " {}.",
+                           [format_atom_string(molecule.nodes[idx]) for idx in shared_atoms],
+                           [format_atom_string(graph_out.nodes[idx]) for idx in shared_out_atoms],
+                           type='inconsistent-data')
     # TODO: These should be turned into warnings.
```
jbarnoud: These are not turned into warning. Aren't they?

vermouth/processors/repair_graph.py
```diff
@@ -84,7 +87,8 @@ def make_reference(mol):
             matches = [{v: k for k, v in match.items()} for match in matches]
         if not matches:
             # DEBUG
```
jbarnoud: The "DEBUG" comment is not needed anymore.

vermouth/tests/test_logging.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
```
jbarnoud: Pylint will be better than me at telling what is missing :smile:

bin/martinize2
```diff
@@ -33,13 +34,105 @@ from vermouth.dssp.dssp import (
...
     combine_mappings
 )
 
+# TODO Since vermouth's __init__.py does some logging (KDTree), this may or may
+# not work as intended. Investigation required.
+
+LOGGER = logging.getLogger('vermouth')
+
+
+class BipolarFormatter:
```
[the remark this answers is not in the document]
pckroon: I put these here because of the separation front-end <=> library. The LoggerAdapters are part of the library since you need (can use) them to make logging calls. The Formatter and Handler are here, because the *library* doesn't need/use them. Now, if we were to make a martinize2 module...

I'll give it some more thought.

vermouth/log_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+        self._find_kwargs()
+
+    def _find_kwargs(self):
+        logsig = inspect.signature(self.log)
+        my_kwargs = set(name for name, param in logsig.parameters.items()
+                        if param.kind not in [param.VAR_POSITIONAL, param.VAR_KEYWORD])
+        if isinstance(self._logger, logging.Logger):
+            child_kwargs = {'lvl', 'msg', 'exc_info', 'stack_info', 'extra'}
+        else:
+            if hasattr(self._logger, 'known_kwargs'):
```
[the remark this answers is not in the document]
pckroon: It's set by the `logger.setter` property. How about the following logic instead? I'm mainly concerned with other people subclassing/extending/using it.
```python3
if not hasattr(self.logger, 'known_kwargs'):
    child_kwargs = {'lvl', 'msg', ...}  # Or, get these from the _log method
else:
    child_kwargs = self.logger.known_kwargs
```

bin/martinize2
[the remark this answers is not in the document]
jbarnoud: It is indeed kind of middle ground between vermouth and martinize. Ideally, we want to empty martinize itself from as much code as possible, especially code like the logger that can easily be used by an other vermouth-based tool. We could have a `interface.py` module.

vermouth/processors/do_mapping.py
pckroon: Two of them are. The TODO can be removed.

vermouth/log_helpers.py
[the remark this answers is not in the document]
jbarnoud: I am fine with the current logic, but it requires some explaining. A comment would be nice; so would a test.

pckroon:
>Could you add the verbose option in martinize2?

Sure. But IIRC there is also already some CLI functionality in the logging module itself.

>Also, we need to figure out what to do with errors. For now there is a mix of error, exception, and critical.

Error = Bad thing happened, I *might* recover, but you can kiss your output goodbye
Exception = Error, but with traceback information
Critical = Very Bad thing happened, I died in a small bonfire

pckroon:
Ready for review, I'm done with it.

PR #114, opened 2018-09-20
Add the -ss option to martinize2

jbarnoud:
The `-ss` option is used to manually set the secondary structures of protein instead of using DSSP. The feature is present in the legacy martinize, and has been requested several times by members of the group (Maria and Lorenzo among others).

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -249,13 +249,18 @@ def entry():
     posres_group.add_argument('-pf', dest='posres_fc', type=float, default=1000,
                               help='Position restraints force constant in kJ/mol/nm^2')
     secstruct_group = parser.add_argument_group('Secondary structure handling')
-    secstruct_group.add_argument('-dssp', nargs='?', const='dssp',
-                                 help='DSSP executable for determining structure')
+    secstruct_exclusion = secstruct_group.add_mutually_exclusive_group()
+    secstruct_exclusion.add_argument('-dssp', nargs='?', const='dssp',
+                                     help='DSSP executable for determining structure')
+    secstruct_exclusion.add_argument('-ss', dest='ss', type=lambda x: x.upper(),
+                                     metavar='SEQUENCE',
+                                     help=('Manually set the secondary '
+                                           'structure of the proteins.'))
+    secstruct_exclusion.add_argument('-collagen', action='store_true', default=False,
+                                     help='Use collagen parameters')
     secstruct_group.add_argument('-ed', dest='extdih', action='store_true', default=False,
                                  help=('Use dihedrals for extended regions '
                                        'rather than elastic bonds'))
-    secstruct_group.add_argument('-collagen', action='store_true', default=False,
-                                 help='Use collagen parameters')
 
     rb_group = parser.add_argument_group('Protein elastic network')
     rb_group.add_argument('-elastic', action='store_true', default=False,
@@ -351,6 +356,10 @@ def entry():
     target_ff = known_force_fields[args.to_ff]
     if args.dssp is not None:
         AnnotateDSSP(executable=args.dssp, savedir='.').run_system(system)
+        AnnotateMartiniSecondaryStructures().run_system(system)
+    elif args.ss is not None:
+        AnnotateResidues(attribute='secstruct', sequence=args.ss,
+                         molecule_selector=selectors.is_protein).run_system(system)
         AnnotateMartiniSecondaryStructures().run_system(system)
     elif args.collagen:
         if not target_ff.has_feature('collagen'):
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -29,6 +29,7 @@ import sys
 import vermouth
 from vermouth.forcefield import FORCE_FIELDS
 from vermouth import DATA_PATH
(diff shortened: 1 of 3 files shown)


jbarnoud:
@pckroon This is ready for review. Codecov seems lost in its reference coverage, though.

bin/martinize2
```diff
@@ -249,13 +250,18 @@ def entry():
     posres_group.add_argument('-pf', dest='posres_fc', type=float, default=1000,
                               help='Position restraints force constant in kJ/mol/nm^2')
     secstruct_group = parser.add_argument_group('Secondary structure handling')
-    secstruct_group.add_argument('-dssp', nargs='?', const='dssp',
-                                 help='DSSP executable for determining structure')
+    secstruct_exclusion = secstruct_group.add_mutually_exclusive_group()
+    secstruct_exclusion.add_argument('-dssp', nargs='?', const='dssp',
+                                     help='DSSP executable for determining structure')
+    secstruct_exclusion.add_argument('-ss', dest='ss', type=lambda x: x.upper(),
```
pckroon: `type=str.upper`

vermouth/dssp/dssp.py
```diff
@@ -462,6 +487,73 @@ def __init__(self, attribute, sequence,
...
+        ----------
+        system: vermouth.system.System
+
+        Returns
+        -------
+        vermouth.system.System
+        """
+        # Test and adjust the length of the sequence. There are 3 valid scenarios:
+        # * the length of the sequence matchesthe number of residues in the
+        #   selection;
```
pckroon: typo

vermouth/dssp/dssp.py
```diff
@@ -462,6 +487,73 @@ def __init__(self, attribute, sequence,
...
+            for molecule in system.molecules
+            if self.molecule_selector(molecule)
+        ]
+        if self.sequence and not molecule_lengths:
+            raise ValueError('There is no molecule to which '
+                             'to apply the sequence.')
+        if (molecule_lengths
+                and len(self.sequence) == molecule_lengths[0]
+                and utils.are_all_equal(molecule_lengths)):
+            sequence = list(self.sequence) * len(molecule_lengths)
```
pckroon: Can `self.sequence` be a string a this point? In that case `list(self.sequence)` will do something else than `[self.sequence]`

vermouth/dssp/dssp.py
```diff
@@ -462,6 +487,73 @@ def __init__(self, attribute, sequence,
...
+        elif len(self.sequence) == 1:
+            sequence = list(self.sequence) * sum(molecule_lengths)
```
pckroon: Same as above

vermouth/tests/test_dssp.py
```diff
@@ -30,6 +30,272 @@
                       'SCSSEEEEEESTTHHHHTTTSSHHHHHHHHHHHHHTTCCEEEEEC')
 
 
+# TODO: The code is very repetitive. There may be a way to refactor it with
+# clever use of parametrize and fixtures.
```
pckroon: I think you can `@parametrize` your `@fixtures`. And it should magically work.
https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures

vermouth/tests/test_dssp.py
```diff
@@ -30,6 +30,272 @@
...
+        expected = list(itertools.chain(
+            *([element] * 3 for element in sequence)
+        ))
+        expected = expected * 3
+        processor = dssp.AnnotateResidues('test', sequence)
+        processor.run_system(multi_mol_system_regular)
+        found = self.sequence_from_system(multi_mol_system_regular, 'test')
+        assert found == expected
+
+    def test_sigle_molecules_cycle_one(self, single_mol_system):
```
pckroon: typo

vermouth/tests/test_dssp.py
```diff
@@ -30,6 +30,272 @@
...
+        """
+        There are no molecules, but the sequence is empty.
+        """
+        system = vermouth.system.System()
+        sequence = ''
+        processor = dssp.AnnotateResidues('test', sequence)
+        try:
+            processor.run_system(system)
+        except ValueError:
+            pytest.fail('Should not have raised a ValueError.')
```
pckroon: Pointless try/except? A test that raises an error also fails.

jbarnoud: A test that raises an exception appears as errored, not failed.

vermouth/tests/test_dssp.py
jbarnoud: You can parametrize fixtures in some capacity. But here I would need to think about a clever architectures of the tests that does not obfuscate them.

vermouth/dssp/dssp.py
jbarnoud: `self.sequence` can be any kind of sequence. Using `list` normalizes the behaviour. is `self.sequence` is `"ABCD"` and `len(molecule_lengths)` is 2, then `self.sequence` will become `['A', 'B', 'C', 'D', 'A', 'B', 'C', 'D']` which is the expected behaviour.

PR #119, opened 2018-09-25
Update the vdw radii to the ones from Bondi

jbarnoud:
Use the radii from Bondi except for the hydrogen. These are the same radii used by VMD. In practice I git better graph on the few protein I tested.

### DIFF
--- a/vermouth/processors/make_bonds.py
+++ b/vermouth/processors/make_bonds.py
@@ -28,7 +28,9 @@ from .processor import Processor
 from ..utils import distance
 
 # Van der Waals radii from A. Bondi, J. Phys. Chem., 68, 441-452, 1964.
+# https://doi.org/10.1021/j100785a001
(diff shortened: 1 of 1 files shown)


vermouth/processors/make_bonds.py
```diff
@@ -27,7 +27,29 @@
 from .processor import Processor
 from ..utils import distance
 
-COVALENT_RADII = {'H': 0.031, 'C': 0.076, 'N': 0.071, 'O': 0.066, 'S': 0.105}
+# Van der Waals radii from A. Bondi, J. Phys. Chem., 68, 441-452, 1964.
+# For hydrogen, we use R.S. Rowland & R. Taylor, J.Phys.Chem., 100, 7384-7391, 1996.
```
pckroon: add DOIs

jbarnoud: done

PR #121, opened 2018-09-26
Group nucleic acid strands

jbarnoud:
Generalize the code for cystein bridges so it can be used to connect DNA strands by adding edges that approximate the hydrogen bonds.

Also add a few tests.


vermouth/tests/test_selectors.py
```diff
@@ -0,0 +1,148 @@
...
+
+import pytest
+
+import vermouth
+from vermouth.pdb.pdb import read_pdb
+from vermouth.tests.datafiles import (
+    PDB_PROTEIN,
+    PDB_NOT_PROTEIN,
+    PDB_PARTIALLY_PROTEIN,
+)
```
pckroon: We can also consider making `vermouth.tests.datafiles` create `fixtures`.

vermouth/tests/test_selectors.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    assert vermouth.selectors.select_backbone(node) == expected
+
+
+@pytest.mark.parametrize('node, attribute, values, expected', (
+    ({}, 'name', ['something', 'other'], False),
+    ({'name': 'nothing'}, 'name', ['something', 'other'], False),
+    ({'name': 'something'}, 'name', ['something', 'other'], True),
+    ({'name': 'nothing', 'plop': 'other'}, 'name', ['something', 'other'], False),
+    ({'name': 'nothing'}, 'name', [], False),
```
pckroon: `'name': 'other'`
`'name': 'somethingother'`

vermouth/tests/test_selectors.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def test_filter_minimal():
+    """
+    Test that :func:`vermouth.selectors.filter_minimal` works as expected.
+    """
+    # Build a molecule that has all even atoms with no positions.
+    molecule = read_pdb(str(PDB_PROTEIN))
+    for atom in list(molecule.nodes.values())[::2]:
+        atom['position'] = None
```
pckroon: del atom['position']?

vermouth/tests/test_tune_cystein_bridges.py
```diff
@@ -0,0 +1,112 @@
...
+
+import copy
+
+import pytest
+
+import vermouth
+from vermouth.pdb.pdb import read_pdb
+from vermouth.processors import tune_cystein_bridges
+from vermouth.utils import distance
+from .test_edge_tuning import simple_protein  # pylint: disable=unused-import
```
pckroon: Do you actually need to import it, or is pytest clever enough?

pckroon:
In general: make sure the types in the docstrings are as broad as possible, and if you add edges based on a distance criterion, add a distance attribute to the edge.

vermouth/tests/test_edge_tuning.py
```diff
@@ -0,0 +1,553 @@
...
+        for molecule in molecules
+        for node in molecule.nodes.values()
+    )
+    for idx, (node, coords) in enumerate(zip(iter_nodes, coordinate_array)):
+        node['coords'] = coords
+        node['serial'] = idx
+    return molecules
+
+
+class TestPruneEdgesBetweenSelection:
```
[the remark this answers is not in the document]
jbarnoud: The classes are there to group tests together and avoid having lost fixtures everywhere. Without the classes you get a litany of seemingly unrelated test functions. It could also latter on allow for simpler function names.

vermouth/processors/tune_cystein_bridges.py
```diff
@@ -124,260 +48,54 @@ def remove_cystein_bridge_edges(molecule, template=UNIVERSAL_BRIDGE_TEMPLATE):
...
-    def run_system(self, system):
-        system.molecules = add_cystein_bridge_threshold(
-            system.molecules, self.threshold,
-            template=self.template,
-            attribute=self.attribute
+        super().__init__(
+            threshold,
+            templates_from=template,
+            templates_to=template,
+            attribute=attribute,
```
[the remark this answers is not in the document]
jbarnoud: The 2 neighboring CYS may be dealt with by the link. Though, I'd say if they are that close, they may be bounded.

Removing the cys bridge is not a link problem, I think. The link for a cys bridge should apply everywhere it matches; and the match should be based on the graph topology only and not the coordinates. Here, we deal with the spatial definition of a cys bridge. It depends on the user and not so much on the data.

vermouth/tests/test_selectors.py
jbarnoud: What are these testing?

vermouth/tests/test_selectors.py
jbarnoud: It should be equivalent in the way it is handled.

vermouth/processors/tune_cystein_bridges.py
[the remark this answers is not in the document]
pckroon: Good point

vermouth/tests/test_selectors.py
pckroon: `({'name': 'other'}, 'name', ['something', 'other'], False),` To make sure any of the allowed names works.
`({'name': 'somethingother'}, 'name', ['something', 'other'], False),` To make sure the equality isn't actually doing membership testing

vermouth/tests/test_selectors.py
pckroon: You don't get to use the word "should" when talking about tests

jbarnoud:
I think we should not overdo the types in the docstrings. They are there as an indication for the user. If you want such detailed type indications, it should go into a proper stub for type hinting so that:
* we can run mypy
* we can define more specific types

PR #122, opened 2018-09-27
Minor performance optimizations

### LINKED ISSUE #60: `Molecule.subgraph` does not preserve the atom order with networkx 2.1

### LINKED ISSUE #61: Molecule.copy does not copy the interactions

jbarnoud:
This PR introduces some minor performance optimizations to reduce the slow portion of the code that are not isomorphism related. The commit messages describe the changes detail.

The performance gain is limited because this PR also rewrites `Molecule.subgraph` and `Molecule.copy` to fix #60 and to fix #61.

On a small example, this branch reduces the run time from 10.321 seconds to 9.148 seconds; so a ~10% speed gain. On a larger example, the run time is reduced from 1m42 to 1m23; so a bit less than 20% gain.

The mapping step is the slowest in both examples and represents about half of the run time, followed by the block and link steps.

Among the non-isomorphism related slow operations, querying the adjacency matrix is surprisingly slow. We should limit as much as possible how often we look for neighbors.

### DIFF
--- a/vermouth/molecule.py
+++ b/vermouth/molecule.py
@@ -377,12 +377,12 @@ class Molecule(nx.Graph):
 
         nodes = set(nodes)
 
-        edges_to_add = [
-            (node, node2)
-            for node in nodes
-            for node2 in set(self[node]) & nodes
-        ]
-        subgraph.add_edges_from(edges_to_add)
+        #edges_to_add = [
+        #    (node, node2)
+        #    for node in nodes
+        #    for node2 in set(self[node]) & nodes
+        #]
+        subgraph.add_edges_from(self.edges_between(nodes, nodes, data=True))
 
         for interaction_type, interactions in self.interactions.items():
             for interaction in interactions:
@@ -646,7 +646,7 @@ class Molecule(nx.Graph):
         residue_graph = graph_utils.make_residue_graph(self)
         return (tuple(residue_graph.nodes[res]['graph'].nodes) for res in residue_graph.nodes)
 
-    def edges_between(self, n_bunch1, n_bunch2):
+    def edges_between(self, n_bunch1, n_bunch2, data=False):
         """
         Returns all edges in this molecule between nodes in `n_bunch1` and
         `n_bunch2`.
@@ -664,14 +664,15 @@ class Molecule(nx.Graph):
             A list of tuples of edges in this molecule. The first element of
             the tuple will be in `n_bunch1`, the second element in `n_bunch2`.
         """
-        result = []
         set_1 = set(n_bunch1)
         set_2 = set(n_bunch2)
         for node1 in set_1:
             cross = set_2 & set(self[node1])
             for node2 in cross:
-                result.append((node1, node2))
-        return result
+                if not data:
+                    yield (node1, node2)
+                else:
+                    yield (node1, node2, self.edges[node1, node2])
 
 
 class Block(Molecule):
--- a/vermouth/molecule.py
+++ b/vermouth/molecule.py
@@ -348,30 +348,45 @@ class Molecule(nx.Graph):
             node_attr = self.node[node]
             yield node, node_attr
 
-    def copy(self, as_view=False):
-        '''
+    def copy(self):
+        """
         Creates a copy of the molecule.
 
-        See Also
-        --------
-        :meth:`networkx.Graph.copy`
-        '''
-        copy = super().copy(as_view)
-        if not as_view:
-            copy = self.__class__(copy)
-        copy._force_field = self.force_field
-        copy.meta = self.meta.copy()
-        return copy
+        Returns
+        -------
+        Molecule
+        """
+        return self.subgraph(self.nodes)
 
-    def subgraph(self, *args, **kwargs):
-        '''
+    def subgraph(self, nodes):
+        """
         Creates a subgraph from the molecule.
 
-        See Also
-        --------
-        :meth:`networkx.Graph.subgraph`
-        '''
-        return self.__class__(super().subgraph(*args, **kwargs))
+
+        Returns
+        -------
+        Molecule
+        """
+        subgraph = self.__class__()
+        subgraph.meta = copy.copy(self.meta)
+        subgraph._force_field = self._force_field
+        subgraph.nrexcl = self.nrexcl
+
+        node_copies = [(node, copy.copy(self.nodes[node])) for node in nodes]
+        subgraph.add_nodes_from(node_copies)
+
+        nodes = set(nodes)
+
+        edges_to_add = [edge for edge in self.edges
+                        if edge[0] in nodes and edge[1] in nodes]
+        subgraph.add_edges_from(edges_to_add)
+
+        for interaction_type, interactions in self.interactions.items():
+            for interaction in interactions:
+                if all(atom in nodes for atom in interaction.atoms):
+                    subgraph.interactions[interaction_type].append(interaction)
+
+        return subgraph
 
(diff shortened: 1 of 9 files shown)


vermouth/molecule.py
```diff
@@ -348,30 +348,48 @@ def atoms(self):
             node_attr = self.node[node]
             yield node, node_attr
 
-    def copy(self, as_view=False):
-        '''
+    def copy(self):
```
pckroon: This changes the API from the NX one

jbarnoud:
I will have to squash my travis commits. There will be a rebase and a forced push in a few minutes.

vermouth/molecule.py
```diff
@@ -348,30 +348,48 @@ def atoms(self):
...
+        subgraph.add_nodes_from(node_copies)
+
+        nodes = set(nodes)
+
+        edges_to_add = [
+            (node, node2)
+            for node in nodes
+            for node2 in set(self[node]) & nodes
+        ]
+        subgraph.add_edges_from(edges_to_add)
```
pckroon: `subgraph.add_edges_from(self.edges_between(nodes, nodes))`? And this might loose edge data.

vermouth/molecule.py
```diff
@@ -646,9 +664,14 @@ def edges_between(self, n_bunch1, n_bunch2):
...
-                for node1, node2 in itertools.product(n_bunch1, n_bunch2)
-                if self.has_edge(node1, node2)]
+        result = []
+        set_1 = set(n_bunch1)
+        set_2 = set(n_bunch2)
+        for node1 in set_1:
+            cross = set_2 & set(self[node1])
+            for node2 in cross:
+                result.append((node1, node2))
+        return result
```
pckroon: Should this be an iterator? And possibly accept a `data=False` kwarg, which will make it yield `(n1, n2, data)`?

vermouth/processors/make_bonds.py
```diff
@@ -62,6 +62,10 @@ def bonds_from_distance(system, fudge=1.1):
     The possible distance between nodes is determined by values in
     `VDW_RADII`.
 
+    Notes
+    -----
+    Elements that are not in `VMD_RADII` do not make bonds.
```
pckroon: Typo

vermouth/processors/make_bonds.py
```diff
@@ -76,25 +80,36 @@ def bonds_from_distance(system, fudge=1.1):
         certain distance from each other. It is probably disconnected.
     """
     system = nx.compose_all(system.molecules)
-    idx_to_nodenum = {idx: n for idx, n in enumerate(system)}
+    idx_to_nodenum = {
+        idx: n
+        for idx, n in enumerate(
+                subn
+                for subn in system
+                if system.nodes[subn].get('element') in VDW_RADII
```
pckroon: Could use a comment

vermouth/processors/make_bonds.py
```diff
@@ -76,25 +80,36 @@ def bonds_from_distance(system, fudge=1.1):
...
+        )
+    }
     max_dist = max(VDW_RADII.get(node.get('element'), 0.2) for node in system.nodes.values())
```
pckroon: Why `0.2` and not, say, `max(VDW_RADII.values())`? And if elements not in VDW_RADII don't make bonds, this seems rather pointless anyway. Just take the largest radius you find for the elements we know: those in `idx_to_nodenum`

jbarnoud:
Ready for review

vermouth/processors/make_bonds.py
```diff
@@ -76,25 +80,36 @@ def bonds_from_distance(system, fudge=1.1):
...
-    positions = np.array([system.node[n]['position'] for n in system], dtype=float)
+    positions = np.array([
+        node['position']
+        for node in system.nodes.values()
+        if node.get('element') in VDW_RADII
```
pckroon: Get them from `idx_to_nodenum`

vermouth/redistributed/kdtree.py
```diff
@@ -694,6 +700,38 @@ def traverse_no_checking(node1, node2):
                           other.tree, Rectangle(other.maxes, other.mins))
         return results
 
+    def sparse_distance_matrix(self, other, max_distance, p=2.0):
+        """
+        Emulate :meth:`scipy.spacial.KDtree.sparse_distance_matrix` without scipy.
```
pckroon:
```
:meth:`scipy.spacial.cKDtree.sparse_distance_matrix`
```

vermouth/tests/test_molecule.py
```diff
@@ -132,3 +120,83 @@ def test_link_predicate_match():
...
+def test_edges_between(edges_between_molecule, edges_between_selections,
+                       bunch1, bunch2, expected):
+    """
+    Test :meth:`vermouth.molecule.Molecule.edges_between`.
+    """
+    selection_1 = edges_between_selections[bunch1]
+    selection_2 = edges_between_selections[bunch2]
+    found = edges_between_molecule.edges_between(selection_1, selection_2)
+    sorted_found = sorted(sorted(edge) for edge in found)
+    sorted_expected = sorted(sorted(edge) for edge in expected)
```
pckroon: I'd make sets of frozensets, but this works as well.

vermouth/tests/test_molecule.py
```diff
@@ -132,3 +120,83 @@ def test_link_predicate_match():
...
+    See Also
+    --------
+    test_subgraph_base
+        This test deals with a larger graph, but no metadata; while the graph
+        in :func:`test_subgraph_base` uses a very small and simple selection
+        but the graph has metadata.
+    """
+    subgraph = edges_between_molecule.subgraph(edges_between_selections[selidx])
+    sorted_found = sorted(sorted(edge) for edge in subgraph.edges)
+    sorted_expected = sorted(sorted(edge) for edge in expected)
```
pckroon: Same as above

vermouth/molecule.py
jbarnoud: The iterator is a good idea. for the `data` argument, do we need it? I'd like to avoid feature creeping the `Molecule` class.

jbarnoud: Now I see the point...

vermouth/processors/make_bonds.py
jbarnoud: From my quick test, it makes the code a bit less readable. It means `idx_to_nodenum` must be an `OrderedDict` because of python < 3.6; and it leads to many more indirections to access the elements.

vermouth/tests/test_molecule.py
jbarnoud: Does not change much.

vermouth/molecule.py
pckroon: I'm going to make your day worse :)
In NX, the data keyword takes either a bool and behaves as it does now, or an attribute key (and a default value?); in which case it doesn't give back the the whole dict, but just the requested element. I'm happy to keep this as is untill we need it though.

vermouth/processors/make_bonds.py
pckroon: Ok, good point about the ordereddict.

vermouth/molecule.py
jbarnoud: :cry: It is doable. Though, the more feature we throw here, the slower the function will be; especially since it is within the loop.

pckroon: Let's just leave it as is for now. And remember to come back here once we need the functionality.

PR #149, opened 2018-10-04
Allow do_average_bead to use different centerings

### LINKED ISSUE #147: Bead placement should allow center of geometry
Beads are placed at the center of mass of the underlying atoms. In martini 3, the center of geometry is generally preferred.  The `do_average_bead` function should be able to compute any center, and the `DoAverageBead` processor should choose what center to use based on a force field-defined variable.

jbarnoud:
Fixes #147 
Note that this solves my need. Though I wonder if the choice of the averaging method should not be at the mapping level rather than the force field one.

### DIFF
--- a/vermouth/tests/test_average_beads.py
+++ b/vermouth/tests/test_average_beads.py
@@ -1,0 +1,133 @@
+# Copyright 2018 University of Groningen
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Tests for the :mod:`vermouth.processors.average_beads` module.
+"""
+
+import pytest
+import networkx as nx
+import numpy as np
+
+from vermouth.processors import average_beads
+
+
+@pytest.fixture
+def mol_with_subgraph():
+    """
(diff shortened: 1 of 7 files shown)


vermouth/tests/test_average_beads.py
```diff
@@ -0,0 +1,133 @@
...
+            mol_with_subgraph, ignore_missing_graphs=False, weight=weight,
+        )
+
+
+def test_shoot_graph(mol_with_subgraph):
+    """
+    Test that :func:`average_beads.do_average_bead` fails if a subgraph is missing.
+    """
+    del mol_with_subgraph.nodes[1]['graph']
+    with pytest.raises(ValueError):
```
pckroon: Regarding reconstructing beads/atoms, missing sidechains, and generating coordinates; I'm not sure this is a good idea.

pckroon:
Also, what's up with Travis's Py3.4?

vermouth/tests/test_average_beads.py
jbarnoud: This behavior did not change: if there is no underlying graph and the `ignore_missing_graphs` keyword argument of `do_average_bead` is `False`, then you get a `ValueError`.

jbarnoud:
> A weight attribute per forcefield is a good idea, but you need to be able to overwrite that per mapping.

We need to have a more potent mapping object that can have its own variables.

> Also, what's up with Travis's Py3.4?

I looked at it yesterday. I did not see anything. Travis still says it supports 3.4.

pckroon:
As for Travis:
https://github.com/travis-ci/travis-ci/issues/10185

pckroon:
Travis is up and running again, but the docs are failing for this PR

pckroon:
For this PR: do you also want to implement the more potent mapping objects that can override the FF mapping policy, or shall we leave that as a later feature? If so, could you make the issue?

jbarnoud:
I just created #153. Implementing this is out of scope of this issue. Also we need COG to be in rather quickly as it is the preferred way in martini3.

PR #150, opened 2018-10-06
Add vim highlighting to the repo

jbarnoud:
Colour in VIM for the force field files. At some point I'll write a rule file for the mapping files.


pckroon:
Slightly out of scope maybe, but ok.

PR #155, opened 2018-10-26
Implement ISMAGS

pckroon:
Initial working implementation of ISMAGS. Still WIP for code cleaning, and docstringing.


jbarnoud:
There are many broken tests. Is that expected at that stage?

pckroon:
I just pushed a testcase that I know fails (found by hypothesis on my local machine). The problem is that not all symmetries are found, and the core of the problem lies in `_refine_node_partitions` *not* switching two nodes around, causing it to miss out on a permutation (confirmed by an if-statement that changes `bottom_partitions` in that specific case). The original paper is mute on this. I'm not sure whether my implementation is flawed or our tests are better.

jbarnoud:
> The original paper is mute on this. I'm not sure whether my implementation is flawed or our tests are better.

It would probably be worth testing this kind of use cases against the reference implementation.

jbarnoud:
I am reading the paper, something I did not see mentioned yet, and that I expect could be an issue, is the non-transitive nature of our node matching ("colors" are "fuzzy"). There are two levels in that potential issue: (1) the subset of tested attributes is different between two subgraph nodes, and  (2) the subgraph node uses link predicates.

Do you have test cases for these crucial features of vermouth that could break with ISMAGS?

pckroon:
Not yet. I also realize they can be breaking. I'm still thinking for a way around the transitivity. I am aware that it is a hard requirement at the moment.
1) is not necessarily a problem I think, but do you have an example?
2) is going to be much harder. At the moment these are also not included in the isomorphism, but instead we filter afterwards. 
The transitivity is required for the symmetry analysis. If we accept a "best effort" on that front (ignoring attributes that are not transitive, or just marking those as unequal), we can switch to a different method of equality checking during the isomorphism generation (`_map_nodes`).

Afterthought: The link/bond order predicates within a link *are* transitive (`'>'` == `'>'`). They're not when applied on the molecule though, but that's also exactly where we can get away from that.

jbarnoud:
Non transitive graph matching is absolutely required to match links, and if I recall correctly, it most checks are done during the semantic feasibility check. PTM can be fuzzy as well, but to a lesser extent.

Block matching should be fine, though. From what I see, there is no fuzziness allowed in the semantic feasibility check, and I cannot see a scenario where it would be a good idea to have it (or, to be exact I do see a couple of scenarii, but I am not sure they are good ideas).

So in the current state, it would at least be useful for RepairGraph and DoMapping, but it cannot be used for DoLink.

pckroon:
As I read the link code link node equality is transitive (given that we can compare LinkPredicates for equality). The transitivity breaks between link and molecule, bu that's also where it is not strictly required. I'll get on the ISMAGS side of the code soon.

jbarnoud:
The semantic feasibility of link matching is based on `molecule.attributes_match` which is not transitive. Assuming 2 nodes, A and B, with keys `{a, b, c}` and `{a, b, c, d}`, respectively, and assuming that the values are equals so only the keys matter, then `attribute_match(A, B)` will be `False` because `d` that is in B (template) is not in A, but  `attribute_match(B, A)` will be `True` because all the attributes from the template (A) are found in B. This is even without link predicates.

Link predicates can be compared (though, not in master yet). But a direct comparison of link predicates in only relevant for link-link comparisons. Link-molecule requires to actually call the predicate, then we can encounter situations where LA == MA, LB == MA, but LA != LB (where LX is a link node, and MX is a molecule node).

I do not know how any of this actually is relevant for ISMAGS, but there should be tests to account for it.

pckroon:
We agree that molecule-link comparison is *not* transitive. By argument is that link-link comparison is/can be. Since the link-link comparison is used for symmetry determination, `{a, b, c} == {a, b, c, d}` should compare `False`. Then when looking at molecule-link equality, `molecule.attributes_match` should be used.

I don't really see how to cram this into tests though.

jbarnoud:
To have link predicates comparable, you would need to add the following method in `molecule.LinkPredicate`:

```python
def __eq__(self, other):
    return (other.__class__ == self.__class__) and (other.value == self.value)
```

I'll think about specific test cases, though having instances of `LinkPredicate` as a possible output for the strategy you use to generate node attribute values would be a first start. It would allow to make sure you use the correct comparison method in the right place.

pckroon:
I've put ISMAGS in the repair_graph and canonicalize_PTM processors, and running martinize2 on my peptide thingy now feels much faster :D

One problem though. It doesn't select nodes with the right atom name as reliably as the old code, and the tests do expect this (and it is desirable behaviour). Sorting the node keys by atomname helps (a lot), but it will still fail if there are nodes missing (and it will happily select 'H' before 'HN'). This can be fixed by switching resdiue and reference in repair_graph (it makes the tests ~30 times as fast, and working), but in that case the actual program becomes slow as molasses. 
It might be that somewhere (in the tests, hopefully) the arguments are switched around. Investigation ongoing.

pckroon:
Ok, selecting the correct nodes based on atomnames works again, so that's good. There's still the funny thing about assuming residue >= reference that makes the tests way faster, but the actual program slower. I don't know why that is.

In addition, I think it only makes sense to use ISMAGS in repair_graph, since in all other cases (PTM, mapping, links) the graphs are asymmetric since matching is done on atomname instead of element.

Hypothesis found another breaking cornercase for ISMAGS. I'll investigate/fix that one, and I think it's finished then.

pckroon:
The failing test is in something I haven't touched, and therefor out of scope for this PR, which is more than big enough already.

pckroon:
I think I'm done with this for now. **DO NOT MERGE** it yet though. After your review I'm going to submit my implementation to networkx, and also get a review there. Then there'll be some back and forth until the whole world is happy with it, and *then* we'll merge.

jbarnoud:
Good! Any idea why the actual program get slower?

pckroon:
Ah yes. That's still super weird to me. Switching it around makes the actual program *so* slow that it doesn't even terminate. So there's definitely something fishy going on there. We'll investigate once your back here.

jbarnoud:
For the record, `RepairGraph` on hemoglobin takes more than 20 minutes. Likely because of the heme. The main culprit is `_map_node`. Here is the line profile: 
[line_prof.txt](https://github.com/marrink-lab/vermouth-martinize/files/2607965/line_prof.txt).

Combined with cProfile, it seems that `lambda n: min(new_candidates[n], key=len)` and `if sgn2 not in sgn_neighbours:` are the main offenders.

jbarnoud:
From further inspection (meaning I read two lines above), the two operations involve querying the adjacency matrix, which is known to be slow in networkx. Some sort of caching of the adjacency matrix, would make a huge difference here.

pckroon:
Thanks for the profiling data. Can you try running it with just heme instead of the full haemoglobin just to be sure? Also, let's do the general review first, and optimisations later.
In addition, a note to self: Using the self isomorphism as profiling case is *wrong*, unless you shuffle and relabel the node IDs in one of the two graphs. Otherwise `_map_nodes` is deceivingly fast: everything will work the first try.

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+            constraints = self._make_constraints(cosets)
+        else:
+            constraints = []
+
+        candidates = self._find_nodecolor_candidates()
+        la_candidates = self._get_lookahead_candidates()
+        for sgn in self.subgraph:
+            extra_candidates = la_candidates[sgn]
+            if extra_candidates:
+                candidates[sgn] = candidates[sgn].union([frozenset(extra_candidates)])
```
jbarnoud: You could generate a set directly rather than creating a list: `{frozenset(extra_candidates)}`. Am I needlessly nitpicking? Yes. So what?

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+    def _find_neighbor_color_count(graph, node, node_color, edge_color):
+        """
+        For `node` in `graph`, count the number of edges of a specific color
+        it has to nodes of a specific color.
+        """
+        counts = Counter()
+        neighbors = graph[node]
+        for neighbor in neighbors:
+            n_color = node_color[neighbor]
+            if (node, neighbor) in edge_color:
```
jbarnoud: Out of curiosity, is it what needs to be changed to deal with directed graphs? Or at least, one of the things.

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+            return
+
+    def analyze_symmetry(self, graph, node_partitions, edge_colors):
+        """
+        Find a minimal set of permutations and corresponding co-sets that
+        describe the symmetry of :attr:`subgraph`.
+
+        Returns
+        -------
+        tuple[set[frozenset], dict]
```
jbarnoud: You can split the returns description, which would increase readability:

```
Returns
-------
permutations: set[frozenset]
    A set of frozenset of pairs of node keys which can be exchanged without
    changing :attr:`subgraph`.
cosets: dict
    The co-sets is a dictionary of node key:
    set of node keys. Every key, value describes which values can be
    interchanged while keeping all nodes less than the key the same.
```

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+            The found permutations and co-sets. Permutations is a set of
+            frozenset of pairs of node keys which can be exchanged without
+            changing :attr:`subgraph`. The co-sets is a dictionary of node key:
+            set of node keys. Every key, value describes which values can be
+            interchanged while keeping all nodes less than the key the same.
```
jbarnoud: > Every key, value describes which values can be interchanged while keeping all nodes less than the key the same.

The sentence is correct, but I got to read it a few times to understand. Changing it to: " Every key-value pairs describes which values can be interchanged while keeping the same all nodes less than the key." Still not ideal, though.

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+        key = hash((tuple(graph.nodes), tuple(graph.edges),
+                    tuple(map(tuple, node_partitions)), tuple(edge_colors.items())))
+        if key in self._symmetry_cache:
+            return self._symmetry_cache[key]
+        node_partitions = list(self._refine_node_partitions(graph,
+                                                            node_partitions,
+                                                            edge_colors))
+        assert len(node_partitions) == 1
+        node_partitions = node_partitions[0]
+        permutations, cosets = self._process_opp(graph,
```
jbarnoud: You should ditch the abbreviation here.

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+                                                 node_partitions,
+                                                 node_partitions,
+                                                 edge_colors)
+        self.__class__._symmetry_cache[key] = permutations, cosets
```
jbarnoud: Why is the cache at the class and not the instance level?

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+
+
+def partition_to_color(partitions):
+    """
+    Creates a dictionary with for every item in partition for every partition
+    in partitions the index of partition in partitions.
+
+    Parameters
+    ----------
+    partitions: collections.abc.Sequence[collections.abc.Iterable]
```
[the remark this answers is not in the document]
pckroon: I hope so, but I don't know how. It's also bothered me in the rest of the codebase, tbh.

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+    def _node_compatibility(self):
+        if self._node_compat_ is not None:
+            return self._node_compat_
+        self._node_compat_ = {}
+        for sgn_part_color, gn_part_color in itertools.product(range(len(self._sgn_partitions)),
+                                                               range(len(self._gn_partitions))):
+            sgn = next(iter(self._sgn_partitions[sgn_part_color]))
+            gn = next(iter(self._gn_partitions[gn_part_color]))
+            if self.node_equality(self.subgraph, sgn, self.graph, gn):
+                self._node_compat_[sgn_part_color] = gn_part_color
```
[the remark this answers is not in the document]
pckroon: Yes. Because I define the colors as being ints I can use them as indices in the partitions. These are not related to node keys.

vermouth/ismags.py
pckroon: Yes. The direction of an edge has to be added to its color, but the nx way of defining things only gives us "outgoing" edges, so it'll involve quite some bookkeeping and gymnastics.

vermouth/ismags.py
pckroon: That's great, good to know, thanks :D

vermouth/ismags.py
pckroon: which one? 'opp'?

jbarnoud: yes

vermouth/ismags.py
pckroon: Partly because of the API I inherit from NX. You need to make a new object for every graph/subgraph pair, which would wipe the cache after every run. It would be much better/saner/cleaner if we could make an ISMAGS object once, and ask it for the isomorphisms between two graphs.

jbarnoud: OK. I do not like it, but I kind of understand. The problem I see is that it can be quite a memory leak in the way it is implemented here. You are never going to clean your many permutations and co-sets because you will always have a reference to them in that cache.

pckroon: Good point, I hadn't considered that. It'll really depend on the graphs in question whether it'll be bad or not. At least I only store a hash of the graph instead of the graph itself as key.

vermouth/ismags.py
```diff
@@ -0,0 +1,1030 @@
...
+                partition.add(item)
+                break
+        else:  # No break
+            partitions.append(set((item,)))
+    return partitions
+
+
+def partition_to_color(partitions):
+    """
+    Creates a dictionary with for every item in partition for every partition
```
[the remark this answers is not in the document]
pckroon: where? The sentence is a bit crooked, but I can't find any typo :)

pckroon:
I'll rebase after approval

vermouth/ismags.py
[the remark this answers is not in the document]
jbarnoud: I thought I caught a repetition, but it is indeed just the sentence that is ugly :wink:

jbarnoud:
Good for rebase. I still need to fish my review, though...

pckroon:
I'll wait with rebasing until you're finished reviewing. I don't want to risk GH losing your review history ;)

vermouth/ismags.py
```diff
@@ -0,0 +1,1046 @@
...
+               for partition in node_partitions):
+            yield node_partitions
+            return
+
+        new_partitions = []
+        output = [new_partitions]
+        for partition in node_partitions:
+            if not are_all_equal(node_edge_colors[node] for node in partition):
+                refined = make_partitions(partition, equal_color)
+                if branch and len(refined) != 1 and\
```
jbarnoud: Please use parentheses instead of escaping the new line:

```python
if (branch and len(refined) != 1
        and len({len(r) for r in refined}) != len([len(r) for r in refined])):
```

PR #156, opened 2018-10-30
Make the logging tests Python 3.8 compatible

pckroon:
In Python 3.8 logging.Formatter suddenly starts to try and validate it's format string using regexes. Since our `FormatCounter` was in fact not a string this broke. This PR fixes that. 
Note that it *must* have a formatstring passed (instead of `''`), otherwise it'll get replaced by a default formatting string, which is just a plain string.


jbarnoud:
Out of curiosity, did you comment on the python mailing list about this? Python 3.8 is not out yet, the change may still be reversible.

pckroon:
https://bugs.python.org/issue34998

pckroon:
Either way, this is an improvement over the ad-hoc ducktyping.

Issue #160, opened 2018-11-13
Links for protein termini not working as intended


pckroon:
Links for protein termini are currently defined as follows:
```
[ link ]
[ molmeta ]
neutral_termini not(true)
[ features ]
neutral_termini
[ atoms ]
BB {"replace": {"atype": "Qd", "charge": 1}}
[ non-edges ]
BB -BB

[ link ]
[ molmeta ]
neutral_termini not(true)
[ features ]
neutral_termini
[ atoms ]
BB {"replace": {"atype": "Qa", "charge": -1}}
[ non-edges ]
BB +BB
```

This means that an amino acid will have it's charge changed even if it has a bond to something that happened to be called differently than a 'BB'. This become rather tedious if there's non-standard residues involved. This can be resolved by adding the following to both links:
```
[ pattern ]
BB {"resname": $protein_resnames}
```
A better solution still would be to the number of bonds somehow.

pckroon:
An even better solution would be to look at the modifications found, and take it from there.

jbarnoud:
I agree that `[ non-edge ]` is not ideal as it forces to make assumptions.

> A better solution still would be to the number of bonds somehow.

Rarely a good idea when you deal with coarse grained model. Assuming the number of bonds is at least as bad as assuming the name of the next bead.

We could count the edges on the residue graph. But it would be very *ad-hoc*.

> An even better solution would be to look at the modifications found, and take it from there.

It would work. Though, in the specific case of the termini, the treatment is likely to be user based and not in accordance to what is in the source structure.

PR #162, opened 2018-11-15
Update martini3 proteins

jbarnoud:
Martini 3 force fields have versions:
* martini30b32 is the open beta
* martini30dev is the development version

The "martini30" force field is removed to avoid ambiguity.


vermouth/data/force_fields/martini30dev/aminoacids.ff
```diff
@@ -0,0 +1,859 @@
...
+ 2   SC3    1     PRO     SC1    2      0    
+
+[bonds]
+#meta {"group": "Side chain bonds"}
+;  i     j   funct   length  force.c.
+  BB   SC1   1       0.333     7500
+
+[ moleculetype ]
+; molname       nrexcl
+HYP                1
```
pckroon: HYP?

vermouth/data/force_fields/martini30dev/aminoacids.ff
```diff
@@ -0,0 +1,859 @@
...
+scfix
+[ molmeta ]
+scfix true
+[ angles ]
+SC1 BB +BB 10 100 15 {"group": "SC-BB-BB and BB-BB-SC scFix", "comment": "SC-BB-BB"}
+
+[ link ]
+resname $protein_resnames
+[ features ]
+scfix
```
pckroon: Fileformat question: Can we use a single link to do all the SC-fix things? It all acts on `SC1 BB +BB +SC1`.

vermouth/data/force_fields/martini30dev/aminoacids.ff
```diff
@@ -0,0 +1,859 @@
...
+[ link ]
+resname $protein_resnames
+[ angles ]
+-BB  BB  SC1  2 100 25 {"group": "BBS angles regular martini", "version": 1}
+
+[ link ]
+resname $protein_resnames
+[ angles ]
+#meta {"group": "First SBB regular martini"}
+SC1 BB +BB 2 100 25 {"version": 1}
```
pckroon: Doesn't this clobber the scfix one?

vermouth/data/force_fields/martini30dev/aminoacids.ff
```diff
@@ -0,0 +1,859 @@
...
+[ non-edges ]
+BB -BB
```
pckroon: I don't like non-edge specifiers, they make assumptions about what can be connected. Maybe `-any` would work, but I'd rather look at modifications found. See #160

vermouth/data/force_fields/martini30dev/aminoacids.ff
jbarnoud: Hydroxy-proline. Very common in collagen.

vermouth/data/force_fields/martini30dev/aminoacids.ff
jbarnoud: It would fail at the termini. It would also miss some that involve ALA or GLY as they do not have a SC1 bead.

vermouth/data/force_fields/martini30dev/aminoacids.ff
jbarnoud: No because of the "version" attribute. All interactions are version 0 by default. The clobbering only works for a given version.

vermouth/data/force_fields/martini30dev/aminoacids.ff
jbarnoud: I do not like it either. But I did not find a simple solution around it, yet. The change is out of scope of this PR, but I am happy to discuss the matter on #160.

PR #164, opened 2018-11-16
Fix-136, removing nodes from links.

NanduTej:
Hey, @pckroon  and @jbarnoud. From what I understood, we had to remove the node_mol whenever we had the atomname is None. So, this is the code change I made. I could not think of tests again. I need some help there.


vermouth/processors/do_links.py
```diff
@@ -282,6 +282,8 @@ def run_molecule(self, molecule):
             for match in matches:
                 for node, node_attrs in link.nodes.items():
                     if 'replace' in node_attrs:
+                        if node_attrs['replace']['atomname'] is None:
+                            molecule.remove_node(node_mol)
```
pckroon: You need to remove `match[node]`, which is the index/key of the node in molecule.

vermouth/processors/do_links.py
```diff
@@ -282,6 +282,8 @@ def run_molecule(self, molecule):
...
                         node_mol = molecule.nodes[match[node]]
                         node_mol.update(node_attrs['replace'])
```
pckroon: Only update `node_mol` if it hasn't been removed.

NanduTej:
Okay, that's great. :)

pckroon:
Another interesting testcase: multiple links that should be applied to the same atoms. The first removes an atom that is involved in a later link.

NanduTej:
Can you help me with an example of how to write a link? I tried it myself, but could not get it correctly.

pckroon:
Indeed it's not as trivial as I assumed/thought. This runs and should get you started:
```python
from vermouth.processors import do_links, DoLinks
from vermouth.molecule import Molecule, Link
import vermouth.forcefield

...

def make_mol(mol_nodes, mol_edges=[], **kwargs):
    mol = Molecule(**kwargs)
    mol.add_nodes_from(mol_nodes)
    mol.add_edges_from(mol_edges)
    return mol


def make_link(mol_nodes, mol_edges=[]):
    mol = Link()
    mol.add_nodes_from(mol_nodes)
    mol.add_edges_from(mol_edges)
    return mol

@pytest.mark.parametrize('mol_nodes, mol_edges, link_nodes, link_edges, expected_nodes, expected_edges', (
    (
         [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})],
         [(0, 1), (1, 2)],
         [[(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c', 'replace': {'atomname': None}})]],
         [[(0, 1), (1, 2)]],
         [(0, {'atomname': 'a'}), (1, {'atomname': 'b'})],
         [(0, 1)],
    ),
))
def test_link_processor(mol_nodes, mol_edges, link_nodes, link_edges,
                        expected_nodes, expected_edges):
    links = [make_link(nodes, edges) for nodes, edges in zip(link_nodes, link_edges)]
    ff = vermouth.forcefield.FORCE_FIELDS['universal']
    ff.links = links

    mol = make_mol(mol_nodes, mol_edges, force_field=ff)

    out = DoLinks().run_molecule(mol)
    assert dict(out.nodes(data=True)) == dict(expected_nodes)
    assert set(out.edges(data=False)) == set(expected_edges)
```

NanduTej:
def candidate_pairs_iter(self):
        """Iterator over candidate pairs of nodes in G1 and G2."""
    
        # All computations are done using the current state!
    
        G1_nodes = self.G1_nodes
        G2_nodes = self.G2_nodes
        min_key = self.G2_node_order.__getitem__
    
        # First we compute the inout-terminal sets.
        T1_inout = [node for node in self.inout_1 if node not in self.core_1]
        T2_inout = [node for node in self.inout_2 if node not in self.core_2]
    
        # If T1_inout and T2_inout are both nonempty.
        # P(s) = T1_inout x {min T2_inout}
        if T1_inout and T2_inout:
            node_2 = min(T2_inout, key=min_key)
            for node_1 in T1_inout:
                yield node_1, node_2
    
        else:
            # If T1_inout and T2_inout were both empty....
            # P(s) = (N_1 - M_1) x {min (N_2 - M_2)}
            # if not (T1_inout or T2_inout):       # as suggested by  [2], incorrect
            if 1:                                  # as inferred from [1], correct
                # First we determine the candidate node for G2
                other_node = min(G2_nodes - set(self.core_2), key=min_key)
>               for node in self.G1:
E               RuntimeError: OrderedDict mutated during iteration

/usr/local/lib/python3.5/dist-packages/networkx/algorithms/isomorphism/isomorphvf2.py:229: RuntimeError

This error pops up when I try to run the the code. We already got the match dictionary here - https://github.com/marrink-lab/vermouth-martinize/blob/c2bae2108675d8e94d14113b16ae12b8d282dbaf/vermouth/processors/do_links.py#L269

I am not able to understand how it is going into the isomorphism file again and giving the error. Also can you help me out to get an efficient way to debug.

NanduTej:
> I guess `match_link` is a [generator](https://jeffknupp.com/blog/2013/04/07/improve-your-python-yield-and-generators-explained/) and it iterate over `molecule`. When you iterate over `matches` you are still iterating over `molecule` hence the problem when you modify the molecule.
> 
> One solution would be to mark the nodes to remove, and to actually remove then after we applied all the links. This would have a side effect to document, though: if a later link refer to the node that was deleted, then it will be applied even though the node is not supposed to be there. It is not ideal as it is not consistent with the behaviour when we modify a node. I guess we could live with this... If not, then we will need to include the deletion marker in the semantic feasibility test.

NanduTej:
I accidentally closed the issue. I am sorry.

NanduTej:
> Another interesting testcase: multiple links that should be applied to the same atoms. The first removes an atom that is involved in a later link.

So, should we still be adding this case in the tests?

pckroon:
`match_link` is indeed a generator. I'm not sure how bad/desirable it is to just wrap the call in `list` (`matches = list(match_link(molecule, link))`).
Making a collection of atoms to be removed would work, and it will only cause issues if it's the same link acting on a removed atom in a different `match` --- although even there you could check whether any of the atoms in match are in the removed atoms collection.

And yes, that testcase should *definitely* be there. Even better, there's 1) the same link acting multiple times on the same atoms of which some get removed, and 2) different links acting in turn, of which the first removes atoms needed/matched by the second.

NanduTej:
> `match_link` is indeed a generator. I'm not sure how bad/desirable it is to just wrap the call in `list` (`matches = list(match_link(molecule, link))`).
> Making a collection of atoms to be removed would work, and it will only cause issues if it's the same link acting on a removed atom in a different `match` --- although even there you could check whether any of the atoms in match are in the removed atoms collection.
> 
> And yes, that testcase should _definitely_ be there. Even better, there's 1) the same link acting multiple times on the same atoms of which some get removed, and 2) different links acting in turn, of which the first removes atoms needed/matched by the second.

So you mean we don't want to wrap it in a list? I just tested with the base case you gave me and it worked without collecting the nodes to be removed.

pckroon:
Wrapping it in `list` will work indeed :)
But the problem there is that you negate all the advantages of having `match_link` being a generator --- the biggest being that all matches are generated immediately and stored in memory. If there are many (which there might) this can cause a tremendous slowdown. What you can do is leave it with `list` for now, make the tests, and then go back and implement the other solution.

jbarnoud:
Indeed, wrapping `match_link` would be functional for the simple case. But I expect it will break in our face rather quickly.

The simpler solution would be to have a `_nodes_to_remove` set, to store there the keys of the nodes to remove, and finally to use `Molecule.remove_nodes_from` after we found all the matches. This option, though will lead to link being applied to nodes that will be deleted. To remedy that, one could check that no node in the match is in `_node_to_remove`.

The cleaner solution, I think, would be o store a `_to_remove=True` attribute in the node to remove, and to modify `vermouth.processors.do_link._atom_match` to return `False` if any of the node has `_to_remove == True`. Then, remove all the nodes that are marked to be deleted.

Test-wise, there should be a test with a link that deletes a node, followed by a link that references that node. In such case, the second link should not be applied. There should also be the opposite test where there is a node that refers to a node that will be deleted in a latter link. In such case, the first link should apply as the node has not yet been marked to be removed.

pckroon:
@jbarnoud I think you're missing an edgecase, and overcomplicating the problem because of it. Consider the following cases with output options:

1) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'})]`; mol edges: `[(0, 1)]`.
Link nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link edges: `[(0, 1)]`.
  a) `[(0, {'atomname': 'a1'})]`
  b) `[(0, {'atomname': 'a'})]`
2) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'a'})]`; mol edges: `[(0, 1), (1, 2)]`.
Link nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link edges: `[(0, 1)]`.
  a) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'a1'})]`
  b) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'a'})]`
3) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})]`; mol edges: `[(0, 1), (1, 2)]`.
Link1 nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link1 edges: `[(0, 1)]`.
Link2 nodes: `[(0, {'atomname': 'b'}), (1, {'atomname': 'c', 'replace': {'atomname': 'c1'}})]`; link1 edges: `[(0, 1)]`.
  a) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})]`
  b) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c1'})]`

In all three cases I think answer a) is the correct one. This can be implemented fairly easily by calling `remove_nodes_from` right after the loop `for match in matches` completes.

jbarnoud:
I am reading this from my phone so I may miss things. But isn't the link that removes a node always the last one in your examples? Le 26 nov. 2018 10:55, Peter C Kroon <notifications@github.com> a écrit :@jbarnoud I think you're missing an edgecase, and overcomplicating the problem because of it. Consider the following cases with output options:
Mol nodes: [(0, {'atomname': 'a'}), (1, {'atomname': 'b'})]; mol edges: [(0, 1)].
Link nodes: [(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]; link edges: [(0, 1)].
a) [(0, {'atomname': 'a1'})]
b) [(0, {'atomname': 'a'})]Mol nodes: [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'a'})]; mol edges: [(0, 1), (1, 2)].
Link nodes: [(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]; link edges: [(0, 1)].
a) [(0, {'atomname': 'a1'}), (2, {'atomname': 'a1'})]
b) [(0, {'atomname': 'a1'}), (2, {'atomname': 'a'})]Mol nodes: [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})]; mol edges: [(0, 1), (1, 2)].
Link1 nodes: [(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]; link1 edges: [(0, 1)].
Link2 nodes: [(0, {'atomname': 'b'}), (1, {'atomname': 'c', 'replace': {'atomname': 'c1'}})]; link1 edges: [(0, 1)].
a) [(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})]
b) [(0, {'atomname': 'a1'}), (2, {'atomname': 'c1'})]
In all three cases I think answer a) is the correct one. This can be implemented fairly easily by calling remove_nodes_from right after the loop for match in matches completes.

jbarnoud:
I miss read indeed. The case I see as a problem is case 3 with the following difference: in link 2, replace `[(0, {'atomname': 'b'}), (1, {'atomname': 'c', 'replace': {'atomname': 'c1'}})]` by `[(0, {'atomname': 'a1'}), (1, {'atomname': 'b', 'replace': {'atomname': 'b1'}})]`. In this case, the link would be applied as there is a `a1` node because of the previous link, and as `b` is not removed yet.

pckroon:
I'm going to repost the three cases so we don't have to scroll up all the time. Let's make your example a 4th case:

1) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'})]`; mol edges: `[(0, 1)]`.
Link nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link edges: `[(0, 1)]`.
a) `[(0, {'atomname': 'a1'})]`
b) `[(0, {'atomname': 'a'})]`

2) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'a'})]`; mol edges: `[(0, 1), (1, 2)]`.
Link nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link edges: `[(0, 1)]`.
a) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'a1'})]`
b) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'a'})]`

3) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})]`; mol edges: `[(0, 1), (1, 2)]`.
Link1 nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link1 edges: `[(0, 1)]`.
Link2 nodes: `[(0, {'atomname': 'b'}), (1, {'atomname': 'c', 'replace': {'atomname': 'c1'}})]`; link2 edges: `[(0, 1)]`.
a) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})]`
b) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c1'})]`

4) Mol nodes: `[(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})]`; mol edges: `[(0, 1), (1, 2)]`.
Link1 nodes: `[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})]`; link1 edges: `[(0, 1)]`.
Link2 nodes: `[(0, {'atomname': 'a1', 'replace': {'atomname': 'a2'}}), (1, {'atomname': 'b1'})]`; link2 edges: `[(0, 1)]`.
a) `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})]`
b) `[(0, {'atomname': 'a2'}), (2, {'atomname': 'c'})]`

In cases 3 and 4 I don't think the second link should be applied since node 'b' is removed by a preceding link.
Case 4:
Apply link 1: `matches = [{0: 0, 1: 1}]`; molecule node 1 (name 'b') is removed.
Molecule now has the following nodes: `[(0, {'atomname': 'a'}), (2, {'atomname': 'c'})]`
Apply link 2: `matches = []`.

Also, saying that link 2 *should* be applied contradicts your earlier comments in this thread.

pckroon:
And if all links should be applied irrespective of whether atoms they apply to are removed earlier in the process: call `_remove_nodes_from` just before `return molecule`.

jbarnoud:
I had two loops switched in my mind when thinking about the code. Indeed, calling `remove_nodes_from` at the end of the match loop is enough, and there is no need to modify the semantic feasibility.

NanduTej:
Hey, just so I understand what we are supposed to do, we make matches into a list and iterates over all the nodes and store a list of nodes for which we would be applying remove_nodes_from after the match loop. And in the case of replace attribute being None, we would not updating the node_mols. Am I right?

pckroon:
Almost. Don't turn matches into a list and you're there. Plus, you got 4 interesting test cases ;)

pckroon:
@NanduTej How's progress? Is there anything else you need from our side?

NanduTej:
> @NanduTej How's progress? Is there anything else you need from our side?

Hey! I had finished the code earlier but didn't get around to pushing it. I was having exams in my college. Please review and let me know if it looks good.

pckroon:
Ok great! I'll have a look tomorrow.
PS. I hope the exams went well :)

vermouth/processors/do_links.py
```diff
@@ -277,13 +277,17 @@ def _build_link_interaction_from(molecule, interaction, match):
...
         links = molecule.force_field.links
+        _nodes_to_remove = []
         for link in links:
             matches = match_link(molecule, link)
             for match in matches:
                 for node, node_attrs in link.nodes.items():
                     if 'replace' in node_attrs:
-                        node_mol = molecule.nodes[match[node]]
-                        node_mol.update(node_attrs['replace'])
+                        if not node_attrs['replace']['atomname']:
```
pckroon: This is not the same as `if node_attrs['replace']['atomname'] is None:`

vermouth/processors/do_links.py
```diff
@@ -295,4 +299,6 @@ def run_molecule(self, molecule):
                     for interaction in interactions:
                         interaction = _build_link_interaction_from(molecule, interaction, match)
                         molecule.add_or_replace_interaction(inter_type, *interaction)
+
+        molecule.remove_nodes_from(_nodes_to_remove)
```
pckroon: Should be indented 4 more spaces

vermouth/tests/test_links.py
```diff
@@ -199,3 +201,72 @@ def test_interpret_order(order, ref_order_type, ref_order_value):
...
+        [(0, {'atomname': 'a1'}), (2, {'atomname': 'a1'})],
+        [],
+    ),
+    (
+        [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})],
+        [(0, 1), (1, 2)],
+        [[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})],
+         [(0, {'atomname': 'a1', 'replace': {'atomname': 'a2'}}), (1, {'atomname': 'b'})]],
+        [[(0, 1)], [(0, 1)]],
+        [(0, {'atomname': 'a2'}), (2, {'atomname': 'c'})],
```
pckroon: Expected output is: `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})]`

vermouth/tests/test_links.py
```diff
@@ -199,3 +201,72 @@ def test_interpret_order(order, ref_order_type, ref_order_value):
...
+        [],
+    ),
+    (
+        [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})],
+        [(0, 1), (1, 2)],
+        [[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})],
+         [(0, {'atomname': 'b'}), (1, {'atomname': 'c', 'replace': {'atomname': 'c1'}})]],
+        [[(0, 1)], [(0, 1)]],
+        [(0, {'atomname': 'a1'}), (2, {'atomname': 'c1'})],
```
pckroon: Expected output is: `[(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})]`

NanduTej:
I made the changes requested. The patch coverage shown here is 0% although we wrote tests for this. Is this something to look into?

vermouth/tests/test_links.py
```diff
@@ -199,3 +201,72 @@ def test_interpret_order(order, ref_order_type, ref_order_value):
...
+        [],
+    ),
+    (
+        [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})],
+        [(0, 1), (1, 2)],
+        [[(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': None}})],
+         [(0, {'atomname': 'b'}), (1, {'atomname': 'c', 'replace': {'atomname': 'c1'}})]],
+        [[(0, 1)], [(0, 1)]],
+        [(0, {'atomname': 'a1'}), (2, {'atomname': 'c'})],
+        [],
```
pckroon: This testcase is functionally identical to the previous one.
Could you change it with this?
```
(
    [(0, {'atomname': 'a'}), (1, {'atomname': 'b'}), (2, {'atomname': 'c'})],
    [(0, 1), (1, 2)],
    [
         [(0, {'atomname': 'a', 'replace': {'atomname': 'a1'}}), (1, {'atomname': 'b', 'replace': {'atomname': 'b1'}})],
         [(0, {'atomname': 'c', 'replace': {'atomname': 'c1'}}), (1, {'atomname': 'b1', 'replace': {'atomname': 'b2'}})]
    ],
    [[(0, 1)]],
    [(0, {'atomname': 'a1'}), (1, {'atomname': 'b2'}), (2, {'atomname': 'c1'})],
    [(0, 1), (1, 2)],
),
```

NanduTej: Okay.

pckroon:
By the way, could you please only force push when you really have to?

NanduTej:
> By the way, could you please only force push when you really have to?

Okay.

vermouth/tests/test_links.py
NanduTej: I think you missed a link edge here. Was it intentional?

pckroon: No, good catch.

vermouth/processors/do_links.py
```diff
@@ -277,13 +277,17 @@ def _build_link_interaction_from(molecule, interaction, match):
...
         links = molecule.force_field.links
+        _nodes_to_remove = []
         for link in links:
             matches = match_link(molecule, link)
             for match in matches:
                 for node, node_attrs in link.nodes.items():
                     if 'replace' in node_attrs:
-                        node_mol = molecule.nodes[match[node]]
-                        node_mol.update(node_attrs['replace'])
+                        if node_attrs['replace']['atomname'] is None:
```
pckroon: This will break if `'atomname'` is not in `node_attrs['replace']`. Replace with `if node_attrs['replace'].get('atomname', False):`

NanduTej:
Thank you :) !

PR #173, opened 2019-01-21
Remove unused deferred_file.py

jbarnoud:
The module is not used and not tested. While the feature is still
desirable, it is not a priority just now. Anyway, the implementation is
not settled and removing the unused module is a cheap way of increasing
the overall coverage.


pckroon:
So long as someone keeps a copy of that code, fine by me. We will need it later on for sure though.

jbarnoud:
Ready for prime time.

PR #174, opened 2019-01-22
Remove deadline for test_maximum_common_subgraph

jbarnoud:
The test test_hypothesis_graph_utils/test_maximum_common_subgraph fails
on a regular basis because it is too slow, or because its timing is
inconsistent.

Hopefully, this is a temporary fix and ISMAGS will make the thing
faster.


jbarnoud:
Hehe, I cannot approve my own PR.

PR #188, opened 2019-02-18
Less scary disclaimer in the README

jbarnoud:
It is much safer to use martinize2 and vermouth than it use to be. We might be less scary in our disclaimer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,7 @@ transformation on molecular structures and topologies using graph algorithms.
 
 ## Disclaimer
 
-**Both Martinize2 and vermouth are at a prototype stage. DO NOT USE IN
-PRODUCTION.**
+Martinize2 and Vermouth are under developpement. Bugs can occur. If you encounter an issue, please report it on <https://github.com/marrink-lab/vermouth-martinize/issues>.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ transformation on molecular structures and topologies using graph algorithms.
 
 ## Disclaimer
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,8 +18,7 @@ transformation on molecular structures and topologies using graph algorithms.
 
 ## Disclaimer
 
-**Both Martinize2 and vermouth are at a prototype stage. DO NOT USE IN
-PRODUCTION.**
+Martinize2 and Vermouth are under developpement. Bugs can occur. If you encounter an issue, please report it on <https://github.com/marrink-lab/vermouth-martinize/issues>.
```
pckroon: ... Bugs can occur, so check your topologies and read the output from martinize2 and VerMoUTH carefully. ...

Might be better?

jbarnoud: My idea was that we will have a small audience in the next few weeks/months. They should be encouraged to report issues, even if they come from their ill-formed files. There are two reasons for that:

1. I'd rather have to many issues reported than missing an important one, and
2. Issues dues to user inputs being broken will tell us what are the most common errors so we can provide heuristics, raise better error messages, or document the most common errors.

pckroon: I agree with both your points. However, I'm a little worried of people using our topologies without checking them. If they contain bugs it'll be a serious waste of everyone's time, and *very* bad PR for us.

jbarnoud:
Is this new that github becomes annoying with "outdated" branches?

pckroon:
Yes, but feel free to hit merge anyway. It did expose mystery issues with #172

PR #194, opened 2019-02-22
Processor to add includes for virtual site-based GoMartini

jbarnoud:
Sebastian's script for VS-based GoMartini generates ITP files to include at the ens of the `[ atoms ]`, `[ virtual_sitesn ]`, and `[ exclusions ]` sections. The processor in this PR allows to add these includes in the ITP files generated by martinize2. It is activated in martinize2 with the `-govs-includes` flag that is mutually exclusive with the `-elastic` one.

To allow this processor, the PR also add a mechanism to tell `write_molecule_itp` to write lines at the beginning or at the end of a section. These lines can be passed as function arguments, or as a meta attribute of the molecules.

The full VS-based GoMartini generation will ultimately become a processor. Though, the processor in the PR makes the method available in the meantime.

The feature in the PR is ready, but it requires some discussion with Paulo and Sebastian to deal with practical details and compatibility with Sebastian's script. One feature that remains to be added is to place the virtual sites in the output structure.


vermouth/tests/gmx/test_itp.py
```diff
@@ -74,3 +78,127 @@ def test_header(tmpdir, dummy_molecule):
...
+            'charge_group': 1, 'charge': 1,
+        }),
+    ))
+    # The molecule has 6 interaction sections. The sections are names
+    # "interaction_<joint>_<part>", where <joint> can be "with" or "only", and
+    # <part> can be "pre", "post", or "pre_post". The <part> indicates what
+    # pre_section_lines and post_section_lines are defined for the section.
+    # When <joint> in "with", then the section contains 2 interactions in
+    # addition to the pre- and post- lines. When <joint> is "only", then the
+    # section only has the lines defined, and no interactions.
```
pckroon: the lines? I guess pre/post lines?
And could/should this be pytest parametrized?

vermouth/tests/gmx/test_itp.py
```diff
@@ -74,3 +78,127 @@ def test_header(tmpdir, dummy_molecule):
...
+        """
+        [ interaction_only_pre_post ]
+        test_prepost_only_0
+        test_prepost_only_1
+        after_prepost_only_0
+        """,
+    ]
+
+    if post_meta:
+        arg_post = post_section_lines
```
pckroon: Set the meta to something wrong

vermouth/tests/gmx/test_itp.py
```diff
@@ -74,3 +78,127 @@ def test_header(tmpdir, dummy_molecule):
...
+    else:
+        arg_post = None
+        molecule.meta['post_section_lines'] = post_section_lines
+    if pre_meta:
+        arg_pre = pre_section_lines
```
pckroon: Same here, set the meta

vermouth/tests/gmx/test_itp.py
jbarnoud: Fixed the phrasing. The test could be parametrized, it was my initial idea. Though, a brief trial with that way lead to a less clear test.

pckroon: I find the test fairly mind-boggling as is, but ok.

vermouth/processors/go_vs_includes.py
```diff
@@ -70,3 +80,51 @@ def run_molecule(self, molecule):
...
+    new_node_id = max(molecule.nodes)
+    new_charge_group = max(node['charge_group'] for node in molecule.nodes.values())
+    for node_id, atom in molecule.nodes(data=True):
+        if atom.get('atomname') == atomname:
+            new_node_id += 1
+            new_charge_group += 1
+            virtual_site_nodes.append((new_node_id, {
+                'resid': atom['resid'],
+                'resname': atom['resname'],
+                'atomname': 'CA',
```
pckroon: Why is the atomname hardcoded?

vermouth/processors/go_vs_includes.py
```diff
@@ -70,3 +80,51 @@ def run_molecule(self, molecule):
...
+                'atype': '{}_{}'.format(prefix, atom['resid']),
+                'charge_group': new_charge_group,
+                'chain': atom['chain'],
+                'position': atom['position'],
+                'charge': 0,
```
pckroon: And the charge

vermouth/processors/sort_molecule_atoms.py
```diff
@@ -0,0 +1,50 @@
...
+        node_order = sorted(molecule, key=partial(_keyfunc, molecule))
+        sorted_nodes = [
+            (node_key, molecule.nodes[node_key])
+            for node_key in node_order
+        ]
+        # remove_nodes_from would be better, but it leads to an OrderedDict
+        # being modified during iteration, and python does not like that.
+        for node in node_order:
+            super(type(molecule), molecule).remove_node(node)
+        molecule.add_nodes_from(sorted_nodes)
```
pckroon: `nx.relabel_nodes(molecule, dict(sorted_nodes))` (or the inverse of sorted_nodes. You'll need to check it makes a Molecule again though. Also, what about interactions and such? In addition, what you're doing here loses all edges.

vermouth/processors/go_vs_includes.py
```diff
@@ -70,3 +80,51 @@ def run_molecule(self, molecule):
...
+    the molecule type name.
+    """
+    # If there are no atoms, then there is nothing to do. We can exit early,
+    # avoiding to deal with empty iterators.
+    if not molecule.nodes:
+        return
+    virtual_site_nodes = []
+    virtual_sites = []
+    new_node_id = max(molecule.nodes)
+    new_charge_group = max(node['charge_group'] for node in molecule.nodes.values())
```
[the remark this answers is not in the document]
jbarnoud: Good catch!

vermouth/processors/go_vs_includes.py
jbarnoud: This was lazyness. I'll fix it.

vermouth/processors/sort_molecule_atoms.py
jbarnoud: From what I understand, `relabel_nodes` changes the key/labels but not the order. I want to do the opposite: changing the order but not the labels. I did not notice the issue with the edges, though. If the labels are not changed, then I can back the edges up, are recreate them.

pckroon: My bad.
`molecule.nodes.move_to_end`? Looks to be a lot easier.
https://docs.python.org/3/library/collections.html#collections.OrderedDict.move_to_end

Also, I glossed over line 36, but that's also something I'm not happy with at all.

vermouth/processors/merge_all_molecules.py
```diff
@@ -0,0 +1,34 @@
...
+    The molecules are merged into the first molecule of the system. Nothing is
+    done if there are no molecules.
+    """
+    def run_system(self, system):
+        if not system.molecules:
+            return system
+
+        molecule = system.molecules[0]
+        for other in system.molecules[1:]:
+            molecule.merge_molecule(other)
```
pckroon: Can't shouldn't this reuse code from MergeChains?

vermouth/processors/sort_molecule_atoms.py
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+Provides a processor that sort atoms within molecules.
```
pckroon: s/sort/sorts/

vermouth/processors/sort_molecule_atoms.py
pckroon: I'm still unhappy with line 41 (was 36).
It can not be a good idea to have to use `super` outside of an inheriting class.

vermouth/processors/merge_all_molecules.py
jbarnoud: It could, but the added logic of listing the chains and generating the inputs for MergeChain, and the added API is almost as much of adding the full new processor. I went for the simpler user interface, but I could leverage MergeChain.

vermouth/processors/sort_molecule_atoms.py
pckroon: `molecule.nodes.move_to_end`, if possible.
Otherwise, just call remove_node (or better, remove_nodes_from, can't see what that doesn't work) and re-add the interactions.

oliverdutton:
I'm working on adding Go potentials for protein folding. Can I ask who is Sebastian, and what have you added here? How would I use this?

I appreciate this may just be a niche thing for a friend, but I'd love to know.

pckroon:
Hi Oliver,

"Sebastian" refers to a former colleague, Sebastian Thallmair. I think he was the first one to implement the combination of Go potentials with Martini3. See also 10.1021/acs.jpclett.9b02868 and 10.1021/acs.jctc.6b00986. He made a script that was capable of interpreting the contact maps generated by a website, and turn those into an itp with virtual sites. Martinize2 at the moment creates the required virtual sites, and adds appropriate `#include` statements at the right spot. 
I would love to have the Go potentials/method/vs fully integrated in martinize2, but I won't have the time anytime soon.

PR #198, opened 2019-03-19
Set up deployment with Travis

pckroon:
Once it works I also want to write a short how-to in the `maintainers` folder. Remind me if I forget.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,13 @@ env:
     - SKIP_WRITE_GIT_CHANGELOG=1
   matrix:
     - WITH_SCIPY=true WITH_CODECOV=true
-install:
+before_install:
   - pip install --upgrade setuptools pip
+install:
   - if [[ ${WITH_SCIPY} == true ]]; then pip install scipy; fi
-  - pip install --upgrade -r requirements-tests.txt
   - pip install --upgrade .
+before_script:
+  - pip install --upgrade -r requirements-tests.txt
 script:
   - coverage run --source=vermouth $(which pytest) -vv vermouth --hypothesis-show-statistics
 after_success:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -65,18 +67,20 @@ jobs:
...
-#         provider: pypi
-#         on:
-#             tags: true
-
+    - stage: deploy
+      python: "3.7"
+      addons: skip
+      env:
+      script: skip
+      after_success: skip
```
jbarnoud: Do you need the explicitly skip the steps?

.travis.yml
```diff
@@ -10,11 +10,13 @@ env:
...
   matrix:
     - WITH_SCIPY=true WITH_CODECOV=true
-install:
+before_install:
   - pip install --upgrade setuptools pip
+install:
   - if [[ ${WITH_SCIPY} == true ]]; then pip install scipy; fi
-  - pip install --upgrade -r requirements-tests.txt
   - pip install --upgrade .
+before_script:
```
jbarnoud: Why do you need these steps to be in `before_install` and `before_script`? Why not just in `install`?

.travis.yml
pckroon: Otherwise it'll run the tests again. I can try without this at a later point.

.travis.yml
pckroon: Need is a large word I think. I expect it to make deploy a little faster since it won't have to install all our testing equipment.

.travis.yml
jbarnoud: I did not think of that. Thanks.

.travis.yml
jbarnoud: Good idea.

pckroon:
I think this is as far as we can test it in a PR: `Skipping a deployment with the pypi provider because the current build is a pull request.`

maintainers/releasing.md
```diff
@@ -0,0 +1,23 @@
...
+- Implement your awesome feature, including docs.
+- Make a PR of your feature, get it reviewed and make sure Travis passes.
+- Get your feature merged in master
+- Create a tag: `git tag v0.5.0 -a`. Come up with a sane version number.
+- Push the tag: `git push --tags`.
+- Profit.
+
+Note that Travis will build releases *twice*. Once because something is pushed
+to master, and once because it's a tagged commit. But since we don't release
+too often this is OK. Also, if this is to change Travis should first come up
```
jbarnoud: The documentation is not the place to spit your frustration :wink:

PR #204, opened 2019-07-11
Fix PTM branch traversal to prevent finding too many involved atoms

pckroon:
Fixes part of #142
Sem-Ver: bugfix

### DIFF
--- a/vermouth/processors/canonicalize_modifications.py
+++ b/vermouth/processors/canonicalize_modifications.py
@@ -78,35 +78,34 @@ def find_ptm_atoms(molecule):
     ptms = []
     while extra_atoms:
         # First PTM atom we'll look at
-        first = next(iter(extra_atoms))
+        orig = next(iter(extra_atoms))
         anchors = set()
         # PTM atoms we've found
         atoms = set()
         # Atoms we still need to see this traversal
-        to_see = set([first])
+        to_see = set()
         # Traverse in molecule.
-        for orig, succ in nx.bfs_successors(molecule, first):
-            # We've seen orig, so remove it
-            if orig in to_see:
-                to_see.remove(orig)
-            if orig in extra_atoms:
+        while True:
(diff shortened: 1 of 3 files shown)


vermouth/processors/canonicalize_modifications.py
```diff
@@ -78,35 +78,34 @@ def find_ptm_atoms(molecule):
...
-            if orig in extra_atoms:
+        while True:
+            succ = molecule[orig].keys()
+            if orig in extra_atoms and orig not in atoms:
                 # If this is a PTM atom, we want to see it's neighbours as
                 # well.
                 to_see.update(succ)
                 atoms.add(orig)
-            else:
+            elif orig not in extra_atoms and orig not in anchors:
```
Tsjerk: 'and orig not in anchors' makes some sense for following the logic, but omitting the check does not change the result of the code (adding an element to a set that already contains that element); I think I'd prefer removing it as it then follows that any 'orig' that failed the first test is an anchor and added to the set as such.

vermouth/processors/canonicalize_modifications.py
```diff
@@ -78,35 +78,34 @@ def find_ptm_atoms(molecule):
...
-        to_see = set([first])
+        to_see = set()
         # Traverse in molecule.
-        for orig, succ in nx.bfs_successors(molecule, first):
-            # We've seen orig, so remove it
-            if orig in to_see:
-                to_see.remove(orig)
-            if orig in extra_atoms:
+        while True:
+            succ = molecule[orig].keys()
```
Tsjerk: You only use 'succ' within the first if-block. Why not put it there?

pckroon:
Travis currently fails on building docs, which is fixed in #206. So merge that first.

pckroon:
Or, I just cherry-pick the appropriate commit.

PR #211, opened 2019-10-03
Overhaul and upgrade PDB parser

pckroon:
This reworks the PDB parser to use the machinery in parser_utils. In addition, adds CLI support for `-ignh` (drop hydrogens from input) and `-model n` (allow selection of specific models). Also implements half support for altloc atoms. The full solution would involve adding a CLI argument for it, but you probably want this per residue, and I'm unsure 1) what the syntax should look like and 2) how to propagate/implement it. Instead I've made it so we always take model A and give a warning for every atom with a B conformation.

Closes #209 as it contains the same change.

Sem-Ver: feature


bin/martinize2
```diff
@@ -375,6 +382,14 @@ def entry():
                      'compatible. The -elastic and -govs-include flags cannot '
                      'be used together.')
 
+    file_extension = args.inpath.suffix.upper()[1:]  # We do not keep the dot
+    if file_extension in ['GRO'] and args.modelidx is not None:
+        parser.error("GRO files don't know the concept of models.")
```
Tsjerk: In principle they do: GRO files may contain multiple frames. They are not numbered, though, so it won't be entirely clear what model selection by number means.

Tsjerk: They do: the notion of frame and model are interchangeable.

vermouth/pdb/pdb.py
```diff
@@ -16,13 +16,408 @@
...
+    # CONNECTIVITY ANNOTATION SECTION
+    ssbond = _skip
+    link   = _skip
+    cispep = _skip
+
+    # MISCELLANEOUS FEATURES SECTION
+    site   = _skip
+
+    # CRYSTALLOGRAPHIC AND COORDINATE TRANSFORMATION SECTION
+    cryst1 = _skip
```
Tsjerk: This is the unit cell and should probably be retained at some point to allow coarse graining a full simulation system. A comment will do here for now and an issue for reference.

vermouth/pdb/pdb.py
```diff
@@ -16,13 +16,408 @@
...
+
+        Parameters
+        ----------
+        line: str
+            The line to parse. Should start with CONECT, but this is not checked
+        lineno: int
+            The line number (not used).
+        """
+        # We can't add edges immediately, since the molecule might not be parsed
+        # yet (does the PDB file format mandate anything on the order of
```
Tsjerk: CONECT records are always at the end of the file. The PDB format specification gives an order of records: http://www.wwpdb.org/documentation/file-format-content/format33/sect1.html
If this may lead to simplification of the code, then it could be listed as an issue.

pckroon: It would simplify the code, but this makes us a bit more resilient to badly formatted pdb files.

Tsjerk: I concur.

PR #213, opened 2019-10-04
Work-around for pypa/pip#7135

pckroon:
This is causing the CI failures on #210  and #211


pckroon:
I'm going to go ahead and merge this. The current docs failure is fixed in #208.

PR #215, opened 2019-10-09
Some long overdue linting

pckroon:
This PR removes some old lint. Mostly whitespaces and imports that are either unnecessary or in the wrong place.

### DIFF
--- a/vermouth/edge_tuning.py
+++ b/vermouth/edge_tuning.py
@@ -324,17 +324,16 @@ def pairs_under_threshold(molecules, threshold,
 def _are_close(threshold, molecules, key_a, key_b):
     if not threshold:
         return False
-    if key_a[0] == key_b[0]:
-        try:
-            length = nx.shortest_path_length(
-                molecules[key_a[0]], key_a[1], key_b[1]
-            )
-        except nx.exception.NetworkXNoPath:
-            return False
-        else:
-            return length < threshold
-    else:
+    if key_a[0] != key_b[0]:
         return False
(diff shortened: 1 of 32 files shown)


vermouth/edge_tuning.py
```diff
@@ -336,6 +333,8 @@ def _are_close(threshold, molecules, key_a, key_b):
             return False
         else:
             return length < threshold
+    else:
```
Tsjerk: Can this be turned around: if a != b: return False and if we're still running do the try-except? Removes a level of indentation.

vermouth/processors/repair_graph.py
```diff
@@ -130,10 +130,14 @@ def make_reference(mol):
...
-                             key=lambda jdx: (res_names[jdx] not in ref_names.values(), res_names[jdx])))}
-        new_reference_names = {old: new for new, old in enumerate(sorted(reference,
-                               key=lambda jdx: (ref_names[jdx] not in res_names.values(), ref_names[jdx])))}
+        new_residue_names = {old: new for new, old in enumerate(sorted(
+            residue,
+            key=lambda jdx: (res_names[jdx] not in ref_names.values(), res_names[jdx])
+        ))}
+        new_reference_names = {old: new for new, old in enumerate(sorted(
+            reference,
+           key=lambda jdx: (ref_names[jdx] not in res_names.values(), ref_names[jdx])
```
Tsjerk: misalignment: one more space before 'key'

PR #216, opened 2019-10-23
Replace all Graph.node with Graph.nodes

pckroon:
since the former was removed in Networkx 2.4.

I'll mint a bugfix release after this PR is merged.


pckroon:
Of course ;)
`grep -nR "\.node[^s_]"`

PR #220, opened 2020-02-03
Fix consistent test failures with strategic use of assume

pckroon:
Since Hypothesis 5.0 st.sampled_from([]) results in an error. Avoid those error by no longer sampling from empty graphs.

### DIFF
--- a/vermouth/tests/molecule_strategies.py
+++ b/vermouth/tests/molecule_strategies.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import hypothesis.strategies as st
-from hypothesis import assume
 from hypothesis_networkx import strategy as hnst
 import vermouth
 import vermouth.molecule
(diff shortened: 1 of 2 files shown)


jbarnoud:
This means we do not test of empty graphs any more. Isn't it an issue? What about setting nodes to an empty list is the graph is empty?

pckroon:
The (branch) coverage doesn't drop, so I don't think it's an issue.
In addition, looking at e.g. test_molecule, it only means that the branch where there is no Graph, /and/ we want to draw >0 interactions is pruned. 
> What about setting nodes to an empty list is the graph is empty?

What do you mean by this?

jbarnoud:
> What do you mean by this?

Instead of
```python
assume(graph)
nodes = data.draw(...)
```

you could have:
```python
if graph:
    nodes = data.draw(...)
else:
    nodes = []
```

PR #224, opened 2020-02-04
Provide list data CLI options for force fields, blocks and modifications

pckroon:
Add `-list-ff` and `-list-block` which list all the known FFs, and all known Blocks and modifications for the selected force fields, respectively.


jbarnoud:
For the same of grepping, I'll have the blocks and modifications as bullet lists (one per line) rather than a single long comma separated line.

pckroon:
That does mean you'll have your screen filled very quickly, and may make it less clear where modifications start, and where it lists blocks/mods for which FF.

Issue #57, opened 2018-03-14
"order" attribute in links is fragile


jbarnoud:
Link depending on the position of two residues in a sequence rely exclusively in the resid to figure out the position in the sequence. This will surely break with branched and cyclic molecules.

One solution would be to consider the "order" attribute a distance in the residue graph, and only use the resid to decide the direction of the sequence. It would solve the issue for branched molecules, but the sequence orientation will break a the junction in cyclic molecules. Also, this would require to cache the residue graph as it would be consulted quite a lot in the linking process.

pckroon:
The directionality we might be able to capture in a directed graph (nx.DiGraph). But fixing cyclic structures is still a challenge then. We could argue something about the maximum in/out degree of residues, and take it from there. But that's also fragile.

pckroon:
Together with #170 and #161 this is solved by using ISMAGS for the link isomorphisms. Meaning every link will only get applied asymmetrically. If a link is supposed to be asymmetric, it should cover more atoms until it is unambiguous.

jbarnoud:
My brain has shown to be in a less than optimal state today, though I do not see how ISMAGS is relevant here. The issue is that we use the resid to deal with the "order" keyword. This means that an edge such as "BB +BB" has to rely on the residue sequence being continuous; it will break for instance with PTM being their own residue.

pckroon:
That was the result of a discussion we had some time ago :) I'll try to summarize it here.

The "order" attribute exists solely to break symmetry and not apply the same link to the same atoms multiple times. ISMAGS also breaks symmetry and can thus be used to serve the same purpose. 
It creates the following new edge cases:
1) I want to apply the same link to the same atoms multiple times. A: Don't. How often do you want it? Instead, add multiple interactions to the same link.
2) My interaction is asymmetric and needs to be applied the right way round. A: You link is underdefined if the atoms the link specifies are symmetric, but the interactions are not. Add more atoms to the link until the graph is no longer symmetric

jbarnoud:
OK I remember. Getting rid of the order would change quite a lot the way of writing links, though; we need to assess if it is worth the trouble.

pckroon:
I think it would enable just removing the order attribute from links, and makes writing them much much easier.
For failsafes and helping the user it might be nice if we had a list of interactions we know are asymmetric, and warn if they can get applied the wrong way round.

jbarnoud:
[...]

PR #226, opened 2020-02-05
Add the notion of domains in the elastic network

### LINKED ISSUE #221: `-merge` does not play well with elastic networks
Using the `-merge` option with the elastic network or a go model can cause issues. There are two behaviours that can be desirable:

* the merge is done *before* the elastic network leading to bonds between the chains
* the merge is done *after* the elastic network leading to *no* bonds between the chains

Disulfide bonds add a layer to the issue: by bridging chains into a single a `Molecule` they always cause elastic bonds between the chains.

Possible solutions are:
* use chains to filter the elastic network
* split molecules around disulfide bridges when building the elastic networks

Test cases:
* simpler case: single chain (should keep working)
* A A merge before -> create bonds between the chains
* A A merge after -> do not create bonds between the chains
* A B merge before -> create bonds between the chains
* A B merge after -> do not create bonds
* A-s-s-A merge before -> create bonds
* A-s-s-A merge after -> do not create bonds (carreful with the bridges, test with more than 1 bridge)

How does it behave with `-sep`?

### LINKED ISSUE #267: -elastic option shows less bonds than martinize1
My current tests with vermouth 0.3.2.dev122 and martini22 models showed that in relation to martinize1, the option -elastic in martinize2 produced as output less EN bonds.

jbarnoud:
Fixes #221 
Fixes #267 

* Add the notion of elastic network domains. Bonds are only added within a domain.
* Add the `-eunit` option to the CLI to specify what the domains are. The valid options are 'all', 'chain', and 'molecule'.

In the future, `-eunit` could accept an index file or a domElnedyn input file. Having that working would mostly mean writing a function that can be use as argument for the `domain_criterion` argument of the processor.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -43,6 +43,7 @@ from vermouth.map_input import (
     generate_all_self_mappings,
     combine_mappings
 )
+from vermouth.processors.apply_rubber_band import always_true, same_chain
 
 # TODO Since vermouth's __init__.py does some logging (KDTree), this may or may
 # not work as intended. Investigation required.
@@ -366,6 +367,11 @@ def entry():
     rb_group.add_argument('-eb', dest='rb_selection',
                           type=lambda x: x.split(','), default=None,
                           help='Comma separated list of bead names for elastic bonds')
+    rb_group.add_argument('-eunit', dest='rb_unit', default='molecule',
+                          choices=['chain', 'molecule', 'all'],
+                          help=('Establish what is the structural unit for the '
+                                'elastic network. Bonds are only created within'
+                                'a unit.'))
 
     go_group = parser.add_argument_group('Virtual site based GoMartini')
     go_group.add_argument('-govs-includes', action='store_true', default=False,
@@ -427,7 +433,7 @@ def entry():
                                    'CanonicalizeModifications step. The '
                                    'resulting file may contain "nan" '
                                    'coordinates making it unreadable by most '
-                                   'softwares.'))
+                                   'software.'))
     debug_group.add_argument('-v', dest='verbosity', action='count',
                              help='Enable debug logging output. Can be given '
                                   'multiple times.', default=0)
@@ -585,27 +591,9 @@ def entry():
         delete_unknown=True,
     )
 
-    # Apply a rubber band elastic network is required.
-    if args.elastic:
-        LOGGER.info('Setting the rubber bands.', type='step')
-        if args.rb_selection is not None:
-            selector = functools.partial(
-                selectors.proto_select_attribute_in,
-                attribute='atomname',
-                values=args.rb_selection,
-            )
-        else:
-            selector = selectors.select_backbone
-        rubber_band_processor = vermouth.ApplyRubberBand(
-            lower_bound=args.rb_lower_bound,
-            upper_bound=args.rb_upper_bound,
-            decay_factor=args.rb_decay_factor,
-            decay_power=args.rb_decay_power,
-            base_constant=args.rb_force_constant,
-            minimum_force=args.rb_minimum_force,
-            selector=selector,
-        )
-        rubber_band_processor.run_system(system)
+    if args.neutral_termini:
+        LOGGER.info('Making termini neutral.', type='step')
+        vermouth.NeutralTermini(force_field=known_force_fields[args.to_ff]).run_system(system)
 
     # Apply position restraints if required.
     if args.posres != 'none':
@@ -638,6 +626,40 @@ def entry():
                 vermouth.MergeChains(chain_set).run_system(system)
         vermouth.NameMolType(deduplicate=not args.keep_duplicate_itp).run_system(system)
         defines = ()
+
+    # Apply a rubber band elastic network is required.
+    if args.elastic:
+        LOGGER.info('Setting the rubber bands.', type='step')
+        if args.rb_unit == 'molecule':
+            domain_criterion = always_true
+        elif args.rb_unit == 'all':
+            vermouth.MergeAllMolecules().run_system(system)
+            domain_criterion = always_true
+        elif args.rb_unit == 'chain':
+            domain_criterion = same_chain
+        else:
+            message = 'Unknown value for -eunit: "{}".'.format(args.rb_unit)
+            LOGGER.critical(message)
+            raise ValueError(message)
+        if args.rb_selection is not None:
+            selector = functools.partial(
+                selectors.proto_select_attribute_in,
+                attribute='atomname',
+                values=args.rb_selection,
+            )
+        else:
+            selector = selectors.select_backbone
+        rubber_band_processor = vermouth.ApplyRubberBand(
+            lower_bound=args.rb_lower_bound,
+            upper_bound=args.rb_upper_bound,
+            decay_factor=args.rb_decay_factor,
+            decay_power=args.rb_decay_power,
+            base_constant=args.rb_force_constant,
+            minimum_force=args.rb_minimum_force,
+            selector=selector,
+            domain_criterion=domain_criterion,
+        )
+        rubber_band_processor.run_system(system)
 
     LOGGER.info('Writing output.', type='step')
     for molecule in system.molecules:
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -358,7 +358,9 @@ def entry():
     rb_group.add_argument('-eu', dest='rb_upper_bound', type=float, default=0.9,
                           help='Elastic bond upper cutoff: F = 0  if rij > up')
     rb_group.add_argument('-ermd', dest='res_min_dist', type=int,
-                          help='the minimum seperation between to residues to have an RB')
+                          help=('the minimum separation between two residues to have an RB'
+                                'the default value is set by the force-field.'),
+                          default=None)
     rb_group.add_argument('-ea', dest='rb_decay_factor', type=float, default=0,
                           help='Elastic bond decay factor a')
     rb_group.add_argument('-ep', dest='rb_decay_power', type=float, default=0,
@@ -372,7 +374,7 @@ def entry():
                           choices=['chain', 'molecule', 'all'],
                           help=('Establish what is the structural unit for the '
                                 'elastic network. Bonds are only created within'
-                                'a unit.'))
+                                ' a unit.'))
 
     go_group = parser.add_argument_group('Virtual site based GoMartini')
     go_group.add_argument('-govs-includes', action='store_true', default=False,
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -43,7 +43,6 @@ from vermouth.map_input import (
     generate_all_self_mappings,
     combine_mappings
 )
-from vermouth.processors.apply_rubber_band import always_true, same_chain
 
 # TODO Since vermouth's __init__.py does some logging (KDTree), this may or may
 # not work as intended. Investigation required.
@@ -631,12 +630,12 @@ def entry():
     if args.elastic:
         LOGGER.info('Setting the rubber bands.', type='step')
         if args.rb_unit == 'molecule':
-            domain_criterion = always_true
+            domain_criterion = vermouth.processors.apply_rubber_band.always_true
         elif args.rb_unit == 'all':
             vermouth.MergeAllMolecules().run_system(system)
-            domain_criterion = always_true
+            domain_criterion = vermouth.processors.apply_rubber_band.always_true
         elif args.rb_unit == 'chain':
-            domain_criterion = same_chain
+            domain_criterion = vermouth.processors.apply_rubber_band.same_chain
         else:
             message = 'Unknown value for -eunit: "{}".'.format(args.rb_unit)
             LOGGER.critical(message)
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -358,7 +358,7 @@ def entry():
     rb_group.add_argument('-eu', dest='rb_upper_bound', type=float, default=0.9,
                           help='Elastic bond upper cutoff: F = 0  if rij > up')
     rb_group.add_argument('-ermd', dest='res_min_dist', type=int,
-                          help=('the minimum separation between two residues to have an RB'
+                          help=('The minimum separation between two residues to have an RB '
                                 'the default value is set by the force-field.'),
                           default=None)
     rb_group.add_argument('-ea', dest='rb_decay_factor', type=float, default=0,
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -357,6 +357,8 @@ def entry():
                           help='Elastic bond lower cutoff: F = Fc if rij < lo')
     rb_group.add_argument('-eu', dest='rb_upper_bound', type=float, default=0.9,
                           help='Elastic bond upper cutoff: F = 0  if rij > up')
+    rb_group.add_argument('-ermd', dest='res_min_dist', type=int,
+                          help='the minimum seperation between to residues to have an RB')
     rb_group.add_argument('-ea', dest='rb_decay_factor', type=float, default=0,
                           help='Elastic bond decay factor a')
     rb_group.add_argument('-ep', dest='rb_decay_power', type=float, default=0,
@@ -657,6 +659,7 @@ def entry():
             minimum_force=args.rb_minimum_force,
             selector=selector,
             domain_criterion=domain_criterion,
+            res_min_dist=args.res_min_dist
         )
         rubber_band_processor.run_system(system)
 
--- a/vermouth/processors/apply_rubber_band.py
+++ b/vermouth/processors/apply_rubber_band.py
@@ -226,7 +226,7 @@ def _apply_selection(graph, selection=None):
     if selection is None:
         selected_nodes = graph.nodes
     else:
-        selected_nodes = (node for node in graph.nodes if node in selection)
+        selected_nodes = [node for node in graph.nodes if node in selection]
 
     return selected_nodes
 
--- a/vermouth/processors/apply_rubber_band.py
+++ b/vermouth/processors/apply_rubber_band.py
@@ -213,7 +213,7 @@ def apply_rubber_band(molecule, selector,
                       lower_bound, upper_bound,
                       decay_factor, decay_power,
                       base_constant, minimum_force,
-                      bond_type, res_min_dist=3):
+                      bond_type, domain_criterion, res_min_dist=3):
     r"""
     Adds a rubber band elastic network to a molecule.
 
@@ -237,6 +237,12 @@ def apply_rubber_band(molecule, selector,
 
     The 'selector' argument takes a callback that accepts a atom dictionary and
     returns ``True`` if the atom match the conditions to be kept.
+
+    Only nodes that are in the same domain can be connected by the elastic
+    network. The 'domain_criterion' argument accepts a callback that determines
+    if two nodes are in the same domain. That callbacks accepts a graph and two
+    node keys as argument and returns whether or not the nodes are in the same
+    domain as a boolean.
 
     Parameters
     ----------
@@ -287,13 +293,13 @@ def apply_rubber_band(molecule, selector,
     constants = compute_force_constants(distance_matrix, lower_bound,
                                         upper_bound, decay_factor, decay_power,
                                         base_constant, minimum_force)
-    connectivity = build_connectivity_matrix(molecule, res_min_dist - 1,
-                                             selection=selection)
-    # Set the force constant to 0 for pairs that are connected. `connectivity`
-    # is a matrix of booleans that is True when a pair is connected. Because
-    # booleans acts as 0 or 1 in operation, we multiply the force constant
-    # matrix by the opposite (OR) of the connectivity matrix.
-    constants *= ~connectivity
+    connected = build_connectivity_matrix(molecule, res_min_dist - 1,
+                                          selection=selection)
+    same_domain = build_pair_matrix(molecule, domain_criterion,
+                                    selection=selection)
+    can_be_linked = (~connected) & same_domain
+    # Multiply the force constant by 0 if the nodes cannot be linked.
+    constants *= can_be_linked
     distance_matrix = distance_matrix.round(5)  # For compatibility with legacy
     for from_idx, to_idx in zip(*np.triu_indices_from(constants)):
         from_key = selection[from_idx]
@@ -309,12 +315,20 @@ def apply_rubber_band(molecule, selector,
             )
 
 
+def always_true(*args, **kwargs):
+    """
+    Returns ``True`` whatever the arguments are.
+    """
+    return True
+
+
 class ApplyRubberBand(Processor):
     def __init__(self, lower_bound, upper_bound, decay_factor, decay_power,
                  base_constant, minimum_force,
                  bond_type=None,
                  selector=selectors.select_backbone,
-                 bond_type_variable='elastic_network_bond_type'):
+                 bond_type_variable='elastic_network_bond_type',
+                 domain_criterion=always_true):
         super().__init__()
         self.lower_bound = lower_bound
         self.upper_bound = upper_bound
@@ -325,6 +339,7 @@ class ApplyRubberBand(Processor):
         self.bond_type = bond_type
         self.selector = selector
         self.bond_type_variable = bond_type_variable
+        self.domain_criterion = domain_criterion
 
     def run_molecule(self, molecule):
         # Choose the bond type. From high to low, the priority order is:
@@ -344,5 +359,6 @@ class ApplyRubberBand(Processor):
                           decay_power=self.decay_power,
                           base_constant=self.base_constant,
                           minimum_force=self.minimum_force,
-                          bond_type=bond_type)
+                          bond_type=bond_type,
+                          domain_criterion=self.domain_criterion)
         return molecule
--- a/vermouth/processors/apply_rubber_band.py
+++ b/vermouth/processors/apply_rubber_band.py
@@ -157,19 +157,16 @@ def build_connectivity_matrix(graph, separation, selection=None):
     return connectivity
 
 
-def build_domain_matrix(graph, are_same_domain, selection):
+def build_pair_matrix(graph, criterion, selection):
     """
-    Build a boolean matrix telling if two nodes belong to the same domain.
-
-    A domain is an ensemble of nodes that can be connected by an elastic
-    network.
+    Build a boolean matrix telling if a pair of nodes fulfil a criterion.
 
     Parameters
     ----------
     graph: networkx.Graph
         The graph/molecule to work on.
-    are_same_domain: Callable
-        A function that determines if two nodes are part of the same domain.
+    criterion: Callable
+        A function that determines if a pair of nodes fulfill the criterion.
         It takes a graph and two node keys as arguments and returns a boolean.
     selection: collections.abc.Collection
         A list of node keys to work on. If this argument is set, then the
@@ -190,7 +187,7 @@ def build_domain_matrix(graph, are_same_domain, selection):
     share_domain = np.zeros((size, size), dtype=bool)
     node_combinations = itertools.combinations(enumerate(selected_nodes), 2)
     for (idx, key_idx), (jdx, key_jdx) in node_combinations:
-        share_domain[idx, jdx] = are_same_domain(graph, key_idx, key_jdx)
+        share_domain[idx, jdx] = criterion(graph, key_idx, key_jdx)
         share_domain[jdx, idx] = share_domain[idx, jdx]
     return share_domain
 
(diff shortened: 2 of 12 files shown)


vermouth/processors/apply_rubber_band.py
```diff
@@ -129,35 +164,69 @@ def build_connectivity_matrix(graph, separation, selection=None):
...
+    # terms of edges, hence `separation + 1`.
+    distance_pairs = nx.all_pairs_shortest_path_length(graph, cutoff=separation + 1)
+    for origin, target_and_distances in distance_pairs:
+        for target in target_and_distances:
+            connectivity[correspondence[origin], correspondence[target]] = True
+    np.fill_diagonal(connectivity, False)
+    return connectivity[:, selection][selection]
+
+
+def build_pair_matrix(graph, criterion, selection):
```
pckroon: /s/selection/selection=None/

vermouth/processors/apply_rubber_band.py
```diff
@@ -227,13 +309,13 @@ def apply_rubber_band(molecule, selector,
...
-    # Set the force constant to 0 for pairs that are connected. `connectivity`
-    # is a matrix of booleans that is True when a pair is connected. Because
-    # booleans acts as 0 or 1 in operation, we multiply the force constant
-    # matrix by the oposite (OR) of the connectivity matrix.
-    constants *= ~connectivity
+    connected = build_connectivity_matrix(molecule, res_min_dist - 1,
+                                          selection=selection)
+    same_domain = build_pair_matrix(molecule, domain_criterion,
+                                    selection=selection)
+    can_be_linked = (~connected) & same_domain
```
pckroon: This only works because you enumerate over `graph.nodes` in the same order in both `build_connectivity_matrix` (correspondence) and `build_pair_matrix`. Fragile to say the least. Especially if 2 months from now we decide it's more efficient to iterate over `selection` in `build_pair_matrix` rather than over `graph.nodes`.
At the very least add a comment to those bits of code, but ideally, construct a correspondence dict elsewhere (e.g. here).

vermouth/processors/apply_rubber_band.py
```diff
@@ -249,12 +331,46 @@ def apply_rubber_band(molecule, selector,
...
+    """
+    return True
+
+
+def same_chain(graph, left, right):
+    """
+    Returns ``True`` is the nodes are part of the same chain.
+
+    Nodes are considered part of the same chain if they both have the same value
+    under the "chain" attribute, or if none of the 2 nodes have that attribute.
```
pckroon: /s/none/neither/

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -0,0 +1,126 @@
...
+        other_component + [0, 0, 1, 2, 0, 1, 2, 3],  # 8
+        other_component + [0, 0, 0, 1, 1, 2, 1, 2],  # 9
+        other_component + [1, 0, 0, 0, 2, 1, 0, 1],  # 10
+        other_component + [2, 1, 0, 0, 3, 2, 1, 0],  # 11
+        other_component + [0, 1, 2, 3, 0, 0, 1, 2],  # 12
+        other_component + [1, 2, 1, 2, 0, 0, 0, 1],  # 13
+        other_component + [2, 1, 0, 1, 1, 0, 0, 0],  # 14
+        other_component + [3, 2, 1, 0, 2, 1, 0, 0],  # 15
+    ])
+    np.fill_diagonal(distance, np.inf)
```
pckroon: This is the wrong place for this. This probably needs to be either in the fixture, or a call to e.g. `nx.shortest_path_length`

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -0,0 +1,126 @@
...
+    The graph is defined as having two domains: one per chain. The extra edges
+    allow to make sure the connectivity does not impact the domain detection.
+    """
+    def have_same_chain(graph, left, right):
+        return graph.nodes[left]['chain'] == graph.nodes[right]['chain']
+
+    expected = np.zeros((16, 16), dtype=bool)
+    expected[:8, :8] = True
+    expected[8:, 8:] = True
+    np.fill_diagonal(expected, False)
```
pckroon: Should this be needed?

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -0,0 +1,126 @@
...
+    Here, the criterion is ``True`` when nodes belong to the same domain.
+    The graph is defined as having two domains: one per chain. The extra edges
+    allow to make sure the connectivity does not impact the domain detection.
+    """
+    def have_same_chain(graph, left, right):
+        return graph.nodes[left]['chain'] == graph.nodes[right]['chain']
+
+    expected = np.zeros((16, 16), dtype=bool)
+    expected[:8, :8] = True
+    expected[8:, 8:] = True
```
pckroon: Belongs elsewhere? as above.

vermouth/processors/apply_rubber_band.py
```diff
@@ -129,35 +164,69 @@ def build_connectivity_matrix(graph, separation, selection=None):
...
+        # matrix.
+        return np.asarray(nx.to_numpy_matrix(graph, nodelist=selection).astype(bool))
+    if selection is None:
+        selection = slice(None, None, None)
+    size = graph.number_of_nodes()
+    correspondence = {node: index for index, node in enumerate(graph.nodes)}
+    connectivity = np.zeros((size, size), dtype=bool)
+    # separation is provided in term of nodes while the path is provided in
+    # terms of edges, hence `separation + 1`.
+    distance_pairs = nx.all_pairs_shortest_path_length(graph, cutoff=separation + 1)
```
pckroon: Note that this will only work for 1-bead-per-backbone models. Maybe run it through graph_utils.make_residue_graph?

bin/martinize2
```diff
@@ -604,6 +588,40 @@ def entry():
...
+            domain_criterion = always_true
+        elif args.rb_unit == 'all':
+            vermouth.MergeAllMolecules().run_system(system)
+            domain_criterion = always_true
+        elif args.rb_unit == 'chain':
+            domain_criterion = same_chain
+        else:
+            message = 'Unknown value for -eunit: "{}".'.format(args.rb_unit)
+            LOGGER.critical(message)
+            raise ValueError(message)
```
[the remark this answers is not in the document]
jbarnoud: This is there to catch errors by the dev such as adding the a keyword in argparse and forgetting it here.

vermouth/processors/apply_rubber_band.py
jbarnoud: It depends if the separation is taken as per particle or per residue. Both can be valid.

vermouth/processors/apply_rubber_band.py
```diff
@@ -129,35 +164,69 @@ def build_connectivity_matrix(graph, separation, selection=None):
...
+    for origin, target_and_distances in distance_pairs:
+        for target in target_and_distances:
+            connectivity[correspondence[origin], correspondence[target]] = True
+    np.fill_diagonal(connectivity, False)
+    return connectivity[:, selection][selection]
```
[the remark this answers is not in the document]
jbarnoud: Because it would return a flat array instead of a 2D one.

vermouth/processors/apply_rubber_band.py
pckroon: True. How do you want to define it? :)
Should be documented somewhere user-facing though.

vermouth/processors/apply_rubber_band.py
fgrunewald: Why don't we build the selection outside these two functions. We kind of do the work double here. Just have a function `selection_to_nodes(graph,  selection)` which returns the collection of nodes that is given to the two functions. In that way it will always be in the same order, without having to build any correspondence dict.

vermouth/processors/apply_rubber_band.py
```diff
@@ -90,11 +90,44 @@ def compute_force_constants(distance_matrix, lower_bound, upper_bound,
...
+    ----------
+    graph: networkx.Graph
+        The graph/molecule to work on.
+    left:
+        One node key from the graph.
+    right:
+        One node key from the graph.
+    separation: int
+        The maximum number of nodes in the shortest path between two nodes of
+        interest for these two nodes to be considered connected. Must be >= 0.
```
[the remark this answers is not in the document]
fgrunewald: I don't even get why you would ask this question? The domain of the shortest path function is >= 0, so if you ask for -1 you won't find any nodes which is OK. So the function works from that point of view also with <0 but I think it is still sensible to put this into the doc string.

vermouth/processors/apply_rubber_band.py
```diff
@@ -90,11 +90,44 @@ def compute_force_constants(distance_matrix, lower_bound, upper_bound,
...
+    -------
+    bool
+    """
+    nodes_are_connected = False
+    try:
+        shortest_path = len(nx.shortest_path(graph, left, right))
+    except nx.NetworkXNoPath:
+        # There is no path between left and right so they are not
+        # connected; which is the default.
+        pass
```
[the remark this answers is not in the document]
fgrunewald: I'm with Jon here, this seems more clearly readable in the way it is than using an inf float. Am I missing something here?

vermouth/processors/apply_rubber_band.py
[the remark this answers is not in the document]
fgrunewald: See Tsjerks comment:
> Because it would return a flat array instead of a 2D one.

vermouth/processors/apply_rubber_band.py
fgrunewald: @Tsjerk @jbarnoud how did martiniz1 deal with this? What was the default?

fgrunewald:
@jbarnoud I'm taking over this PR. It seems mostly done but we need to push a little with the impending release of martini3.

vermouth/processors/apply_rubber_band.py
jbarnoud: Martinize 1 would not be able to have more than 1 backbone bead.

jbarnoud:
Thanks, @fgrunewald. This are a bit hectic here.

vermouth/tests/test_apply_rubber_band.py
fgrunewald: @pckroon In what sense is this not needed?

vermouth/tests/test_apply_rubber_band.py
fgrunewald: But now it is there and Jon made it work ...

jbarnoud: I like my targets to be very explicit. It could go in a fixture, though.

mjtadema:
A quick remark, I just tried this branch with `-eunit all` and it seems to fail for me because the "selected_nodes" object becomes a generator while everywhere it's used it's assumed to be a list.
I narrowed down the culprit to this line producing the generator https://github.com/marrink-lab/vermouth-martinize/blob/d3a8342bf657d1664b87e814cb807e1e76035334/vermouth/processors/apply_rubber_band.py#L229 
I think it should be changed to `selected_nodes = [node for node in graph.nodes if node in selection]`. 
Unless there is some good reason why it is currently a generator?

fgrunewald:
> A quick remark, I just tried this branch with `-eunit all` and it seems to fail for me because the "selected_nodes" object becomes a generator while everywhere it's used it's assumed to be a list.
> I narrowed down the culprit to this line producing the generator
> 
> https://github.com/marrink-lab/vermouth-martinize/blob/d3a8342bf657d1664b87e814cb807e1e76035334/vermouth/processors/apply_rubber_band.py#L229
> 
> I think it should be changed to `selected_nodes = [node for node in graph.nodes if node in selection]`.
> Unless there is some good reason why it is currently a generator?

No good reason really. It's a leftover from refactoring that part of the code. We're still testing this branch though. So I wouldn't count on it producing a proper EN.

vermouth/processors/apply_rubber_band.py
fgrunewald: @jbarnoud why do we actually build selections for these functions? I just tested and it works if I simply take the one constructed here 

https://github.com/marrink-lab/vermouth-martinize/blob/d3a8342bf657d1664b87e814cb807e1e76035334/vermouth/processors/apply_rubber_band.py#L307-L315

jbarnoud: IIRC one selection is there to select the backbone (or whatever you want to apply the rubber band to), and one is there to see if it is the same chain.

vermouth/tests/test_apply_rubber_band.py
pckroon: It is needed indeed. We don't expect rubber bands between a node and itself.

vermouth/processors/apply_rubber_band.py
```diff
@@ -150,40 +150,38 @@ def build_connectivity_matrix(graph, separation, selection=None):
...
+        return np.asarray(nx.to_numpy_matrix(graph, nodelist=selected_nodes).astype(bool))
     size = graph.number_of_nodes()
     correspondence = {node: index for index, node in enumerate(graph.nodes)}
+    # the matrix will be reduced before returning it but nx.all_pairs_shortest_path_length
+    # does not take a subset of nodes
     connectivity = np.zeros((size, size), dtype=bool)
     # separation is provided in term of nodes while the path is provided in
     # terms of edges, hence `separation + 1`.
     distance_pairs = nx.all_pairs_shortest_path_length(graph, cutoff=separation + 1)
+    #only gets "positive" entries due to the cutoff argument above
```
pckroon, line 176: Space after #

vermouth/processors/apply_rubber_band.py
```diff
@@ -309,10 +322,14 @@ def apply_rubber_band(molecule, selector,
     constants = compute_force_constants(distance_matrix, lower_bound,
                                         upper_bound, decay_factor, decay_power,
                                         base_constant, minimum_force)
+    # we select the nodes here so the selection list has the same
+    # order in build_connectivity_matrix and build_pair matrix
+    selected_nodes = _apply_selection(molecule, selection=selection)
+    print(selected_nodes)
```
pckroon, line 328: leftover

vermouth/processors/apply_rubber_band.py
```diff
@@ -309,10 +322,14 @@ def apply_rubber_band(molecule, selector,
...
     connected = build_connectivity_matrix(molecule, res_min_dist - 1,
-                                          selection=selection)
+                                          selected_nodes=selected_nodes)
     same_domain = build_pair_matrix(molecule, domain_criterion,
-                                    selection=selection)
+                                    selected_nodes=selected_nodes)
```
pckroon: Since selected_nodes is used multiple times, it should definitely be a list or tuple. i.e. not a generator.

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -55,9 +55,10 @@ def disconnected_graph():
     return graph
 
 
+
+
```
pckroon, line 59: lint

vermouth/processors/apply_rubber_band.py
fgrunewald: done

vermouth/processors/apply_rubber_band.py
fgrunewald: @jbarnoud regarding this, you've already created some files for 2 BB bead protein models. Could some of those for testing? Otherwise I think this discussion is useless until someone wants to do a 2BB bead protein. So I'd follow Peter and just write it in the documentation somewhere. I think at this stage a CLI argument is not needed. We already have a ton for the EN.

jbarnoud: @fgrunewald GitHub does not let me attach my file so I sent it to you by email. You could look at DNA though.

jbarnoud: The mail did not go through. Here is a link https://www.dropbox.com/s/ivkjggnlapacze5/martini30bb.tar.gz?dl=0

fgrunewald: Thanks! I will give it a try and see how it goes. Probably we should also look at DNA again but I don't have a good reference yet. Let's see.

fgrunewald:
@jbarnoud Thanks for sharing the multi-backbone files. I think I discovered an additional problem with the EN code. In build connectivity matrix the bond separation is used as criterion to determine, if two nodes are eligible for a rubber band based on the so called 'res_min_dist' variable that is referred to as residue minimum distance. This works for single BB bead proteins where only the BB bead is used for the EN but it breaks down for literally anything else.  For example if I want to include the SC1 beads I suddenly decrease res_min_dist by one, because SC1 are 1 bond more separated from the BB beads. So instead of building the connectivity matrix in bond-space it should be build in residue space. Is that correct or am I missing something here?

jbarnoud:
> @jbarnoud Thanks for sharing the multi-backbone files. I think I discovered an additional problem with the EN code. In build connectivity matrix the bond separation is used as criterion to determine, if two nodes are eligible for a rubber band based on the so called 'res_min_dist' variable that is referred to as residue minimum distance. This works for single BB bead proteins where only the BB bead is used for the EN but it breaks down for literally anything else. For example if I want to include the SC1 beads I suddenly decrease res_min_dist by one, because SC1 are 1 bond more separated from the BB beads. So instead of building the connectivity matrix in bond-space it should be build in residue space. Is that correct or am I missing something here?

I would be careful. The res_min_dist is also there (and maybe mostly there) to assure you do not have elastic bonds between particles that would already be connected by a bonded interaction (e.g. dihedral angle). So if you add SC1, you should not reduce res_min_dist.

fgrunewald:
I think I haven't been clear enough. Imagine the following graph

2                  8
|                   |
1    4           7
|     |            |
0 - 3  -  5  - 6 

0,3,5,6 are BB and 1,4,7 SC1, and 2,8 SC2. Currently if I set res_min_dist = 2 (I know it should be 3 for martini but for sake of argument), I get a rubber band between 0 and 6, which is OK. But if I also include SC1 I get a rubber band between 0-6, 0-7, 1-7, 1-6 but also 0-4. The last one I would not expect to be there, because my residue minimum distance is 2.

jbarnoud:
I see your point. I guess the issue lies in what the minimum distance is actually meant to mean. Indeed both behaviour you describe can make sense depending on the definition. Because proteins never had more than one BB bead, there isn't really any expectations. I would look at what DNA does and go for that. If the issue does not exist for DNA, then I would go for the simplest: being keeping things as they are until it becomes an issue.

fgrunewald:
DNA is completely screwed. With he bond definition you get EN between VS and other atoms even within a residue. Makes sense right if you set the bond distance. Also produces incorrect Elnadyn itps. So I think we should fix this here and now. Although I really don't want to.

jbarnoud:
It should be possible to include the VS in the distance matrix by counting a VS relationship as a bond. How are the itp incorrect besides the VS issue?

fgrunewald:
> It should be possible to include the VS in the distance matrix by counting a VS relationship as a bond. How are the itp incorrect besides the VS issue?

You also get bonds within a nucleobase, because currently we don't use a residue criterion for constructing the connectivity matrix. I also talked to @paulocts about Martini 2,3 with EN, Elnadyn etc. and all are using the residue definition to establish which bonds are connected. There were some other issues with DNA as well but I think we solved most of them.

vermouth/processors/apply_rubber_band.py
fgrunewald: multi back-bone bead models are working now, although the default is still name BB

vermouth/processors/apply_rubber_band.py
pckroon: Still left over

fgrunewald:
>Should there be a test for 2 backbone bead models?<

How are they different from selecting more than one bead based on the bead name? I have pointed out that the backbone selector currently only captures the BB version. So if you want to have a multiple backbone bead model you'll need to select the beads manually via the appropriate flag.

bin/martinize2
```diff
@@ -357,6 +357,8 @@ def entry():
                           help='Elastic bond lower cutoff: F = Fc if rij < lo')
     rb_group.add_argument('-eu', dest='rb_upper_bound', type=float, default=0.9,
                           help='Elastic bond upper cutoff: F = 0  if rij > up')
+    rb_group.add_argument('-ermd', dest='res_min_dist', type=int,
+                          help='the minimum seperation between to residues to have an RB')
```
pckroon, line 361: Could use a comment to say the default comes from the FF

vermouth/processors/apply_rubber_band.py
```diff
@@ -21,9 +21,14 @@
...
 from .processor import Processor
 from .. import selectors
+from ..graph_utils import make_residue_graph
 
+# the bond type of the RB
 DEFAULT_BOND_TYPE = 6
-
+# the minimum distance between the resids
+# of two beads to have an RB
+# RMD = res_min_dist
```
pckroon, line 30: leftover?

vermouth/processors/apply_rubber_band.py
```diff
@@ -158,30 +163,24 @@ def build_connectivity_matrix(graph, separation, selected_nodes):
...
     """
-    if separation < 0:
-        raise ValueError('Separation has to be zero or positive.')
-    if separation == 0:
-        # The connectivity matrix with a separation of 0 is the adjacency
-        # matrix. Thankfully, networkx can directly give it to us a a numpy
-        # matrix.
-        return np.asarray(nx.to_numpy_matrix(graph, nodelist=selected_nodes).astype(bool))
+    res_graph = make_residue_graph(graph)
+    distance_pairs = nx.all_pairs_shortest_path_length(res_graph, cutoff=separation)
```
pckroon, line 167: What happens when make_residue_graph starts returning a DiGraph? I think this will still work, but the resulting matrix will simply become asymmetric.

vermouth/processors/apply_rubber_band.py
```diff
@@ -158,30 +163,24 @@ def build_connectivity_matrix(graph, separation, selected_nodes):
...
+    # only gets "positive" entries due to the cutoff argument above
     size = graph.number_of_nodes()
-    correspondence = {node: index for index, node in enumerate(graph.nodes)}
     # the matrix will be reduced before returning it but nx.all_pairs_shortest_path_length
     # does not take a subset of nodes
     connectivity = np.zeros((size, size), dtype=bool)
```
pckroon, line 172: TODO: optimize me, try to create a sparse matrix in case scipy is available.
Please add a comment

vermouth/processors/apply_rubber_band.py
```diff
@@ -158,30 +163,24 @@ def build_connectivity_matrix(graph, separation, selected_nodes):
...
-    for origin, target_and_distances in distance_pairs:
-        for target in target_and_distances:
-            connectivity[correspondence[origin], correspondence[target]] = True
+    for origin_residue, matchs_distances in distance_pairs:
+        for target_residue in matchs_distances:
+            origin_nodes = res_graph.nodes[origin_residue]['graph'].nodes()
+            target_nodes = res_graph.nodes[target_residue]['graph'].nodes()
+            for origin, target in itertools.product(origin_nodes, target_nodes):
+                connectivity[node_to_idx[origin], node_to_idx[target]] = True
     np.fill_diagonal(connectivity, False)
```
pckroon: You will still get elastic bonds within a residue (but not from a bead to itself). `all_pairs_shortest_path_length` will return the item `{1: {1: 0}}` as well.
Is this the desired behaviour?

vermouth/processors/apply_rubber_band.py
```diff
@@ -200,42 +199,21 @@ def build_pair_matrix(graph, criterion, selected_nodes):
     numpy.ndarray
         A boolean matrix.
     """
-    size = len(selected_nodes)
+    size = len(graph.nodes)
     share_domain = np.zeros((size, size), dtype=bool)
```
pckroon, line 203: comment: TODO optimize me.
As above.

vermouth/processors/apply_rubber_band.py
```diff
@@ -307,12 +285,18 @@ def apply_rubber_band(molecule, selector,
     selection = []
     coordinates = []
     missing = []
+    node_to_idx = {}
+    idx_to_node = {}
+    node_count = 0
     for node_key, attributes in molecule.nodes.items():
```
pckroon, line 291: `for node_count, (node_key, attributes) in enumerate(molecule.nodes.items()):`?
And you could replace node_count with node_idx, but no strong preference from my side

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -120,23 +81,175 @@ def have_same_chain(graph, left, right):
...
+# and separation in residue space. selection is tested at the
+# end with a more integral test
+def test_build_connectivity_matrix(disconnected_graph, separation, outcome):
+    idx_to_node = {idx: node for idx, node in enumerate(disconnected_graph.nodes)}
+    selection = list(range(0, 16, 2))
+    resid_attr = {0: 1, 1: 1, 2: 2, 3: 2, 4: 3, 5: 3, 6: 4, 7: 4,
+                  8: 5, 9: 5, 10: 6, 11: 6, 12: 7, 13: 7, 14: 8, 15: 8}
+    nx.set_node_attributes(disconnected_graph, resid_attr, 'resid')
+    connected = build_connectivity_matrix(disconnected_graph, separation, idx_to_node, selection)
+    pairs=[]
```
pckroon, line 106: lint

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -120,23 +81,175 @@ def have_same_chain(graph, left, right):
...
+    graph.add_edges_from(edges)
+    assert are_connected(graph, 1, 2, 1) == outcome
+
+@pytest.mark.parametrize('nodes, chain, edges, outcome', (
+    ([1, 2, 3],
+     {1:"A", 2:"A", 3:"C"},
+     [(1, 2), (2, 3)],
+     True),
+    ([1, 2, 3],
+     {1:"A", 2:"B", 3:"C"},
```
pckroon, line 138: lint

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -120,23 +81,175 @@ def have_same_chain(graph, left, right):
...
-    selected_nodes = apply_rubber_band._apply_selection(disconnected_graph, selection)
-    if selection == None:
-       selection = disconnected_graph.nodes
-    assert set(selected_nodes) == set(selection)
+
+    force_field = vermouth.forcefield.ForceField("test")
+    molecule = vermouth.molecule.Molecule(force_field=force_field)
+    molecule.meta['test'] = True
+    # The node keys should not be in a sorted order as it would mask any issue
+    # due to the keys being accidentally sorted.
```
pckroon, line 167: Nice, I like it 👍

vermouth/processors/apply_rubber_band.py
fgrunewald: @pckroon  yes it does return those but these are the ones on the diagonal and that is set to False after in line 179

vermouth/processors/apply_rubber_band.py
fgrunewald: yes

vermouth/processors/apply_rubber_band.py
pckroon: No, they're not :)

The numbers returned by `all_pairs_shortest_path_length` are node keys in the residue graph. The diagonal set to False is on a bead level.

fgrunewald: ahh yes I am stupid didn't think about it. 1) you don't get bonds within a residue, we have tested this a lot; 2) this is the connectivity matrix so things within one residue are connected. It get's inverted and then you have these beads set to false.

vermouth/processors/apply_rubber_band.py
fgrunewald: @pckroon well yes if the DiGraph is used you will get an asymmetric matrix and it will break probably 50% of the time. Is a directed residue graph sensible though?

pckroon: The only thing that would have to be changed then is the use of np.triu_indices, so that wouldn't be too bad.
And yes, very much. The most thorough discussion on it is probably in #57

bin/martinize2
```diff
@@ -358,7 +358,9 @@ def entry():
     rb_group.add_argument('-eu', dest='rb_upper_bound', type=float, default=0.9,
                           help='Elastic bond upper cutoff: F = 0  if rij > up')
     rb_group.add_argument('-ermd', dest='res_min_dist', type=int,
-                          help='the minimum seperation between to residues to have an RB')
+                          help=('the minimum separation between two residues to have an RB'
+                                'the default value is set by the force-field.'),
```
pckroon, line 362: Missing space, capital

vermouth/processors/apply_rubber_band.py
fgrunewald: @pckroon so we put a comment with it, first? Using the full matrix will probably slow it down a lot for large protein complexes.

pckroon: Leave it as is for now. The general structure still works, it'll just be a fast bughunt

PR #230, opened 2020-02-09
[WIP] Move the parsing infrastruture for ff files

jbarnoud:


vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+            # the way they are treated differently in the library, at least on how
+            # they generate edges. Here we move the all the impropers into their own
+            # [ impropers ] section.
+            propers = []
+            impropers = []
+            for dihedral in block.interactions.get('dihedrals', []):
+                if dihedral.parameters and dihedral.parameters[0] == '2':
+                    impropers.append(dihedral)
+                else:
+                    propers.append(dihedral)
```
pckroon: This makes the `[dihedral]` section magic special, since it'll be the only one (so far) where we interpret parameters. Not necessarily a bad thing, but something to realize/document.

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+            block.interactions['dihedrals'] = propers
+            block.interactions['impropers'] = impropers
+            block.make_edges_from_interactions()
+        for link in self.links:
+            link.make_edges_from_interactions()
```
pckroon: Don't we need the dihedral trick for Links? Or Modifications? The better place to address this would be in the parse_dihedral method, which I blindly assume exists.

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+        open_contexts = [
+            self.current_block, self.current_link, self.current_modification]
+        return open_contexts != ([None] * len(open_contexts))
+
+    def _new_block(self):
+        self.current_block = Block(force_field=self.force_field)
+
+    def _new_link(self):
+        self.current_link = Link(force_field=self.force_field)
+        self.links.append(self.current_link)
```
pckroon: Why not finish the existing block when making a new one? Why/how are links and blocks different?

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+
+    @SectionLineParser.section_parser('moleculetype', '*', context_type='block')
+    @SectionLineParser.section_parser('link', '*', context_type='link')
+    @SectionLineParser.section_parser('modification', '*', context_type='modification')
+    def _interactions(self, line, lineno=0, context_type=''):
+        context = self.get_context(context_type)
+        interaction_name = self.section[-1]
+        delete = False
+        if interaction_name.startswith('!'):
+            interaction_name = interaction_name[1:]
```
pckroon: Missing comment: "remove '!'"

vermouth/tests/test_ff_files.py
```diff
@@ -36,6 +36,22 @@ def test_moleculetype():
...
+        lines = """
+            [ moleculetype ]
+            GLY  3
+            
+            [ moleculetype ]
+            VAL 2
+            """
+        lines = textwrap.dedent(lines).splitlines()
+        ff = vermouth.forcefield.ForceField(name='test_ff')
+        vermouth.ffinput.read_ff(lines, ff)
```
pckroon: This (code) layout does mean that FFs cannot be named using an `info.ff` file. They instead will always get their name from the folder. Again, not bad, but it should be a deliberate choice.

vermouth/tests/test_ff_files.py
```diff
@@ -1041,7 +1057,7 @@ def test_variables():
     del ff.variables['float']
     assert ff.variables == {'integer': 1, 'string': 'mass',
                             'dictionary': {'a': 1, 'b': {'plop': 'toto'}}}
-
+test_variables()
```
pckroon: Strange leftover?

fgrunewald:
Hi guys, 

FYI I'm taking over this PR.

Cheers,
Fabian

jbarnoud:
Thanks. Sorry I couldn't finish it earlier. On 10 Mar 2020 10:21, fgrunewald <notifications@github.com> wrote:Hi guys,
FYI I'm taking over this PR.
Cheers,
Fabian

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+        ------
+        KeyError
+            If the section header is unknown.
+        """
+        prev_section = self.section
+
+        ended = []
+        section = self.section + [line.strip('[ ]').casefold()]
+        if tuple(section[-1:]) in self.METH_DICT:
+            self.section = section[-1:]
```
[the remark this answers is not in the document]
fgrunewald: using [-1:] keeps it a list which is sections assumed type

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+                ended.append(section.pop(-2))  # [a, b, c, d] -> [a, b, d]
+                wildcard_section = tuple(section[:-1] + ['*'])
+            self.section = section
+
+        result = None
+        if prev_section:
+            result = self.finalize_section(prev_section, ended)
+
+        action = self.header_actions.get(tuple(self.section))
+        if action is not None:
```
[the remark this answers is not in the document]
fgrunewald: Should be equivalent. I'll carry this one over to the new PR.

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+            The result returned by calling the registered method.
+        """
+        line = _substitute_macros(line, self.macros)
+        wildcard_section = tuple(self.section[:-1] + ['*'])
+        end_section = []
+        if self.section:
+            end_section = self.section[-1]
+        if tuple(self.section) in self.METH_DICT:
+            method, kwargs = self.METH_DICT[tuple(self.section)]
+        elif wildcard_section in self.METH_DICT and end_section not in self.METH_DICT:
```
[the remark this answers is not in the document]
fgrunewald: I'd agree. Also checking using the current test and force-field files this statement always evaluates to True. So it is redundant.

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+        try:
+            return method(self, line, lineno, **kwargs)
+        except Exception as error:
+            raise IOError("Problems parsing line {}. I think it should be a "
+                          "'{}' line, but I can't parse it as such."
+                          "".format(lineno, self.section)) from error
+
+    def finalize(self, lineno=0):
+        for block in self.blocks.values():
+            # Because of hos they are described in gromacs, proper and improper
```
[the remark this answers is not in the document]
fgrunewald: done

vermouth/ffinput.py
fgrunewald: @pckroon This is a result of what we discussed on the issue that the FF format and its parser currently are like a GROMACS itp format/parser but actually they are not. So if we make the FF parser a subclass of a proper GROMACS itp parser, we can also start interpreting the section without the danger. Then it should still be documented but at least there is a clear connection to the parent format so to speak. Therefore this will become part of the reordered parser hierarchy after #230 and  #236 are merged to master.

vermouth/ffinput.py
fgrunewald: Yes this is right. In fact the blocks get finished in the '[moleculetype]' line parser and the links here, all while the finalize_section method is completely empty. I suggest to move both to the finalize section method.

jbarnoud: Links do not need a name to be added to the list. Blocks are added to the dict as soon as the name is known.

fgrunewald: But is that necessary though? You can keep the self.current_block until the section is finished can't you? @pckroon  and I just spent some time tracking down to see how the information is actually added to the block dict. and the current setup makes it very obscure especially because links and blocks are treated differently and the finalize section method is completely empty.

jbarnoud:
I had issues with the finalise section being called at the beginning of the next subsection rather than at the beginning of the next section of the same level. Because I had things I wanted in the finalise section method (see the finalise one) I ended up avoiding finalise section. But do as you want, I was rushing when I wrote that. On 12 Mar 2020 15:01, fgrunewald <notifications@github.com> wrote:
@fgrunewald commented on this pull request.



In vermouth/ffinput.py:
> +    def _new_block(self):
+        self.current_block = Block(force_field=self.force_field)
+
+    def _new_link(self):
+        self.current_link = Link(force_field=self.force_field)
+        self.links.append(self.current_link)

But is that necessary though? You can keep the self.current_block until the section is finished can't you? @pckroon  and I just spent some time tracking down to see how the information is actually added to the block dict. and the current setup makes it very obscure especially because links and blocks are treated differently and the finalize section method is completely empty.

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+
+        return result
+
+    def parse_section(self, line, lineno):
+        """
+        Parse `line` with line number `lineno` by looking up the section in
+        :attr:`METH_DICT` and calling that method.
+
+        On the contrary to the SectionLineParser, we can have wildcard sections.
+        Wildcard sections cannot have children sections.
```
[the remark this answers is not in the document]
fgrunewald: @pckroon This parser doesn't do these type of .rtp style sections, which poses three new questions in my opinion:

1) why do we need a wild card parsing element anyways? Is it just because we want avoid writing a method for all bonds angles dihedrals generic interaction? In this case isn't this rather dangerous, because we read any generic interaction without actually knowing, if they are correct. For example typos are directly parsed through. 

2) Why is the rtp parser not a subclass of one of the generic parser classes (i.e. SecionLineParser)?

3) Why do some parsers live in the gmx folder and others on the main folder level of vermouth? Wouldn't it make more sense to have all parsers in one place, all writing utilities in one place. The question if they are GMX format or NAMD is in my opinion secondary.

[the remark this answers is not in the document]
pckroon: I agree with all your suggestions here. Let me get a bit more in depth.

1) For the ITP and FF parsers we don't. I strongly agree with the typo guarding by spelling out all allowed sections.
2) Because that didn't exist when the rtp parser was written.
3) I don't think it /really/ matters.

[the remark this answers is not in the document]
fgrunewald: @jbarnoud what's your take on issue 1?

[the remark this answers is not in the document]
jbarnoud: Issue 1 mostly came out of frustration from the first martinize: if gromacs comes out with a novel section I did not want martinize 2 to lag behind. Though, it happens rarely enough; typos are much more likely and the complexity the wildcard adds is probably not worth it.

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: done

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: done

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: done

vermouth/ffinput.py
fgrunewald: @pckroon Should we carry this one over to the new design base don an itp-reader , because it interprets GROMACS format?

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: @pckroon shall we remove the wildcard section then?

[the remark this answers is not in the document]
pckroon: yes.

vermouth/ffinput.py
pckroon: Sounds like a good idea to me.

PR #232, opened 2020-03-02
Added multiprocessing to MakeBonds to speed up bond creation

mjtadema:
So as i discussed with @pckroon, martinizing my system is rather slow.
I figured it must be possible to parallelize bond creation in separate molecules when you're really sure they should be separate.
This is what I am using right now, it seems to work fine for my system and it's much faster.
Probably the same thing can be done for modifying the termini which would make it even faster.

"Real" run times with GNU time:
* original 4m27,699s
* mp 1m5,754s

What are your thoughts? Is there a better way to do it? If you want to maybe actually include this in martinize i can change things to make it a bit neater.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -95,7 +95,8 @@ def read_system(path, ignore_resnames=(), ignh=None, modelidx=None):
 
 def pdb_to_universal(system, delete_unknown=False, force_field=None,
                      bonds_from_name=True, bonds_from_dist=True, bonds_fudge=1,
-                     write_graph=None, write_repair=None, write_canon=None):
+                     write_graph=None, write_repair=None, write_canon=None,
+                     threads=None):
     """
     Convert a system read from the PDB to a clean canonical atomistic system.
     """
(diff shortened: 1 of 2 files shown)


pckroon:
Hi Matthijs,

I don't like this as is, sorry. Based on your work and ideas I do have an idea on how to do this better though, so that's good. Could you open 2 new PRs with the following? Shouldn't be too much work based on what  you have here.
1) Add the multiprocessing bit to the `run_system` method of `vermouth/processors/processor.py`. That way you'd get free multiprocessing for *all* processors. It should be behind a CLI flag that sets the number of CPUs used (default 1). If that number is 1, no multiprocessing should be done. Note that threading is the wrong term.
2) Modify `make_bonds.py` in the following way: By default it's `run_system` method will call `nx.disjoint_union` to compose all molecules in the system in a single molecule (https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/processors/make_bonds.py#L255-L260). Then call `super().run_system`. The rest of the code in `MakeBonds.run_system` should be moved to `MakeBonds.run_molecule`. Alternatively, if a CLI flag is set, don't merge the molecules and just run `super().run_system`, which will effectively run `run_molecule` on all molecules separately.

Feel free to ask if you need help :)

mjtadema:
The way you describe it makes way more sense. I'll see what i can do.

PR #233, opened 2020-03-04
Processor multiprocessing

mjtadema:
This is what i came up with for multiprocessing in the base run_system method.
It will always fall back to the old method if `nproc` is not set.
So far it seems to work fine in combination with my changed make_bonds function in #234 .

### DIFF
--- a/vermouth/processors/processor.py
+++ b/vermouth/processors/processor.py
@@ -22,7 +22,9 @@ class Processor:
     """
     An abstract base class for processors. Subclasses must implement a
     `run_molecule` method.
+    Has nproc attribute that is changed by a CLI flag
     """
+    nproc = 1
     def run_system(self, system):
         """
         Process `system`.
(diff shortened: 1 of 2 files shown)


mjtadema:
It seems we need to add a test for `run_system` with multiprocessing, but I am not entirely sure where to put it..

mjtadema:
> Some small comments. The downside of this approach is that every Processor being run creates it's own Pool of processes, which might create a lot of overhead. I'm not sure how bad it is though.

Instead of setting the class attribute `nproc` and creating the pool in `run_system`, maybe we can set the attribute to be the actual pool itself. That way the same pool should be shared among all processors, no?

pckroon:
Yes, that would be an option.

fgrunewald:
stale PR; also it's so far behind it will probably be easier to make a new one.

PR #234, opened 2020-03-04
Makebonds molecule

mjtadema:
Here are my changes to make bonds. The CLI flag is also included here.
I added an argument group for experimental arguments, might come in handy for some other stuff as well. 
I tested this on my system in combination with #233 and i get the same behavior as before.
I didn't change any other processors yet but indeed it should be much easier now to implement multiprocessing in any other processor that has a run_molecule method.


mjtadema:
Related to #233

pckroon:
Is this superseded by #222?

PR #236, opened 2020-03-12
Refactor/ff parser

fgrunewald:
Took over #230 

closes #230


vermouth/ffinput.py
```diff
@@ -51,196 +55,275 @@
...
+        if tuple(section[-1:]) in self.METH_DICT:
+            self.section = section[-1:]
+        else:
+            wildcard_section = tuple(section[:-1] + ['*'])
+            while (tuple(section) not in self.METH_DICT
+                   and wildcard_section not in self.METH_DICT
+                   and len(section) > 1):
+                ended.append(section.pop(-2))  # [a, b, c, d] -> [a, b, d]
+                wildcard_section = tuple(section[:-1] + ['*'])
+            self.section = section
```
fgrunewald: This code in combination with 152 fails to actually do what it is supposed to do. Instead of dealing with wildcard sections of the type [ALA]['atoms'] it actually only deals with sections of type [moleculetype] [*], which seems a workaround to not enumerate all bonds angles etc. in METH_DICT. In fact it fails for reading the .rtp file in the universal force-field. Also, this is not at all caught by any of the tests.

fgrunewald: My bad, the read_ff parser is not supposed to read .rtp files. But obviously the rtp file parser is also not a subclass of any of the base parsers in parser_utils.py.

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+
+        self.header_actions = {
+            ('moleculetype', ): self._new_block,
+            ('link', ): self._new_link,
+            ('modification', ): self._new_modification,
+        }
+
+    def parse_header(self, line, lineno=0):
+        """
+        Parses a section header with line number `lineno`. Sets :attr:`section`
```
pckroon: The sphinx problem is that the class has no 'section' mentioned as one of it's Attributes in it's docstring (even though SectionLineParser does). Try changing the reference to ``:attr:`~SectionLineParser.section` ``. (or `vermouth.parser_utils.SectionLineParser.section`). Or something along those lines. I'll do a full review of your code asap.

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+            for dihedral in block.interactions.get('dihedrals', []):
+                if dihedral.parameters and dihedral.parameters[0] == '2':
+                    impropers.append(dihedral)
+                else:
+                    propers.append(dihedral)
+            block.interactions['dihedrals'] = propers
+            block.interactions['impropers'] = impropers
+            block.make_edges_from_interactions()
+        for link in self.links:
+            link.make_edges_from_interactions()
```
pckroon: Should be done in finalize_section

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+        self.force_field.blocks.update(self.blocks)
+        self.force_field.links.extend(self.links)
+        modifications = {mod.name: mod for mod in self.modifications}
+        self.force_field.modifications.update(modifications)
```
pckroon: Even the self.(blocks|links|modifications) you could remove in favour of just adding them to self.force_field as they're being parsed.

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+            'block': self.current_block,
+            'link': self.current_link,
+            'molmeta': self.current_link,
+            'modification': self.current_modification,
+        }
+        return possible_contexts[context_type]
+
+    def has_context(self):
+        open_contexts = [
+            self.current_block, self.current_link, self.current_modification]
```
pckroon: Not a great fan of this linebreak, but meh.

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+        return open_contexts != ([None] * len(open_contexts))
+
+    def _new_block(self):
+        self.current_block = Block(force_field=self.force_field)
+
+    def _new_link(self):
+        self.current_link = Link(force_field=self.force_field)
+#        self.links.append(self.current_link)
```
pckroon: Remove leftover comment

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+
+    def _new_modification(self):
+        self.current_modification = Link(force_field=self.force_field)
+        self.modifications.append(self.current_modification)
```
pckroon: Why not append modifications in the same place as blocks and links?

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+                          'before the blocks, links, and modifications.')
+        tokens = _tokenize(line)
+        _parse_variables(tokens, self.force_field, 'variables')
+
+    @SectionLineParser.section_parser('moleculetype')
+    def _block(self, line, lineno=0):
+        name, nrexcl = line.split()
+        self.current_block.name = name
+        self.current_block.nrexcl = int(nrexcl)
+#        self.blocks[self.current_block.name] = self.current_block
```
pckroon: Remove old comment

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+    @SectionLineParser.section_parser('variables')
+    def _variables(self, line, lineno=0):
+        if self.has_context():
+            raise IOError('The [variables] section must be defined '
+                          'before the blocks, links, and modifications.')
+        tokens = _tokenize(line)
+        _parse_variables(tokens, self.force_field, 'variables')
+
+    @SectionLineParser.section_parser('moleculetype')
+    def _block(self, line, lineno=0):
```
pckroon: Move the contents of _new_block to here? I didn't check where _new_block gets called though.

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+
+    @SectionLineParser.section_parser('moleculetype', 'patterns')
+    @SectionLineParser.section_parser('moleculetype', 'features')
+    @SectionLineParser.section_parser('moleculetype', 'non-edge')
+    @SectionLineParser.section_parser('modifications', 'non-edge')
+    def _invalid_out_of_link(self, line, lineno=0):
+        raise IOError('The "{}" section is only valid in links.'
+                      .format(self.section[-1]))
+
+    @SectionLineParser.section_parser('link', context_type='link')
```
pckroon: call _new_link here. Same comment as with _new_block above

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+    @SectionLineParser.section_parser('link', 'molmeta', context_type='molmeta')
+    def _link(self, line, lineno=0, context_type=''):
+        tokens = collections.deque(_tokenize(line))
+        _parse_link_attribute(tokens, self.current_link, context_type)
+
+    @SectionLineParser.section_parser('modification')
+    def _modification(self, line, lineno=0):
+        self.current_modification.name = line
```
pckroon: _new_modification here

vermouth/parser_utils.py
```diff
@@ -286,8 +286,18 @@ def is_section_header(line):
...
+        ------
+        IOError
+            The line starts like a section header but looks misformatted.
         """
-        return line.startswith('[') and line.endswith(']')
+        if line.startswith('[') :
+            if line.endswith(']'):
+                return True
+            else:
+                raise IOError('Section header looks misformatted.')
```
pckroon: Alternative:
```python
if line.startswith('[') or line.endswith(']'):
    if line.startwith('[') and line.endswith(']'):
        return True
    raise IOError('Misformatted section header')
return False
```

vermouth/tests/test_ff_files.py
```diff
@@ -1041,7 +1057,7 @@ def test_variables():
     del ff.variables['float']
     assert ff.variables == {'integer': 1, 'string': 'mass',
                             'dictionary': {'a': 1, 'b': {'plop': 'toto'}}}
-
+test_variables()
```
pckroon: This looks like it doesn't belong here

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+            The sections that have been ended.
+        """
+
+        # type comparison is what makes this work!
+        
+        if type(self.current_block) != type(None):
+           self.blocks[self.current_block.name] = self.current_block
+
+        if type(self.current_link) != type(None):
+           self.links.append(self.current_link)
```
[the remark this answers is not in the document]
fgrunewald: @pckroon Because a block is a molecule and a molecule without atoms/graph evaluates to none. More precisely overwriting the __eq__ method in molecule has lead to the behavior that a molecule without atoms evaluates to None. This means unless you make a type comparison you cannot have a block without atoms.  If we want that is a completely different question. I don't know enough about the use case but it is test no. 1 in the test_ff_files and appears in other more integrated tests as well. Perhaps we could rais an IO error for these cases?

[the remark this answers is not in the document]
pckroon: Just to make sure we're on the same page, when you say "evaluates to none", do you mean:
a) `bool(mol) == bool(None) == False`
b) `mol == None`
c) `mol is None`

If there's tests allowing empty blocks/molecules I suggest we keep supporting that.

vermouth/ffinput.py
fgrunewald: _new_block get's called as the section header is read. This parser actually parses the first line after the header. I would not move it, because the current setup is consistent with links and modifications, which cannot be moved to their respective sections  for the reason that link does not have to be followed by a line with information. Also it would somehow defy the current organization into a parser for headers and sections, where sections initialize the context and prep everything to be read.

vermouth/ffinput.py
fgrunewald: This will not work, because there are instances where you have:
`[link]`
`[atoms]`

In that case the method is not called.

vermouth/ffinput.py
fgrunewald: same as with links

vermouth/ffinput.py
pckroon: Ok, good point. Agreed.

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+        else:
+            while (tuple(section) not in self.METH_DICT
+                   and len(section) > 1):
+                ended.append(section.pop(-2))  # [a, b, c, d] -> [a, b, d]
+            self.section = section
+                          
+        result = None
+
+        if len(prev_section) != 0:
+            print(len(prev_section))
```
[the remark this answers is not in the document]
fgrunewald: done

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+            result = self.finalize_section(prev_section, ended)
```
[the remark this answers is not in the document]
fgrunewald: I'd say the first 3 lines, so it doesn't seem worth it

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+        #    method, kwargs = self.METH_DICT[wildcard_section]
+        else:
+            raise IOError("Can't parse line {} in section '{}' because the "
+                          "section is unknown".format(lineno, self.section))
+        try:
+            return method(self, line, lineno, **kwargs)
+        except Exception as error:
+            raise IOError("Problems parsing line {}. I think it should be a "
+                          "'{}' line, but I can't parse it as such."
+                          "".format(lineno, self.section)) from error
```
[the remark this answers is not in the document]
fgrunewald: In this case it could make sense to rewrite the parent method. I don't see directly why the parent is organized as it is. Also it is more a dummy method at the moment, which is only overwritten by the ffinput.

vermouth/parser_utils.py
fgrunewald: done

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: a) but not c) and b) actually causes an attribute error for empty molecules

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
fgrunewald: done

vermouth/ffinput.py
fgrunewald: this is actually quite nice because it makes the finilize method in ffinput completely obsolete.

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: I changed to the use of is.

vermouth/ffinput.py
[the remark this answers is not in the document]
fgrunewald: @pckroon what do you think about reorganizing the parent method?

vermouth/tests/test_ff_files.py
fgrunewald: why not?

vermouth/ffinput.py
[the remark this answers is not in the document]
pckroon: ok

vermouth/ffinput.py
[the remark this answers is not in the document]
pckroon: AFAIK the parent method does the same as this one. The only difference i can spot is that this one does something with `end_section`, which it then discards.

vermouth/tests/test_ff_files.py
pckroon: pytest invokes test methods. So there's absolutely no need to do so by hand.

vermouth/ffinput.py
```diff
@@ -51,196 +55,337 @@
...
+            propers = []
+            impropers = []
+            for dihedral in block.interactions.get('dihedrals', []):
+                if dihedral.parameters and dihedral.parameters[0] == '2':
+                    impropers.append(dihedral)
+                else:
+                    propers.append(dihedral)
+            block.interactions['dihedrals'] = propers
+            block.interactions['impropers'] = impropers
+            block.make_edges_from_interactions()
```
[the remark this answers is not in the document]
fgrunewald: done

vermouth/ffinput.py
```diff
@@ -337,6 +311,51 @@ def _interactions(self, line, lineno=0, context_type=''):
...
             )
 
+    @SectionLineParser.section_parser('moleculetype', 'dihedrals', context_type='block')
+    def _dih_interactions(self, line, lineno=0, context_type=''):
+          context = self.get_context(context_type)
+          interaction_name = self.section[-1]
+          delete = False
+          if interaction_name.startswith('!'):
+              interaction_name = interaction_name[1:]
+              delete = True
```
pckroon: This will always be false, unless you also add the '!dihedrals' section

Issue #235, opened 2020-03-06
read_ff assigns incorrect atoms for interaction type virtual_sitesn


fgrunewald:
Dear all, 

The parser for the force-field assigns incorrect atoms for the interaction type virtual_sitesn. The problem with virtual_sitesn as far as I can tell is the following:

read_ff assumes that for all GROMACS interactions a list of atoms comes first and is subsequently followed by parameters. This is not the case for this interaction. In this case you have first the atom for construction of the VS followed by an integer number, which defines the way the VS is constructed, followed by the number of atoms. 

I assume this hasn't been called because in the amino-acid definitions the convention with the '--' syntax is used and then probably the output parser fixes the order. Doesn't change the fact that the function isn't doing what it is supposed to do. 

Cheers,
Fabian

pckroon:
HI Fabian,

indeed virtual_sitesn seems a little funny. 
In `vermouth.gmx.itp.write_molecule_itp` the order gets shuffled again to what GMX expects.
```python3
                if name == 'virtual_sitesn':
                    to_join = [atoms[0], parameters] + atoms[1:]
```
My guess, based on the rest of the code is that we change the semantics of the section: first all atoms, then all the parameters. So, instead of e.g. `1 2 4 5 6` (atom 1 constructed by COM of 4 5 and 6), it would be `C1 C4 C5 C6 -- 2` (atomname C1 constructed by COM of C4 C5 and C6). Note that the `--` is needed to specify the end of atoms and the start of parameters, since it's impossible to know how many atoms to expect.

So, to get back to your analysis of the issue: the parser keeps the order as-is, and the writer puts it back to what GMX expects.

We should rename the section to something else to prevent mistakes, besides clear documentation.

pckroon:
Note that this will be resolved after #230 is merged.

fgrunewald:
still is incorrect when using just indices but this can be pointed out in a FF doc or otherwise one can use an itp reader, which will come soon. In fact it might be better to ban files based on indices all together and just have the FF format with atom names

fgrunewald:
So for vermouth FF-reader this will be the default behaviour and otherwise one can use the itp reader. I'll keep this in mind for the Documentation but the issue appears covered.

PR #237, opened 2020-03-19
Itp reader

fgrunewald:
This is the first version of an itp reader to tackle several issues we had with the ff-reader being unable to deal with certain GROMACS itp specifics. The itp reader could either form a base class for the ff reader. It could also be used to perform automatic checks of files written by vermouth. The itp reader has the following differences to the ff-reader:

- it deals correctly with VS supplied using the index method solving issue #235 
- it supports `#ifdef` statements in the correct places otherwise raises an IO error 
- it generates nodes based on atom indices thereby allowing for multiple atom-names in one block
- it has parsers for all GROMACS interactions
- it does not create edges from interactions at the moment; we can check if we would want this and which interactions to use and how to deal with VS
- it probably can parse various tokens from the ff format, but is not intended to. So these are unchecked. Because interpretation of these special tokens happens so deep in the base_parser it was much easier to just leave this in. 
- worth mentioning: it actually does split dihedrals into impropers and regular dihedrals 

At the moment it cannot deal with:
- `#include` statements
- `#define` statements as used in the topology to indicate, if something is defined. 
- `[settles]` directives are not read at all


fgrunewald:
@pckroon seems like some test fail for a reason I don't fully understand at the moment. Individually executed tests are fine. But when the full coverage is run some, some tests of test_ff_files fail. I suspect that this is an inheritance issue.

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        'position_restraints':1,
+        'distance_restraints':2,
+        'dihedral_restraints':4,
+        'orientation_restraints':2,
+        'angle_restraints':4,
+        'angle_restraints_z':2
+    }
+
+    def __init__(self, force_field):
+
```
pckroon: needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        Returns
+        -------
+        collections.abc.Callable
+            The method that should be used to parse `line`.
+        """
+
+        if self.is_section_header(line):
+            return self.parse_header
+        elif self.is_def(line):
+            return self.parse_def
```
pckroon: I wouldn't call it def, but pragma.

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        Parameters
+        ----------
+        line: str
+            A line of text.
+
+        Returns
+        -------
+        bool
+            ``True`` iff `line` is a def statement.
+
```
pckroon: needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        """
+        if line.startswith('#') :
+           return True
+        else:
+           return False
```
pckroon: `return line.startswith('#')`

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+            If the def sections are missformatted
+        """
+                        
+        if line == '#endif' and self.current_meta != None:
+           self.current_meta=None
+
+        elif line == '#endif' and self.current_meta == None:
+           raise IOError("Your #ifdef section is orderd incorrectly." 
+			 "At line {} I read #endif but I haven not read"
+                         "a ifdef before.".format(lineno)) from error
```
pckroon: The logic might be clearer if you do, e.g.
```python3
if line.startswith('#endif'):
    if self.current_meta == None:
        ...
    else:
        ...
else:
     if self.current_meta == None:
        ...
    else:
        ...
```

Also, add a guard against unknown pragma's, such as `#if`, or `#include`.

pckroon: Missing space for layout.
Also, NameError, error not defined.

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+
+
+        elif line != '#endif' and self.current_meta == None:
+           condition, tag = line.split()
+           self.current_meta = {'tag':tag,'condition':condition}
```
pckroon: missing spaces after :

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+
+        else:
+           raise IOError("Your #ifdef section is orderd incorrectly." 
+			 "At line {} I read #endif but I haven not read"
+                         "a ifdef before.".format(lineno)) from error
```
pckroon: Wrong error message, and same comment as before

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        object
+            The result of calling :meth:`finalize_section`, which is called
+            if a section ends.
+
+        Raises
+        ------
+        KeyError
+            If the section header is unknown.
+        """
+
```
pckroon: Needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        else:
+            while (tuple(section) not in self.METH_DICT
+                   and len(section) > 1):
+                ended.append(section.pop(-2))  # [a, b, c, d] -> [a, b, d]
+            self.section = section
+                          
+        result = None
+
+        if len(prev_section) != 0:
+            print(len(prev_section))
```
pckroon: Leftover debug

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+            result = self.finalize_section(prev_section, ended)
+
+        action = self.header_actions.get(tuple(self.section))
+        if action:
+            action()
+                           
+        return result
+
+    def finalize_section(self, pervious_section, ended_section):
+
```
pckroon: Needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+    @SectionLineParser.section_parser('moleculetype', 'bonds', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'angles', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'dihedrals', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'impropers', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'constraints', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'pairs', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'exclusions', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'virtual_sites2', context_type='block')
+    @SectionLineParser.section_parser('moleculetype', 'position_restraints', context_type='block')
+
```
pckroon: Needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+    def _interactions(self, line, lineno=0, context_type=''):
```
pckroon: This code will only ever process blocks, right? So factor out the context_type argument.

Also, shouldn't this create molecules?

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+
```
pckroon: Needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+            natoms=n_atoms,
+            delete=delete,
+            current_meta=self.current_meta
+            )
+   
+        # we need to reorder the atoms and interactions, because for virtual_sitesn
+        # the function type is on section position followed by an undefined number
+        # of atom indices from which the VS is built
+ 
+        if '--' not in line:
```
pckroon: `--` is not valid ITP syntax

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+           first_atom = context.interactions['virtual_sitesn'][-1].atoms[0]
+           other_atoms = context.interactions['virtual_sitesn'][-1].atoms[2:]
+           atoms = [first_atom] + other_atoms
+   
+           sec_atom = context.interactions['virtual_sitesn'][-1].atoms[1]
+           params = [sec_atom]+context.interactions['virtual_sitesn'][-1].parameters
+       
+           context.interactions['virtual_sitesn'][-1].atoms[:]=atoms
+           context.interactions['virtual_sitesn'][-1].parameters[:]=params
```
pckroon: Since you do the reshuffling *after* adding the interaction, could you try this with e.g.
```
[ atoms ]
1 ...
2 ...
3 ...

[ virtuals_sitesn ]
1 10 2 3
```

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+
+
+def _some_atoms_left(tokens, atoms, natoms):
+    """
+    Return True if the token list expected to contain atoms.
+
+    If the number of atoms is known before hand, then the function compares the
+    number of already found atoms to the expected number. If the '--' token if
```
pckroon: typo: if -> is

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+                         .format(token)) from error
+    modifications = {}
+    for key, value in attributes.items():
+        try:
+            if '|' in value:
+                modifications[key] = Choice(value.split('|'))
+        except TypeError:
+            pass
+    attributes.update(modifications)
+    return attributes
```
pckroon: Not relevant for ITP

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+
+def _parse_block_atom(tokens, context):
+
+    # deque does not support slicing
+    first_six = (tokens.popleft() for _ in range(6))
+    index, atype, resid, resname, name, charge_group = first_six
+    
+    if str(index) in context:
+        msg = ('There is already an atom named "{}" in the block "{}". '
+               'Atom names must be unique within a block.')
```
pckroon: error message no longer consistent with the actual check

vermouth/molecule.py
```diff
@@ -996,6 +996,24 @@ def add_atom(self, atom):
...
+        ----------
+        atom: collections.abc.Mapping
+            The attributes of the atom to add. Must contain 'atomname'
+
+        Raises
+        ------
+        ValueError
+            If `atom` does not contain 'atomname'
+        """
+        self.add_node(index, **atom)
```
pckroon: I don't see the added value of this method.

vermouth/parser_utils.py
```diff
@@ -177,6 +177,7 @@ def dispatch(self, line):
         collections.abc.Callable
             The method that should be used to parse `line`.
         """
+
```
pckroon: Needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+        raise IOError(msg.format(name, context.name))
+
+    atom = {
+        'atomname': name,
+        'atype': atype,
+        'resname': resname,
+        'resid': int(resid),
+        'charge_group': int(charge_group),
```
pckroon: Why not (also) add the index here?

vermouth/gmx/itp_read.py
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    # charge and mass are optional, but charge has to be defined for mass to be
+    if tokens:
+        atom['charge'] = float(tokens.popleft())
+    if tokens:
+        atom['mass'] = float(tokens.popleft())
+
+    attributes={}
+    context.add_atom_from_index(dict(collections.ChainMap(attributes, atom)), index=index)
```
pckroon: Shouldn't index be an int?

vermouth/tests/test_itp_files.py
```diff
@@ -0,0 +1,593 @@
...
+
+        [ atoms ]
+        1 P4 1 ALA BB 1
+        2 P3 1 ALA SC1 2
+        3 P2 1 ALA SC2 3
+
+        [ bonds ]
+        1 2 1 0.2 100
+        2 3 4 0.6 700
+        3 1 -- 9 plop toto
```
pckroon: Not valid ITP

vermouth/tests/test_itp_files.py
```diff
@@ -0,0 +1,593 @@
...
+            ),
+        ]
+        assert block.interactions['bonds'] == bonds
+
+
+    @staticmethod
+    def test_interaction_by_index():
+        """
+        Create an interaction and refer to atoms by index.
+        """
```
pckroon: What does this test differently than the previous?

fgrunewald:
> I think this is currently overcomplicated because of the "inherited" code.

fgrunewald:
@jbarnoud @pckroon I figured out why the ff_files_tests fail when the itp reader is added to vermouth and it does not look good. Basically what happens is the following:

Both the `FFDirector` object and the `ITPDirector` object are sub-classes of the `SectionLineParser` class. When `FFDirector` is imported the METH_DICT is populated with the methods tagged with  the `@SectionLineParser.section_parser` statement. But when you also subsequently import the `ITPDirector`, the methods from that subclass overwrite the methods of same name in the `FFDirector` object and also add methods with different name to `METH_DICT` of the `FFDirector`. This basically means the following works or `read_FF`:
`import vermouth.gmx.itp_read`
`import vermouth.ffinput`

But the following order will fail:
`import vermouth.ffinput`
`import vermouth.gmx.itp_read`

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: I think we did not conclusively discuss if ITPs should spawn molecules or blocks. This also raises the questions if then the ITPdirector should populate a system instead of a force-field. I think it is OK to consider an ITP to be a block. This would sort of keep it streamline with Martinize idea to build larger molecules from blocks. It has the disadvantage of the debate of index versus atomname for nodes. I can already tell you all my tests I did for the polymers tell me that both are exactly equivalent, meaning as long as links and blocks are both either name or index based the entire code just works, which is great news on its own.

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: yes this does indeed fail :/

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/parser_utils.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: sure why not

vermouth/tests/test_itp_files.py
fgrunewald: done

vermouth/tests/test_itp_files.py
fgrunewald: basically indices are used to define the atoms of interactions and within the FF context this check was checking for atom names.

vermouth/molecule.py
fgrunewald: If you want to decouple the atom-name from the node name to use an index as node name this is the cleanest way to do. Otherwise it becomes a shit show of having to rename the node, atomnames, and or interactions. In fact there are a bunch of checks and functions that prevent this or that directly delete corresponding interactions.

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/molecule.py
pckroon: I don't see the added value over just calling block.add_node(index, **attrs)

vermouth/gmx/itp_read.py
fgrunewald: done

vermouth/gmx/itp_read.py
fgrunewald: new implementation can parse VS as regular interaction

vermouth/gmx/itp_read.py
fgrunewald: I removed it

vermouth/molecule.py
fgrunewald: ahh yes that also works of course, but I thought this is more clear in comparison to the original parser. But I can remove it.

fgrunewald:
@pckroon the current version misses the codecov basically as a result of error messages being called in the test but as a 'stack' of errors. Basically the function parsing the line yields and IOError, which triggers a bunch of other IOErrors from functions that have called this function. The error messages seem sensible in the sense that you get the highest level error something like "Cannot parse this line" and then when you scroll up you see an error like "Atom reference not found". But for some reason I don't understand codecov considers the last type of lines raising the error as not covered by tests. Can you take a look? This probably also holds for the ffinput.py.

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+               # The indices in the file are 1-based
+                reference = int(reference) - 1
+                try:
+                    reference = atom_names[reference]
+                except IndexError:
+                    msg = ('There are {} atoms defined in the block "{}". '
+                           'Interaction in section "{}" cannot refer to '
+                           'atom index {}.')
+                    raise IOError(msg.format(len(context), context.name,
+                                             section, reference + 1))
```
pckroon: Although formally correct (so this is a suggestion), we can make the parser a bit more lenient by not enforcing that atoms are in order and sequential, so long as their indices are unique. Also, note that here you don't check that the index is actually what you expect it is.

I would suggest looking for the node with the correct ID in `context.nodes`. Which means you may want to create a dict of `atom_id: node_idx` beforehand.

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+                    msg = ('Atom names in blocks cannot be prefixed with + or -. '
+                           'The name "{}", used in section "{}" of the block "{}" '
+                           'is not valid in a block.')
+                    raise IOError(msg.format(reference, section, context.name))
+            all_references.append(reference)
+        return all_references
+
+
+    def _base_parser(self, tokens, context, section, atom_idxs):
+
```
pckroon: needless empty line

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+        # Normalize the atom references.
+        # Blocks and links treat these references differently.
+        # For blocks:
+        # * references can be written as indices or atom names
+        # * a reference cannot be prefixed by + or -
+        # * an interaction cannot create a new atom
+        # For links:
+        # * references must be atom names, but they can be prefixed with one or
+        #   more + or - to signify the order in the sequence
+        # * interactions create nodes
```
pckroon: Not really up to date, since links are not considered here.

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+
+        treated_atoms = self._treat_block_interaction_atoms(atoms, context, section)
+
+        if self.current_meta:
+            meta = {self.current_meta['condition']:self.current_meta['tag']}
```
pckroon: Missing space

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+            atoms=treated_atoms,
+            parameters=parameters,
+            meta=meta,)
 
-        context = self.current_block
-        interaction_name = self.section[-1]
-        delete = False
+        interaction_list = context.interactions.get(section, [])
+        interaction_list.append(interaction)
+        context.interactions[section] = interaction_list
```
pckroon: `context.interactions[section] = context.interactions.get(section, []) + [interaction]`?

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
-    return True
+        # deque does not support slicing
+        first_six = (tokens.popleft() for _ in range(6))
+        index, atype, resid, resname, name, charge_group = first_six
+        # since the index becomes the node name and all graphs start with 0
+        # index it makes more sense to also directly start at 0
 
+        if int(index) < 1:
+            msg = ('In section {} is a negative atom reference, which is not allowed.')
+            raise IOError(msg.format(section.name))
```
pckroon: I think I would decouple the atomid (number specified in the itp file) from the node index (the key in `block.nodes`)

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
 
-def _parse_atom_attributes(token):
-    """
-    Parse bracketed tokens.
+        index = int(index) -1
```
pckroon: missing space

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
-        return False
-    return True
+        # deque does not support slicing
+        first_six = (tokens.popleft() for _ in range(6))
+        index, atype, resid, resname, name, charge_group = first_six
+        # since the index becomes the node name and all graphs start with 0
+        # index it makes more sense to also directly start at 0
 
+        if int(index) < 1:
+            msg = ('In section {} is a negative atom reference, which is not allowed.')
```
pckroon: superfluous brackets

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
-
-def _get_atoms(tokens, natoms):
-    atoms = []
-    while tokens and _some_atoms_left(tokens, atoms, natoms):
-        token = tokens.popleft()
-        if token.startswith('{'):
-            msg = 'Found atom attributes without an atom reference.'
-            raise IOError(msg)
+        atom = {
+            # for bookkeping purposes let's keep the actual index i.e. +1 here
```
pckroon: typo

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+            'index'   : index +1,
```
pckroon: missing space

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
-        atom['mass'] = float(tokens.popleft())
-
-    attributes={}
-    context.add_atom_from_index(dict(collections.ChainMap(attributes, atom)), index=index)
+            atom['charge'] = float(tokens.popleft())
+        if tokens:
+            atom['mass'] = float(tokens.popleft())
+
+        attributes = {}
+        context.add_node(index, **dict(collections.ChainMap(attributes, atom)))
```
pckroon: No need to ChainMap if attributes is {}. Just `context.add_node(index, atom)` will do. But see above, I would generate a new number here, and use that as node index, rather than index directly.

vermouth/parser_utils.py
```diff
@@ -34,8 +34,8 @@ class SectionParser(type):
     """
     def __new__(mcs, name, bases, attrs, **kwargs):
         obj = super().__new__(mcs, name, bases, attrs, **kwargs)
-        if not hasattr(obj, 'METH_DICT'):
-            obj.METH_DICT = {}
+        #if not hasattr(obj, 'METH_DICT'):
+        obj.METH_DICT = {}
```
pckroon: As described in the corresponding issue, I'm reasonably sure this is only a partial fix. Leave this as-is, but leave the issue open.

vermouth/tests/test_itp_files.py
```diff
@@ -59,21 +59,21 @@ def test_atoms():
...
+        2 P3\t1\tALA  SC1 2 -3
         3 P2  1  ALA  SC2 3 -3 72
         """
         lines = textwrap.dedent(lines).splitlines()
         ff = vermouth.forcefield.ForceField(name='test_ff')
         vermouth.gmx.itp_read.read_itp(lines, ff)
         block = ff.blocks['GLY']
         assert len(block.nodes) == 3
-        assert block.nodes['1'] == {'atomname': 'BB', 'atype': 'P4',
+        assert block.nodes[0] == {'index':1,'atomname': 'BB', 'atype': 'P4',
```
pckroon: missing space

vermouth/tests/test_itp_files.py
```diff
@@ -59,21 +59,21 @@ def test_atoms():
...
                                      'resname': 'ALA', 'resid': 1,
                                      'charge_group': 1}
-        assert block.nodes['2'] == {'atomname': 'SC1', 'atype': 'P3',
+        assert block.nodes[1] == {'index':2,'atomname': 'SC1', 'atype': 'P3',
                                       'resname': 'ALA', 'resid': 1,
                                       'charge_group': 2, 'charge': -3.0}
-        assert block.nodes['3'] == {'atomname': 'SC2', 'atype': 'P2',
+        assert block.nodes[2] == {'index':3,'atomname': 'SC2', 'atype': 'P2',
```
pckroon: idem

vermouth/tests/test_itp_files.py
```diff
@@ -537,10 +397,40 @@ def test_defs(def_statements, bonds):
...
+         1 P4 1 ALA SC1 1
+         2 P4  1 ALA SC1 1
+         [ bonds ]
+         1   2 A B
+         #if
+         """))
     def test_def_fails(def_fail_statements):
         """
-        test if incorrectly formatted ifdefs raise 
+        test if incorrectly formatted ifdefs raise
```
pckroon: Not just ifdefs, but pragmas in general.

vermouth/tests/test_itp_files.py
```diff
@@ -573,21 +463,91 @@ def test_multiple_interactions():
...
+                atoms=[0, 1, 2, 3], parameters=['1','B'], meta={}),
               vermouth.molecule.Interaction(
-                atoms=['1','2','3','4'], parameters=['1','C'], meta={})]
-            
+                atoms=[0, 1, 2, 3], parameters=['1','C'], meta={})]
+
         assert ff.blocks['GLY'].interactions['dihedrals'] == dih
 
+    @staticmethod
+    def test_excluions():
```
pckroon: typo

vermouth/gmx/itp_read.py
```diff
@@ -18,67 +18,33 @@
...
+    atom_idxs = {'bonds'               : [0, 1],
+                 'position_restraints' : [0],
+                 'angles'              : [0, 1, 2],
+                 'constraints'         : [0, 1],
+                 'dihedrals'           : [0, 1, 2, 3],
+                 'pairs'               : [0, 1],
+                 'exclusions'          : [":"],
+                 'virtual_sitesn'      : [0, "2:"],
+                 'virtual_sites2'      : [0, 1, 2, 3],
+                 'virtual_sites3'      : [0, 1, 2, 3]}
```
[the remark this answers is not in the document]
fgrunewald: done

vermouth/gmx/itp_read.py
```diff
@@ -244,289 +208,187 @@ def _block(self, line, lineno=0):
...
+            if isinstance(idx, int):
+                atoms.append([tokens[idx], {}])
+                remove.append(idx)
+            elif isinstance(idx, str):
+                _slice = _string_to_slice(idx)
+                atoms += [[atom, {}] for atom in tokens[_slice]]
+                idx_range = range(0, len(tokens))
+                remove += [i for i in idx_range[_slice]]
+            else:
+                raise IOError
```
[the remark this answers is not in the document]
fgrunewald: @pckroon Can you clarify this a bit more? Basically you want to make the code slimmer? I'm not against it but this was a nightmare to hack out to begin with. I'm basically happy that it works ;)

vermouth/gmx/itp_read.py
fgrunewald: @pckroon not sure I understand this? So attributes needs to be a dict of what? If I simply plug in your suggestion I get an attribute error and I believe throughout the rest of the code the chain mapping is also used in these kinds of situations.

fgrunewald:
> Still a few small comments. Personally I would decouple the atomid (in the itp file) from the node index (key in block.nodes), so that you can choose to relax the parser a little, and no longer require that atomids are in order and sequential. Yes, I know it's not according to the itp spec, but it's still unambiguous. You may want to put that flexibility behind an option passed to the parser. It's up to you whether you want to put this in though.

I think we should not decouple the atomid from the node key. I understand your reasons but my points in favor of not doing so are:

- You can spot typos more easily, same as with the ff-files wildcard section, which we removed.
- In principle someone could assume that there is a correspondence, which would be legit because itp files adhere to this standard. 
- If you know programming, it is really easy to subcalss the itp reader to make your own and overwrite the functions that impose these restrictions. This is what I am doing for the polyply parser that requires dangling bonds. It is very little repetition. Thus I would keep the format as strict as possible and implement exceptions as subclass, which in fact would be a new file format.

vermouth/gmx/itp_read.py
```diff
@@ -123,9 +122,10 @@ def parse_pragma(self, line, lineno=0):
                 condition, tag = line.split()
                 self.current_meta = {'tag': tag, 'condition': condition.replace("#","")}
             elif self.current_meta is not None:
-                raise IOError("Your #ifdef section is orderd incorrectly."
-                              "At line {} I read #ifdef but there is still"
-                              "an open ifdef section from before.".format(lineno))
+                raise IOError("Your #ifdef/#ifndef section is orderd incorrectly."
+                              "At line {} I read {} but there is still"
+                              "an open #ifdef/#ifndef sectiona from"
```
pckroon, line 128: typo s/sectiona/section/

vermouth/gmx/itp_read.py
```diff
@@ -68,12 +68,13 @@ def dispatch(self, line):
 
         if self.is_section_header(line):
             return self.parse_header
-        elif self.is_pragma(line):
+        elif ITPDirector.is_pragma(line):
```
pckroon, line 71: Still call self.is_pragma

vermouth/gmx/itp_read.py
```diff
@@ -392,8 +399,8 @@ def _parse_block_atom(self, tokens, context):
         # index it makes more sense to also directly start at 0
 
         if int(index) < 1:
-            msg = ('In section {} is a negative atom reference, which is not allowed.')
-            raise IOError(msg.format(section.name))
+            msg = ('One of your atoms has a negative atom reference, which is not allowed.')
```
pckroon, line 403: Superfluous brackets

PR #242, opened 2020-04-03
List of protein-defining resnames now read from force field.

mnmelo:
The list of protein resnames is now parsed from the `protein_resnames` macro in the `.ff` files. This allows flexibility in what is processed as a protein.

For instance, I created a new residue `TMP` but its inclusion caused the molecule not to be recognized as a protein. It was therefore not dssp'ed, even though the residue has valid `N CA C O` and is parsed fine by DSSP.

With this modification I have the flexibility to define my own `universal` and `martini304` ffs, in which I set
```
[ macros ]
protein_resnames "TMP|GLY|ALA|CYS|VAL|LEU|ILE|MET|PRO|HYP|ASN|GLN|ASP|ASP0|GLU|GLU0|THR|SER|LYS|LYS0|ARG|ARG0|HIS|HISH|PHE|TYR|TRP|"
```
A nice side effect is that now stuff like `GLU0` and `HISH` work too if they're in the atomistic pdb, provided `protein_resnames` in the `universal` ff is taught about them.

Also did some minor cleaning and, of course, added a new(ish) quote (I had added it at some point to the old martinize; probably got lost in the translation; remove if you don't like it.)

No idea how best to design a test for this.


vermouth/selectors.py
```diff
@@ -20,31 +20,40 @@
...
-def is_protein(molecule):
+def is_protein(molecule, canonical=False):
     """
-    Return True if all the residues in the molecule are protein residues.
+    Return True if all the residues in `molecule` are protein residues.
 
     The function tests if the residue name of all the atoms in the input
-    molecule are in ``PROTEIN_RESIDUES``.
+    molecule are defined as such in the molecule's force field (under the
+    'protein_resnames' macro). If `canonical` is set to `True`, or if the
```
pckroon, line 35: Not a macro at this point, but an attribute.

pckroon:
For some reason Travis didn't report back, but it's here: https://www.travis-ci.org/github/marrink-lab/vermouth-martinize/builds/670397929
Tests for is_protein and selector failed due to this change.

vermouth/selectors.py
mnmelo: I really meant to say that it should be a macro in the .ff, because otherwise this feature is little documented elsewhere. I'll clarify.

mnmelo:
If it's set as an API detail, then it's on the user not to reuse the macro. We could instead have another section `[ protein_resnames ]`. Seems cleaner, but adds redundancy with the currently-employed macro, that is also used for `link` logic. 

Reading the resnames from the blocks seems a no-go: for my case I need the `universal` ff to have the new residue. I already add it to `aminoacids.rtp` but I can't rely on parsing that file because there's quite some non-protein stuff in there.

pckroon:
I'll think about it a bit. It sort of makes sense to have a separate (unique) section for it. On the other hand, there's indeed the option to add it to block, but then you indeed run into the issue that you're no longer reading .rtp. I'll get back to you with a decision on monday at the latest.

mnmelo:
Thanks. Found a way to add a test. Let me know when you decide.

pckroon:
My suggestion is to (besides block, link, and modification) introduce a `[ forcefield ]` section. This we can later on use to set e.g. FF names, or whether mapping should be done by CoG or CoM, and other FF wide options. For now, I suggest something like the following:
```
[ forcefield ]
[ protein_resnames ]
ALA GLY PHE
TYR SER ...
```
I'm open to other names and formats, so feel free to pitch in. The duplication with the $protein_resnames macro is unfortunate, but I suggest we leave it be for now. We can later decide whether we want to auto-generate that macro.

Could you also add #TODO comments regarding: 1) autogenerate the macro, 2) make sure the forcefield section is unique?

fgrunewald:
@pckroon whatever happened to this? I think there is still merit in somehow fixing this.

fgrunewald:
with #517 this has been solved and only needs to be implemented into the pipline to take effect

PR #247, opened 2020-04-07
Fix doc build failure

pckroon:
Docs failed due to duplicate description of `reference_graphs` attribute: once in the class docstring, and once in the property docstring.


pckroon:
Build currently failed since github pages are down for maintenance. And the networkx inventory file comes from there.
I'll restart the build later today.

PR #251, opened 2020-04-15
include all GROMACS types in the ffinput

### LINKED ISSUE #249: Missing GROMACS interactions in ffinput
Apparently DNA breaks with the current list of GROMACS interactions because we missed VS3 in the definitions. I'll list all gromacs interactions from the current version 2020 and have a fix later today.

fgrunewald:
@pckroon I have added all missing interaction types from the manual GROMACS 2020. I am not sure we will interpret all of them correctly but now the default behavior is the same as from before the refactoring. This will also fix #249 and the DNA building fail. 

Can you check that the dihedral interactions are located at the correct place? You are doing some stuff with the dihedral I do not fully understand something with the phase.


fgrunewald:
> Dihedrals and impropers should probably indeed be moved to the _dih_interactions method.

@pckroon  
- only from block or also links and modifications?
- I guess we don't have a test which shows where this matters?
- it basically only splits the dihedrals from a block into proper and imporoper and normal ones. So this is a convenience thing right?

pckroon:
For all.

I'll need to dig into the code to check the behaviour exactly. IIRC there was also some minor magic with the dihedral phases resulting in multiple interactions. And the creation of edges is pretty important for the application of links and modifications! I'll get back on this with more details tomorrow.

pckroon:
It does indeed seem to be only the improper/dihedral splitting, and as I read it the dihedral phase is not taken into account anywhere here (it just sets the interactions to whatever was just parsed).
The same splitting should be done for modifications and links though. Could you still add those, and then we can merge this as far as I'm concerned. There's not much point in adding tests until we know what the desired behaviour is.

PR #254, opened 2020-04-20
Fix #252 by making sure attributes from the residue_graph are propagated to reconstructed atoms

### LINKED ISSUE #252: DSSP error (duplicated nr. of residues)
I'm getting the following error. The protein has 156 residues. So it appears that at some step, they are being incorrectly duplciated.

```
Traceback (most recent call last):
  File "/lustre1/work/courtade/software/miniconda3/envs/py37/bin/martinize2", line 669, in <module>
    entry()
  File "/lustre1/work/courtade/software/miniconda3/envs/py37/bin/martinize2", line 545, in entry
    AnnotateDSSP(executable=args.dssp, savedir='.').run_system(system)
  File "/lustre1/work/courtade/software/miniconda3/envs/py37/lib/python3.7/site-packages/vermouth/processors/processor.py", line 38, in run_system
    mols.append(self.run_molecule(molecule))
  File "/lustre1/work/courtade/software/miniconda3/envs/py37/lib/python3.7/site-packages/vermouth/dssp/dssp.py", line 444, in run_molecule
    annotate_dssp(molecule, self.executable, self.savedir)
  File "/lustre1/work/courtade/software/miniconda3/envs/py37/lib/python3.7/site-packages/vermouth/dssp/dssp.py", line 277, in annotate_dssp
    annotate_residues_from_sequence(molecule, attribute, secstructs)
  File "/lustre1/work/courtade/software/miniconda3/envs/py37/lib/python3.7/site-packages/vermouth/dssp/dssp.py", line 390, in annotate_residues_from_sequence
    raise ValueError(msg.format(len(sequence), len(residues)))
ValueError: The sequence length does not match the number of residues. The sequence has 156 elements for 312 residues.
```
I tried with a different PDB file, it throws the same error.

Edit: The error was triggered by reading a PDB without hydrogens (incorrectly defined termini?). I managed to solve it by running pdb2gmx on the all-atom coordinates before martinizing

pckroon:
-

### DIFF
--- a/vermouth/processors/repair_graph.py
+++ b/vermouth/processors/repair_graph.py
@@ -193,6 +193,13 @@ def make_reference(mol):
         residue = residues.nodes[residx]['graph']
         reference = _get_reference_residue(residues.nodes[residx], mol.force_field)
         if 'mutation' in residues.nodes[residx]:
+            # We need to do this, since we need to propagate all common
+            # attributes in the input residue to any reconstructed atoms. 
+            # make_residue_graph already picks up on all the common attributes
+            # and sets those as residue/node attributes, but any mutation needs
+            # to be propagated as well, otherwise the newly constructed atoms
+            # will end up with the old residue name. This way, we can just add
+            # *all* attributes in the residue graph to the new atoms.
             resname = residues.nodes[residx]['mutation'][0]
             residues.nodes[residx]['resname'] = resname
         add_element_attr(reference)
@@ -274,6 +281,10 @@ def make_reference(mol):
         # "unsort" the matches
         match = {old_ref_names[ref]: old_res_names[res] for ref, res in match.items()}
 
+        # The residue graph has attributes which are common to all atoms in the
+        # input residue, so propagate those to new atoms. These attributes are
+        # things like residue name (see above in case of mutations), resid, 
+        # insertion code, etc.
         reference_graph.add_node(residx, reference=reference, found=residue, 
                                  match=match, **residues.nodes[residx])
     reference_graph.add_edges_from(residues.edges())
--- a/vermouth/processors/repair_graph.py
+++ b/vermouth/processors/repair_graph.py
@@ -194,6 +194,7 @@ def make_reference(mol):
         reference = _get_reference_residue(residues.nodes[residx], mol.force_field)
         if 'mutation' in residues.nodes[residx]:
             resname = residues.nodes[residx]['mutation'][0]
+            residues.nodes[residx]['resname'] = resname
         add_element_attr(reference)
         add_element_attr(residue)
         # We are going to sort the nodes of reference and residue by atomname.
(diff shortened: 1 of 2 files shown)


vermouth/processors/repair_graph.py
```diff
@@ -194,6 +194,7 @@ def make_reference(mol):
         reference = _get_reference_residue(residues.nodes[residx], mol.force_field)
         if 'mutation' in residues.nodes[residx]:
             resname = residues.nodes[residx]['mutation'][0]
+            residues.nodes[residx]['resname'] = resname
```
fgrunewald, line 197: Technically this works. But wouldn't it be more sane to apply mutations and modifications already in `make residue_graph`. The current fix seems a bit out of place here and it is not really clear, why the statement is there, unless you know that `get_reference_residue` also applies modifications and mutations. Plus if the resname was already change in the residue graph, `_get_reference_residue` would already get the proper residue and you only need to deal with modifications.  

Also do these modifications come from the pdb or the user command line input?

vermouth/tests/test_repair_graph.py
```diff
@@ -112,9 +112,9 @@ def build_system_mod(force_field):
         # Residue 2 has missing atoms (HA1 and HA2, the hydrogen of the C
         # alpha; HN the hydrogen on the N-ter nitrogen and O, the oxygen at the
         # C-ter)
-        (17, {'resid': 2, 'resname': 'GLY', 'atomname': 'N', 'chain': 'A', 'element': 'N'}),
-        (18, {'resid': 2, 'resname': 'GLY', 'atomname': 'CA', 'chain': 'A', 'element': 'C'}),
-        (19, {'resid': 2, 'resname': 'GLY', 'atomname': 'C', 'chain': 'A', 'element': 'C'}),
+        (17, {'resid': 2, 'resname': 'GLY', 'atomname': 'N', 'chain': 'A', 'element': 'N', 'common': 'a'}),
+        (18, {'resid': 2, 'resname': 'GLY', 'atomname': 'CA', 'chain': 'A', 'element': 'C', 'common': 'a'}),
+        (19, {'resid': 2, 'resname': 'GLY', 'atomname': 'C', 'chain': 'A', 'element': 'C', 'common': 'a'}),
```
fgrunewald, line 117: What does common a mean? Most elements are explained as comments, maybe add one for those as well. Something like "random attribute needed for checking attribtues"

vermouth/tests/test_repair_graph.py
```diff
@@ -273,7 +288,7 @@ def test_repair_graph_with_mutation_modification(system_mod, resid, mutations, m
     for node_idx in mol:
         if mol.nodes[node_idx].get('resid') == resid:
             if mutations:
-                assert mol.nodes[node_idx]['resname'] == mutations[0]
+                assert mol.nodes[node_idx]['resname'] == mutations[0], mol.nodes[node_idx]
```
fgrunewald, line 291: As far as I can tell this is a completely useless addition. Mol.nodes[node_idx] can be anything even a completely empty dictionary for this statement to assert without error.

pckroon: The second argument to assert gets printed on failure. But indeed it's left over bug-trace output. I'll remoev it

vermouth/processors/repair_graph.py
pckroon: I don't disagree that this may not be the best place. But, here's my main arguments against putting it in make_residue_graph: make_residue_graph is used in more places in the codebase, and is, in general, more generally applicable. THerefor it pays to keep that as simple as possible. I'll split it out into new functions though, that should simplify make_reference a bit.

The mutations and modifications come from the CLI at the moment. I guess they could also be added to the PDB parser, if there's sane syntax for it which I doubt.

pckroon: I couldn't find a decent way of splitting it up into smaller functions, so I added a pile of comments instead.

PR #260, opened 2020-05-13
Fix a bug where a block could only have one improper dihedral

pckroon:


jbarnoud:
Isn't it a data issue? So far the solution to not overwrite an interaction was to attach a value to the  `version` meta attribute. By default, each interaction has `version=0`. So this should be working without this PR:

```
[ dihedrals ]
BB SC1 SC2 SC3 1 1 2 3 4
BB SC1 SC2 SC3 1  {"version":1}
```
I agree this is annoying to write and error-prone. My concern is that this PR makes improper behave differently from the other interactions. If this is not an issue, then please test that links can replace the impropers.

vermouth/tests/test_ff_files.py
```diff
@@ -513,6 +513,68 @@ def test_param_effector_fail():
...
+            BB SC1 SC2 SC3 1 1 2 3 4
+            BB SC1 SC2 SC3 1
+            """,
+            {'dihedrals': [vermouth.molecule.Interaction(atoms='BB SC1 SC2 SC3'.split(),
+                                                         parameters='1 1 2 3 4'.split(),
+                                                         meta={}), 
+                           vermouth.molecule.Interaction(atoms='BB SC1 SC2 SC3'.split(),
+                                                         parameters=['1'],
+                                                         meta={})],
+             'impropers': []}
```
fgrunewald, line 543: Jon is right, at least this test should fail. The default behavior is interactions overwrite each other if they have the same atom names involved.

vermouth/tests/test_ff_files.py
```diff
@@ -513,6 +513,68 @@ def test_param_effector_fail():
...
+            """,
+            {'impropers': [vermouth.molecule.Interaction(atoms='BB SC1 SC2 SC3'.split(),
+                                                         parameters='2 1 2 3 4'.split(),
+                                                         meta={}),
+                           vermouth.molecule.Interaction(atoms='BB SC1 SC2 SC3'.split(),
+                                                         parameters=['2'],
+                                                         meta={})],
+             'dihedrals': []}
+        ],
+    ))
```
fgrunewald, line 559: same as above

vermouth/ffinput.py
```diff
@@ -359,7 +359,7 @@ def _dih_interactions(self, line, lineno=0, context_type=''):
           # [ impropers ] section.
 
         propers = []
-        impropers = []
+        impropers = context.interactions.get('impropers', [])
```
fgrunewald, line 362: I think this is OK. We should cache the already filtered impropers, which is implicitly done for dihedrals. I cannot remember which mechanism actually causes interactions of same atom names to be overwritten, but it is the default.

pckroon:
No, it's not a data issue. With every dihedral parsed `context.interactions['impropers']` would get reset to `[]`.
It is possible this got intoduced in refactorin the parser code though... I didn't track it down.

pckroon:
Well, this was encountered in the wild where a block had multiple impropers (for different atoms), and only the last got applied. The test cases may be misleading in this case. 
And apparently the interaction deduplication is broken.

fgrunewald:
@pckroon @jbarnoud I think we have a misunderstanding here. In fact also in the older version (0.3.1) interactions are read in duplicate (i.e. you can have duplicate section types and they end up in the blocks). However, when doing the links somehow they are removed or condensed, I can't remember at the moment. But it is not a parser feature that takes care of this. It is deeper in the code.

I also checked the current dev version for some proteins against the version from October (0.3.1) so before the new parser. And they yield except for some very minimal differences in one angle the same result. So what ever happened the program still works, also for complex cases. 

I suggest fix the CI stuff and then merge this PR. For investigating the duplication problem, we can make a new issue? When we write we should probably have a better idea of what can and cannot be done.

pckroon:
> I suggest fix the CI stuff and then merge this PR. For investigating the duplication problem, we can make a new issue? When we write we should probably have a better idea of what can and cannot be done.

I completely agree with this.

pckroon:
https://github.com/PyCQA/pylint/issues/2242
I suggest we leave the broken pylint CI for what it is for now. Hopefully it'll get fixed upstream soon.

fgrunewald:
@pckroon What's up with this one? I thought the idea was to remove the pylint stuff? We need to merge this before M3 implementation. So this is important

vermouth/tests/test_ff_files.py
```diff
@@ -513,6 +513,69 @@ def test_param_effector_fail():
...
+        [
+            """
+            [ dihedrals ]
+            BB SC1 SC2 SC3 1 1 2 3 4
+            BB SC1 SC2 SC4 1
+            """,
+            {'dihedrals': [vermouth.molecule.Interaction(atoms='BB SC1 SC2 SC3'.split(),
+                                                         parameters='1 1 2 3 4'.split(),
+                                                         meta={}), 
+                           vermouth.molecule.Interaction(atoms='BB SC1 SC2 SC4'.split(),
```
fgrunewald, line 540: It's very nitpicking but why use split here? It kind of masks that these are lists of parameters. I am with Jon to have targets as explicit as possible. But I am also fine keeping it like this.

Issue #262, opened 2020-05-21
-nter and -cter flags do not work for multi-chain molecules


pckroon:
For proteins that consist of multiple chains which are connected through e.g. a disulfide bond the -nter and -cter flags don't work. They'll only apply to the terminal residues with the lowest/highest resid. This means that only 2 out of 4 (in case of two chains) termini will get patched.
The selection of n/cter should be changed to compare the resid of the terminal residue to the resid of the neighbouring residue, rather than min/max.

pckroon:
#270 is a partial fix. If the temini also have disulfide bridges they are flagged as being a terminal residue.
The definitive solution would be to introduce chemical information, but that makes it FF specific.
Maybe sane warnings are enough: e.g. You specified a nter modification, but no matching residue was found.

fgrunewald:
@pckroon this was closed and reopened. What is the deal with it?

pckroon:
> #270 is a partial fix. If the temini also have disulfide bridges they are flagged as being a terminal residue.

PR #270, opened 2020-06-04
Detect N and C ter for automatic annotation of modification by comparing resid with their neighbour

pckroon:
This is a *partial* fix for #262. The remaining edge case is a two chain protein, in which chain terminal residues are connected through e.g. a disulfide bridge. Or a single chain protein which cyclizes through a non-peptide bond between the terminal residues.


vermouth/processors/annotate_mut_mod.py
```diff
@@ -105,22 +105,22 @@ def residue_matches(resspec, residue_graph, res_idx):
     res_node = residue_graph.nodes[res_idx]
     residue = {key: res_node.get(key)
                for key in 'chain resid resname insertion_code'.split()}
-    out = True
-    if resspec.get('resname', '')[-3:] == 'ter':
-        # Find all residues with degree 1: the ones with the lowest resid will
-        # be cter, the ones with the highest resid nter.
-        termini = [idx for idx in residue_graph if residue_graph.degree[idx] == 1]
-        get_resid = lambda idx: residue_graph.nodes[idx].get('resid')
+    if resspec.get('resname', '')[-3:] == 'ter' and residue_graph.degree[res_idx] == 1:
```
fgrunewald, line 108: Suppose I name a modification random-string-ter, the code would treat it as C or N terminus. This is something which becomes very hard to debug for the user and potentially us.  So I'd change the check to check for the complete name.

vermouth/processors/annotate_mut_mod.py
```diff
@@ -105,22 +105,22 @@ def residue_matches(resspec, residue_graph, res_idx):
...
+        # Find all residues with degree 1: the ones with a resid lower than
+        # their neighbour will be Nter, those with a resid higher than their
+        # neighbour Cter
         # FIXME: Once residue_graph is a digraph we can do something much much
         #        more clever, addressing arbitrarily branched polymers and
         #        termini
+        neighbour = list(residue_graph[res_idx])[0]  # Only one neighbour by definition.
+        get_resid = lambda idx: residue_graph.nodes[idx].get('resid', 0)
```
fgrunewald, line 116: I don't like these lambda constructs too much, but I leave that up to you.

vermouth/processors/annotate_mut_mod.py
pckroon: This gets caught in 121

pckroon:
I can;t find the lint on line 189

fgrunewald:
> I can;t find the lint on line 189

Perhaps a bit strict, but formerly 189 now line 185 pylint complains about too many characters on one line, 101 instead of 100. You're call if you wanna keep this.

vermouth/processors/annotate_mut_mod.py
fgrunewald: I know but I am still not convinced that this is a good choice, because fundamentally you are treating C and N terminus special and the code is also not general for any terminus that would get caught in this. It is only for proteins specifically following the standard convention of writing N to C terminus. So I'd be very much in favor of either making it completely special checking c-ter or n-ter or making it general. But as you write in the fixme the general solution will probably require DiGraphs.

pckroon:
I think that should be ok now, since the check is precise and explicit, right? Or did you have something else in mind?

Issue #256, opened 2020-04-27
Run is taking a long time with a medium-sized protein


alkihis:
Hi,

I'm using Martinize2 with vermouth 0.3.2.dev117 on a 2019 16" MacBook Pro.

After some runs, average step times for [BMRA-TRIM.pdb.zip](https://github.com/marrink-lab/vermouth-martinize/files/4539295/BMRA-TRIM.pdb.zip) are: 
~3-4mn for "Guessing the bonds", ~2mn for "Read input", ~30s for "Repairing the graph", ~15s for "Applying the links".
The other steps are sort of immediate.

This protein (less than 10k atoms) was "martinized" in seconds by old versions of martinize (before the martinize2 rewrite). 

Command line:
`martinize2 -f BMRA-TRIM.pdb -o system.top -x BMRA-CG.pdb`

Is those runtimes are expected?

Thank you,
Louis

jbarnoud:
We expect martinize 2 to run in matters of minutes, even for medium size proteins. This is the cost of the flexibility of this version.

The real issue is the regression mentioned in the edit. There should not be a doubling of the running time when upgrading to a newer version of martinize 2.

pckroon:
Also, we have not spent any time profiling and optimizing yet. Hopefully we can find the time to do that at some point.

pckroon:
Ok, I've done some profiling using the provided PDB. Most of the time (84%) is spent in networkx.quotient_graph, which gets called by our graph_utils.make_residue_graph (4 times, 60%) and make_bonds (1 time, 24%).
Those should be our first optimization targets I think.
I'm not super experienced with processing profiling data though. Maybe @jbarnoud can create more insights from this data? It contains the output produced by cProfile.
[profile.zip](https://github.com/marrink-lab/vermouth-martinize/files/4592253/profile.zip)

pckroon:
My first round of optimization resulted in https://github.com/marrink-lab/vermouth-martinize/tree/optimize. I've attached the profiling data again. [profile2.zip](https://github.com/marrink-lab/vermouth-martinize/files/4629374/profile2.zip)

Currently the isomorphism in do_mapping is the slowest (cumulative) component, followed by _bonds_from_distance. The latter is my next target.

@alkihis Do you see a similar improvement with this code on your end?

alkihis:
Hi, thank you for the good work. 
[...]

PR #271, opened 2020-06-05
First round of optimizations

pckroon:
This PR reduces runtime from >6 minutes to <1 minute on my laptop, using the protein from #256. 
Currently the slowest thing is the isomorphism in do_mapping. We can probably do something clever there as well, by checking whether any of the mappings cross residue boundaries, and simplifying the problem from there (per residue isomorphism, or even mapping by atom name). But that's significantly more work.

### DIFF
--- a/vermouth/processors/make_bonds.py
+++ b/vermouth/processors/make_bonds.py
@@ -25,7 +25,7 @@ from .. import KDTree
 from ..molecule import Molecule
 from .processor import Processor
 from ..utils import format_atom_string
-from ..graph_utils import collect_residues
+from ..graph_utils import collect_residues, partition_graph
 
 from ..log_helpers import StyleAdapter, get_logger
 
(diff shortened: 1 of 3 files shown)


vermouth/processors/make_bonds.py
```diff
@@ -94,7 +94,11 @@ def _bonds_from_distance(graph, nodes=None, non_edges=None, fudge=1.0):
             if subn in nodes and graph.nodes[subn].get('element') in VDW_RADII
         )
     }
-    max_dist = max(VDW_RADII.values()) * fudge
+    if idx_to_nodenum:
+        max_dist = max(VDW_RADII[graph.nodes[idx]['element']] for idx in idx_to_nodenum.values())
+    else:
+        max_dist = 0
+    max_dist *= fudge
```
fgrunewald, line 101: This change is also for the speed up or does that fix an error? Basically idx_to_nodenum is Flase if an element is not in VDW_radii? The previous functions is just horrible to read. So I don't really have a good grasp on what that dict actually contains.

pckroon: It's a speedup. It used to take the largest atom (Se currently) to determine the maximum possible cutoff for the pairs. Now it takes the largest atom present in the system.
And indeed. nodes that have an "unknown" element are not present in idx_to_nodenum, which maps indices in the coordinate array to node keys. The if/else guards the case where there are no atoms with known elements.

I'll add a comment.

PR #278, opened 2020-06-11
Implement stricter share_moltype

### LINKED ISSUE #276: Mutation of single chain produces only 1 itp
When mutating a amino-acid in one chain of a dimer or so, martinize2 produces a single itp-file with that mutation for both chains. When using -sep everything is fine, but the expected behavior should be to have two itp-files produced.

pckroon:
Fixes #276. 
Specific question for review: are there more node attributes that can differ between molecules and still result in identical ITPs?

### DIFF
--- a/vermouth/molecule.py
+++ b/vermouth/molecule.py
@@ -701,11 +701,21 @@ class Molecule(nx.Graph):
         bool
             True iff other has the same shape as this molecule.
         """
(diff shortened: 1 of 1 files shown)


pckroon:
I think this is the safest option, since it's most likely to flag molecules as different when they should be the same

PR #279, opened 2020-06-22
Change the DSSP processor to run on Windows. 

### LINKED ISSUE #207: DSSP errors hard to debug
Currently, if something goes wrong with DSSP it is really hard for the user to figure out exactly what went wrong for two reasons:
1) The pdb file we feed to DSSP is fed through /dev/stdin and can not be inspected afterwards. It should instead go through a (temporary) file that survives if errors happen, or `-vv` is specified from the command line
2) Error output from DSSP is very sparse.

The combination of the two makes it particularly hard why DSSP said e.g. "empty protein, or no valid complete residues", even though all residue names are normal/correct.

### LINKED ISSUE #253: run_dssp doesn't work on Windows
The DSSP processor doesn't work on windows, since it tries the following:
```python3
process = subprocess.Popen(
        [executable, "-i", "/dev/stdin"],
        stderr=subprocess.PIPE,
        stdout=subprocess.PIPE,
        stdin=subprocess.PIPE
    )
```
In other words, it tries telling dssp to read the protein from stdin. Windows being what it is, this doesn't work so well and blows up with a "No such file" error. 

My suggestion would be to write the atomistic structure to a temporary file, and feed that to dssp.

pckroon:
In addition, change some tests to be friendlier to systems with iffy PATH handling
fixes #253 
fixes #207

### DIFF
--- a/vermouth/tests/helper_functions.py
+++ b/vermouth/tests/helper_functions.py
@@ -16,6 +16,7 @@
 Contains helper functions for tests.
 """
 import operator
+import os
 import networkx.algorithms.isomorphism as iso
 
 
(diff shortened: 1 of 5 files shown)


vermouth/tests/helper_functions.py
```diff
@@ -57,3 +58,24 @@ def equal_graphs(g1, g2,
...
+        Names to look for in PATH.
+
+    Returns
+    -------
+    os.PathLike or None
+    """
+    for folder in os.getenv("PATH", '').split(os.pathsep):
+        for name in names:
+            fullpath = os.path.join(folder, name)
+            if os.path.isfile(fullpath):
```
jbarnoud, line 80: I think `which` on unix is looking for the first *executable* file. So there will be a small discrepancy here. We can probably live with it, though.

vermouth/tests/test_dssp.py
```diff
@@ -361,6 +364,37 @@ def test_run_dssp(savefile, tmpdir):
         assert not os.listdir(str(tmpdir))
 
 
+@pytest.mark.parametrize('pdb, loglevel,expected', [
+    (PDB_PROTEIN, 10, True),  # DEBUG
```
jbarnoud, line 368: It would be better to have the log level constants rather than their underlying numerical values.

PR #280, opened 2020-06-26
Remove any unknown atoms that are not matched to any known modification

### LINKED ISSUE #201: Increase usability by removing unrecognized atoms
Currently, if a PTM is not recognized martinize turns into a mushroom. This is not great. Instead, all atoms that are not recognized should be removed and warning(s) issued. This should either be done by CanonicalizeModifications, or a separate processor. In the second case, CanonicalizeModifications should mark all unknown atoms.

Similarly, the error message when RepairGraph can't find a reference block for a specific residue name should be better (IIRC). It's probably undesirable to remove complete unrecognized residues though.

pckroon:
Issue a warning when this happens

Minor linting
Fixes #201

### DIFF
--- a/vermouth/processors/canonicalize_modifications.py
+++ b/vermouth/processors/canonicalize_modifications.py
@@ -276,14 +276,18 @@ def fix_ptm(molecule):
         try:
             identified = identify_ptms(residue, res_ptms, options)
         except KeyError:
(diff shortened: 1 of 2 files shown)


pckroon:
No, nothing starting from PDB and ending with an ITP. The added test does makes sure the output is as expected though.

PR #281, opened 2020-06-26
Integration tests

### LINKED ISSUE #183: Tests need to run martinize2
#103 broke martinize2, this would not happen if martinize2 was executed during the tests.

### LINKED ISSUE #264: elnedyn and elnedyn22: mapping issues with THR and VAL
Tests with vermouth 0.3.2.dev122: I tried to generate some models with -ff elnedyn or -ff elnedyn22. THere are some mapping issues with THR and VAL. For other force fields is properly working.

### LINKED ISSUE #265: elnedyn22p: mapping issues with VAL
Tests with vermouth 0.3.2.dev122: I tried to generate some models with -ff elnedyn22p. There is some mapping issues with VAL residues. For other force fields as martini22 is properly working.

pckroon:
Sets up integration tests based on files provided by @paulocts

Fixes #183

Will need cleanup and rebasing before it's ready for review.


fgrunewald:
two small comments I noticed when doing the EN, which might be relevant here as well:

- direct comparison of interactions won't work because martinize has these ifdef statements (i.e. meta is populated differently) and also the force-constant is set as variable

- mapping of martinize1 and martinize2 is different so EN will not be the same. Generally for the martini2 stuff martinize1 is taken as reference as far as I understand

pckroon:
I rebased it to clean up the history a bit. It still requires 1 file (tier-1/bpti/martinize2/molecule_0.itp) to be checked by paulo

vermouth/tests/integration_tests/test_integration.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def assert_equal_blocks(block1, block2):
+    """
+    Asserts that two blocks are equal to gain the pytest rich comparisons,
+    which is lost when doing `assert block1 == block2`
+    """
+    assert block1.name == block2.name
+    assert block1.nrexcl == block2.nrexcl
+    # assert block1.force_field == block2.force_field  # Set to be equal
```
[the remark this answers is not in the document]
pckroon: I prefer to keep it to emphasize the difference

vermouth/tests/integration_tests/test_integration.py
```diff
@@ -0,0 +1,167 @@
...
+
+INTEGRATION_DATA = Path(datafiles.TEST_DATA/'integration_tests')
+
+PATTERN = '{path}/{tier}/{protein}/martinize2/'
+
+
+MARTINIZE2 = find_in_path()
+
+
+def assert_equal_blocks(block1, block2):
```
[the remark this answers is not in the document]
pckroon: I just tried, and it became a mess: `(pytest_assertion plugin: representation of details failed: .../lib/python3.6/site-packages/_pytest/assertion/util.py:377: ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all().`. The only way to make it work would be to basically reimplement the pytest reporting machinery, which seems rather out-of-scope.
Similarly for the 'distance' attribute on edges, and interactions, these seem plagued by rounding issues, and I can't figure out a way to make the output remotely useful.

vermouth/tests/integration_tests/test_integration.py
```diff
@@ -0,0 +1,167 @@
...
+    assert block1.name == block2.name
+    assert block1.nrexcl == block2.nrexcl
+    # assert block1.force_field == block2.force_field  # Set to be equal
+    # Assert the order to be equal as well...
+    # assert list(block1.nodes) == list(block2.nodes)
+    # ... as the attributes
+    assert OrderedDict(block1.nodes(data=True)) == OrderedDict(block2.nodes(data=True))
+    edges1 = {frozenset(e[:2]): e[2] for e in block1.edges(data=True)}
+    edges2 = {frozenset(e[:2]): e[2] for e in block2.edges(data=True)}
+    assert edges1 == edges2
```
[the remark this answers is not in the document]
pckroon: No go, since that trashes the valuable comparison from pytest.

vermouth/tests/integration_tests/test_integration.py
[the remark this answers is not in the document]
pckroon: Oh, adding a custom reporter for np.ndarray does not work. You'd need to get into the assertion rewriting, which is not possible https://github.com/pytest-dev/pytest/issues/5347

pckroon:
I removed the M1 directories. They were not used in the tests, and only in the hand crafting of the files in the M2 directories. Those files have been carefully vetted.
I don't see a feasible way of creating a custom pytest reporter for Molecule that gives me remotely comparable output to what I have now.

pckroon:
The low patch coverage is kind of expected, since the integration tests run in a separate interpreter, and the coverage they generate is not recorded.

vermouth/data/force_fields/elnedyn22/modifications.ff
```diff
@@ -22,3 +22,8 @@ BB {"replace": {"atype": "P5", "charge": 0}}
 NH2-ter
 [ atoms ]
 BB {"replace": {"atype": "P5", "charge": 0}}
+
+[ modification ]
+ASP0
+[ atoms ]
+SC1 {"resname": "ASP", "replace": {"atype": "P3", "charge": 0}}
```
fgrunewald, line 29: Is it clever to call it ASP0? I recall from talks with paulo that it should be AS0. In any case at the moment we use up to three or four naming convention for neutral residues. I would argue before merging we should unify these.

vermouth/data/mappings/elnedyn/elnedyn22/modifications.mapping
```diff
@@ -105,3 +105,28 @@ N   BB 0
...
+[modification]
+[ from ]
+universal
+[ to ]
+elnedyn22
+[ from blocks ]
+ASP0
+[ to blocks ]
+ASP0
+[ from nodes ]
```
fgrunewald, line 119: Again the naming convention should be checked.

vermouth/data/mappings/martini30/modifications.mapping
```diff
@@ -0,0 +1,150 @@
...
+N   BB
+CA  BB
+C   BB
+O   BB
+
+[ modification ]
+[ from ]
+universal
+[ to ]
+martini30dev
```
fgrunewald, line 108: Am I seeing it correctly that here m3-dev, m3-beta mappings are in one file? Shouldn't they be distributed according to the same logic as the force-field directory?

vermouth/processors/locate_charge_dummies.py
```diff
@@ -193,7 +192,6 @@ def locate_all_dummies(molecule, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
         dummy_keys = [pair[1] for pair in dummy_pairs]
         locate_dummy(molecule, anchor_key, dummy_keys, attribute_tag)
 
-
 class LocateChargeDummies(Processor):
     def __init__(self, attribute_tag=DEFAULT_DUMMY_ATTRIBUTE):
         super().__init__()
```
fgrunewald, line 197: This processor has no documentation on what it actually does apart form the technical aspect. This makes it impossible to review and also is not very sustainable as a package processor. I guess this is needed for the polarizeable side chains? @pckroon  can you provide some context.

fgrunewald (changes requested):
Overall looks good to me. Few small comments double checking some things. Also before minting the new release we should include all interactions in the ff-parser and itp parser.

bin/martinize2
```diff
@@ -354,7 +354,7 @@ def entry():
                           help='Write elastic bonds')
     rb_group.add_argument('-ef', dest='rb_force_constant', type=float, default=500,
                           help='Elastic bond force constant Fc in kJ/mol/nm^2')
-    rb_group.add_argument('-el', dest='rb_lower_bound', type=float, default=0.5,
+    rb_group.add_argument('-el', dest='rb_lower_bound', type=float, default=0,
```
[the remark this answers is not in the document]
jbarnoud: I checked and it is consistent with martinize (no idea why it wasn't before).

bin/martinize2
```diff
@@ -364,7 +364,7 @@ def entry():
                           default=None)
     rb_group.add_argument('-ea', dest='rb_decay_factor', type=float, default=0,
                           help='Elastic bond decay factor a')
-    rb_group.add_argument('-ep', dest='rb_decay_power', type=float, default=0,
+    rb_group.add_argument('-ep', dest='rb_decay_power', type=float, default=1,
```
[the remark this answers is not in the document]
jbarnoud: Same as above.

vermouth/processors/locate_charge_dummies.py
jbarnoud: Indeed, it deals with polarizable side chains.

vermouth/data/force_fields/elnedyn22/modifications.ff
pckroon: These are modifications, which do not end up in the PDB as resname (although it's possible to rename e.g. res ASP + mod ASP-H -> res AS0, should we so desire).
I suggest we decide on a naming convention at a later date, and leave it for now.

vermouth/data/mappings/martini30/modifications.mapping
pckroon: The mapping directory is a swamp that needs to be drained. But not now. We can decide on a sane folder structure once we take a look at the parsers.

vermouth/processors/locate_charge_dummies.py
pckroon: @fgrunewald Completely agree. The first order of business should be tests though, so this fix wouldn't have been needed in the first place. Not this PR or release though.

pckroon:
> Also before minting the new release we should include all interactions in the ff-parser and itp parser.

Disagree. A release is long overdue, and using not-yet-implemented interactions causes an understandable error message.

fgrunewald (approved):
Breaking out comments to issues this is good to go.

pckroon:
Created #299 #300 #301

PR #286, opened 2020-07-29
Implement VMD bonds-from-distance algorithm

### LINKED ISSUE #263: -cys auto is not working
My current tests with vermouth 0.3.2.dev122 and martini22 models showed that -cys auto ist not working. The extra bond between the cysteine bridge was not added in the example a tested. In martinize1 works well.

pckroon:
This increases the fudge factor to 20%, but prevents bonds being made between H atoms

Fixes #263

### DIFF
--- a/vermouth/processors/make_bonds.py
+++ b/vermouth/processors/make_bonds.py
@@ -64,16 +64,24 @@ def _bonds_from_distance(graph, nodes=None, non_edges=None, fudge=1.2):
     """Add edges to `graph` between `nodes` based on distance.
 
     Adds edges to `graph` between nodes in `nodes`, but will never add an edge
-    that is in `non_edges`, nor between H atoms. Edges are added based on a
-    simple distance criterion. The criterion can be adjusted using `fudge`.
-    Nodes need to have an element attribute that is in VDW_RADII in order to be
-    eligible.
+    that is in `non_edges`, nor between H atoms. It will also not create edges
+    where H atoms bridge separate residues. Residues are defined by the
+    '_res_serial' attribute of nodes.
+    Edges are added based on a simple distance criterion. The criterion can be
+    adjusted using `fudge`. Nodes need to have an element attribute that is in
+    VDW_RADII in order to be eligible.
 
     Parameters
     ----------
     graph: networkx.Graph
+        Nodes in the graph must have the attributes 'element', 'position', and
+        '_res_serial'.
     nodes: collections.abc.Collection[collections.abc.Hashable]
+        The nodes that should be considered for making edges. Must be in
+        `graph`.
     non_edges: collections.abc.Container[frozenset[collections.abc.Hashable, collections.abc.Hashable]]
+        A container of pairs of node keys between which no edge should be added,
+        even when they are close enough.
     fudge: float
     """
     if not nodes:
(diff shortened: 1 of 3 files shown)


pckroon:
Go and provide CONECT records ;)

pckroon:
I'm currently very unhappy with this PR 😞 
We need the increased fudge for SS bonds, which are now occasionally (regularly) missed. However, as-is too many bonds are made (for example between sidechain H and backbone N of a different residue). This again causes issues for the elastic network.

pckroon:
The implemented solution is to disallow bonds between residues involving hydrogens

vermouth/processors/make_bonds.py
```diff
@@ -60,19 +60,28 @@
...
+def _bonds_from_distance(graph, nodes=None, non_edges=None, fudge=1.2):
     """Add edges to `graph` between `nodes` based on distance.
 
     Adds edges to `graph` between nodes in `nodes`, but will never add an edge
-    that is in `non_edges`. Edges are added based on a simple distance
-    criterion. The criterion can be adjusted using `fudge`. Nodes need to have
-    an element attribute that is in VDW_RADII in order to be eligible.
+    that is in `non_edges`, nor between H atoms. It will also not create edges
+    where H atoms bridge separate residues. Residues are defined by the
+    '_res_serial' attribute of nodes.
```
fgrunewald, line 69: _res_serial is like a resid but continuous for all residues in the system?

vermouth/processors/make_bonds.py
```diff
@@ -128,6 +137,14 @@ def _bonds_from_distance(graph, nodes=None, non_edges=None, fudge=1.0):
...
+        resserial1 = atom1['_res_serial']
+        resserial2 = atom2['_res_serial']
+
+        # Forbid H-H bonds, and in addition, prevent hydrogens from making bonds
+        # to different residues.
+        if element1 == 'H' and element2 == 'H' or \
+                (resserial1 != resserial2 and (element1 == 'H' or element2 == 'H')):
+            continue
 
         bond_distance = 0.5 * (VDW_RADII[element1] + VDW_RADII[element2])
```
fgrunewald, line 149: Out of curiosity, what happens if I have a very strange atom-name or an incomplete element list in the pdb? Could I access the VDW_RADII or are they hard coded?

vermouth/processors/make_bonds.py
pckroon: Almost. _res_serial takes things like chain ID and insertion codes into account, and it thereby guaranteed to be unique for all residues.

vermouth/processors/make_bonds.py
pckroon: They're currently hardcoded, but should be moved to FF files.

pckroon:
> What happens if the element column diverges from the atom name column? Lately PCTS and I had to deal with some not so great pdb files hence the question about the use cases here.

Element takes precedence in determine the element. This is done by the PDB parser. If no element column is present the first letter (not number) of the atom name is taken as element. This is not ideal in case of e.g. Cl, but close enough for now.

> Could I prevent making of bonds between two residues and not need to provide all connect records?

Not really. You can reduce the fudge factor from the CLI until all superfluous bonds are gone, and then add any missing bonds via CONECT records.
This is actually the main problem I try to solve here, since most bonds crossing residues erroneously are caused by hydrogens.

PR #287, opened 2020-07-29
The great linting exercise of 2020

pckroon:
Due to the failed pylint run on #280 I decided to get some linting done.
Mostly fixing whitespace and indent issues, nothing serious.

### DIFF
--- a/vermouth/gmx/itp_read.py
+++ b/vermouth/gmx/itp_read.py
@@ -185,7 +185,7 @@ class ITPDirector(SectionLineParser):
 
         return result
 
-    def finalize_section(self, pervious_section, ended_section):
+    def finalize_section(self, previous_section, ended_section):
         """
         Called once a section is finished. It appends the current_links list
         to the links and update the block dictionary with current_block. Thereby it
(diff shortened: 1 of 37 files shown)


pckroon:
The sudden drop in pylint scores is due to the pylint message "E1102: pytest.mark.parametrize is not callable (not-callable)". This is obviously wrong, and caused by a version increase in pytest. This seems to be fixed in pytest upstream. However, since it would cause loud errors anyway, I've disabled it in the pylintrc

vermouth/gmx/itp_read.py
```diff
@@ -271,7 +271,8 @@ def _interactions(self, line, lineno=0):
             atom_idxs=atom_idxs,
             )
 
-    def _split_atoms_and_parameters(self, tokens, atom_idxs):
+    @staticmethod
+    def _split_atoms_and_parameters(tokens, atom_idxs):
         """
```
fgrunewald, line 276: I can't exactly remember why but for some inheritance reasons these were not static methods of the class. How badly do you wanna change this? I would advertise to come back to this as soon as we redo the parsers. I know it seems like almost an equivalent change but it had it's reasons.

fgrunewald: Same for the next three methods

pckroon: I don't think the staticmethod has any influence on the inheritance behaviour, but I'll change it.

PR #288, opened 2020-07-29
Quickfix the automatic enabling of elastic networks for elnedyn FFs

pckroon:


pckroon:
I'll ignore the pylint failure, since that's fixed in #287

PR #289, opened 2020-07-29
Sort written interactions by atomcount

pckroon:


vermouth/gmx/itp.py
```diff
@@ -184,10 +184,9 @@ def write_molecule_itp(molecule, outfile, header=(), moltype=None,
 
     # Write the interactions
     conditional_keys = {True: '#ifdef', False: '#ifndef'}
-    for name, interactions in molecule.interactions.items():
-        # Do not write an empty section.
-        if not interactions:
-            continue
```
fgrunewald, line 190: Since this got deleted, what happens if you have an empty section? I don't see immediately where that is prevented?

pckroon: Those get removed by `sort_interactions`. https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/molecule.py#L360

PR #292, opened 2020-07-31
add newline after writing gro file

fgrunewald:
I added as default behavior that the gro file has a newline after the box line. This will make sure VMD can directly open the file. Note that we will however not be able to trajectory files. But I think we don't do trajectories anyways.


pckroon:
Please also adapt the tests. Looks good to me otherwise.

PR #294, opened 2020-08-04
Automatically patch termini to be C-ter/N-ter

### LINKED ISSUE #179: Missing termini atoms not reconstructed
For a protein, if the termini are missing atoms (hydrogens on the N-ter, oxygen on the C-ter) they are not flagged as missing, and are not reconstructed. Solution would be to automatically add those modifications to terminal protein residues. Related to the `-nt` option of martinize2, and #15. 

Should be solved together with #15.

### LINKED ISSUE #248: Default termini patching
By default termini should be patched to nter:N-ter and cter:C-ter. The question is where this should be done. Ideally, this is done in the CLI/martinize2, but this will present challenges since that has no idea what a termini is: a CLI option to patch GLY1 should probably clobber the appropriate default terminus. 
An alternative would be to have the annotation processor set the defaults. This would require a bit of redesign where first all terminal residues are gathered and compared to the specified modifications.

Bonus points if we use the force field's definition of what a protein is in this (#242)

pckroon:
Fixes #248 
Fixes #179

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -530,9 +530,15 @@ def entry():
     if args.neutral_termini:
         args.modifications.append(['cter', 'COOH-ter'])
         args.modifications.append(['nter', 'NH2-ter'])
-    elif not args.modifications:
-        args.modifications.append(['cter', 'C-ter'])
-        args.modifications.append(['nter', 'N-ter'])
+    else:
+        if args.modifications:
+            resspecs, mods = zip(*args.modifications)
+        else:
+            resspecs, mods = [], []
+        if not any('cter' in resspec for resspec in resspecs):
+            args.modifications.append(['cter', 'C-ter'])
+        if not any('nter' in resspec for resspec in resspecs):
+            args.modifications.append(['nter', 'N-ter'])
 
     # Reading the input structure.
     # So far, we assume we only go from atomistic to martini. We want the
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -530,6 +530,9 @@ def entry():
     if args.neutral_termini:
         args.modifications.append(['cter', 'COOH-ter'])
         args.modifications.append(['nter', 'NH2-ter'])
+    elif not args.modifications:
+        args.modifications.append(['cter', 'C-ter'])
+        args.modifications.append(['nter', 'N-ter'])
 
     # Reading the input structure.
     # So far, we assume we only go from atomistic to martini. We want the


bin/martinize2
```diff
@@ -530,6 +530,9 @@ def entry():
     if args.neutral_termini:
         args.modifications.append(['cter', 'COOH-ter'])
         args.modifications.append(['nter', 'NH2-ter'])
+    elif not args.modifications:
+        args.modifications.append(['cter', 'C-ter'])
+        args.modifications.append(['nter', 'N-ter'])
```
fgrunewald, line 535: Two questions:

- the difference between COOH-ter and C-ter is that the later one is charged, right?
- what happens if I have modifications but not nter and ceter in those? Then I get mushroom? Shouldn't this be rather checking if nter and cter are in modifications and if not patch them.

fgrunewald: This PR seems to also fix spurious assignment of the protonation state of the termini.

pckroon: > the difference between COOH-ter and C-ter is that the later one is charged, right?

Correct

> what happens if I have modifications but not nter and ceter in those? Then I get mushroom? Shouldn't this be rather checking if nter and cter are in modifications and if not patch them.

Then your termini patching is detected from the PDB you provide. Checking whether cter/nter are in the list might be the better option, but you could also patch it by doing e.g. '-mod gly1:C-ter', or even '-mod A-cter:C-ter' (etc).

> This PR seems to also fix spurious assignment of the protonation state of the termini.

Magic :) What do you mean by spurious assignment in this case?

fgrunewald: > termini patching is detected from the PDB you provide 

I think this is not very clear to users because this is implicit. I'd argue it is better to check in the list of modifications and otherwise have an explicit flag maybe for reading it from pdb?

spurious assignment means in this case that martinize probably detects it from pdb but it is not clear to PCTS nor me when and how it does this and how many hydrogens need to be there etc. So basically sometimes you get neutral and sometimes charged termini.

pckroon: I'm not sure I agree that the automatic detection should be shielded behind a CLI flag, especially when termini getting "corrected" automatically. 
Currently the PTM detection logs e.g. the following for all modifications it detects:
```
    INFO - general - Identified the modifications ['NH2-ter'] on residues ['ASN1']
    INFO - general - Identified the modifications ['COOH-ter'] on residues ['SER20']
```
Does this message need to be improved?

fgrunewald: The question I am trying to ask: Is it clear enough for users how to set the portonation state of termini:

- by default they are charged
- with nt flag they are uncharged
- if I want to set them individually I use the modification syntax 
- if I want the pdb to define it I do what?

pckroon: > if I want to set them individually I use the modification syntax

or the -cter/-nter flags.

> if I want the pdb to define it I do what?

Hah! That's now impossible :') If you do '-cter none' you'll end up with a chemically nonsense, uncapped terminus. 
What's your opinion on a special "input" or "pdb" modification that explicitly says "guess it from the input structure"?
I think that can be a separate PR though

fgrunewald: I don't think many people will use it anyways, because pronation states often are an artifact of structure determination. So yeah let's make this an enhancement issue.

PR #305, opened 2020-09-18
Relax the PDB reader and writer to be more tolerant to missing attributes

pckroon:
Also makes charges numbers


vermouth/tests/pdb/test_read_pdb.py
```diff
@@ -182,3 +183,52 @@ def test_altloc(caplog):
...
+                'position': np.array([0.5207,  -0.4601,  -0.4552]),
+                'charge': 0, 'element': 'C'},
+            2: {'atomname': 'C05', 'atomid': 6, 'resid': 1,
+                'resname': 'UNK', 'chain': '', 'altloc': '',
+                'insertion_code': '', 'occupancy': 0.0, 'temp_factor': 0.0,
+                'position': np.array([0.5207, -0.4601, -0.4552]),
+                'charge': 0, 'element': 'C'}
+        }
+    ]
+    for mol, n_attrs in zip(mols, nodes):
```
fgrunewald, line 227: Can you be sure the order of moles is always the same as that of nodes?

vermouth/tests/pdb/test_read_pdb.py
```diff
@@ -182,3 +183,52 @@ def test_altloc(caplog):
...
+        for n_idx in mol.nodes:
+            for attr in mol.nodes[n_idx]:
```
fgrunewald, line 229: The problem with this line is that you are only checking the attributes that are in mol. It does not tell you if the pdb parser would in the future forget to parse an attribute, because something is messed up. It is better to also add an assert that the number of attributes for the parsed molecule and the actual molecule is exactly the same to be sure.

fgrunewald (changes requested):
Overall looks good. I just think one of the tests should be a bit tighter to make it more future proof.

vermouth/tests/pdb/test_read_pdb.py
pckroon: I think so. The first molecule in the PDB should also always be the first molecule in the system.

fgrunewald (approved):

PR #314, opened 2020-09-29
Limit the line length for the command in the top

jbarnoud:
Martinize2 writes the command it has been invoked with in
the header of the TOP file it writes. If that command is too
long, then it can pass the maximum line length that grompp
can read.

This commit truncate the command if it passes the limit.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -686,9 +686,17 @@ def entry():
         LOGGER.debug("Writing molecule {}.", molecule, type='step')
 
     # Write the topology if requested
+    # grompp has a limit in the number of character it can read per line
+    # (due to the size limit of a buffer somewhere in its implementation).
+    # The command line can be longer than this limit and therefore
+    # prevent grompp from reading the topology.
+    gromacs_char_limit = 4000  # the limit is actually 4095, but I play safe
+    command = ' '.join(sys.argv)
+    if len(command) > gromacs_char_limit:
+        command = command[:gromacs_char_limit] + ' ...'
     header = [
         'This file was generated using the following command:',
-        ' '.join(sys.argv),
+        command,
         VERSION,
     ]
     if None not in ss_sequence:


bin/martinize2
```diff
@@ -686,9 +686,17 @@ def entry():
...
 
     # Write the topology if requested
+    # grompp has a limit in the number of character it can read per line
+    # (due to the size limit of a buffer somewhere in its implementation).
+    # The command line can be longer than this limit and therefore
+    # prevent grompp from reading the topology.
+    gromacs_char_limit = 4000  # the limit is actually 4095, but I play safe
+    command = ' '.join(sys.argv)
+    if len(command) > gromacs_char_limit:
+        command = command[:gromacs_char_limit] + ' ...'
```
pckroon, line 696: Set it to 80, and linebreak properly. Keep printing the complete command, just linewrap it.
Also see: https://docs.python.org/3/library/textwrap.html

pckroon (changes requested):
> ...pass the maximum line length that grompp can read.

Are you kidding me? ><

jbarnoud:
I individually ignore all the ligands of the PDB for which we do not have a mapping, so it seems that my command line is longer than 4094 characters... Anyway, that odd corner case should not prevent the output file from being read.

bin/martinize2
jbarnoud: I considered that, but:
1. it is a fairly rare corner case so it should not affect the general case
2. the line, once wrapped, will make a header that will be very long and push the relevant information (the topology) far bellow
3. the line can be easily copy-pasted, which is not the case if it is wrapped because the comment character will pollute the text block; I am fine with if a pathologic line cannot be copied, but I'd rather not affect the legit ones.

pckroon: Good arguments. Approved.

pckroon (approved):

PR #316, opened 2020-10-02
Allow to ignore specific warnings

### LINKED ISSUE #313: CLI option to ignore alternate locations in PDB
Currently, if there are alternate locations in a PDB input, then martinize issues a warning for each alternate location it encounters that is not `A` or empty. This means that if a PDB file uses alternate locations, then many warnings are issued, the number of which is only predictable by analysing the input file. These warnings are difficult to deal with in scripts because ignoring just these warnings means one needs to analyse the input files or run martinize twice to know how many warnings to ignore. This can easily lead to `-maxwarn 9999` being common in scripts.

A solution would be an option to choose the alternate, but this option is unlikely to be used. A simpler option is one that ignores alternates other than `A` or empty. This is the behaviour the most likely to be useful. I suggest `-ignore-alternate` with a default to `False`.

jbarnoud:
Fixes #313 by implementing the way of ignoring specific warning types discussed in the issue.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -271,9 +271,35 @@ def _cys_argument(value):
 
 
 def maxwarn(value):
+    """
+    Given a maxwarn specification, split it in a warning type, and the number
+    to ignore.
+
+    >>> maxwarn('3')
+    (None, 3)
+    >>> maxwarn('general:15')
+    ('general', 15)
+    >>> maxwarn('inconsistent-data')
+    ('inconsistent-data, None)
+
+    Parameters
+    ----------
+    value: str
+        A warning type and a count, separated by a colon.
+
+    Returns
+    -------
+    tuple[str, int]
+        A warning type and the associated count to ignore. Either element can be
+        None if not specified.
+
+    Raises
+    ------
+    argparse.ArgumentTypeError
+    """
     msg = ("Values for the -maxwarn option must be the name of a "
            "warning type, a number, or following the format "
-           "'<warning-type>:<count>' where <warning-type> if the name "
+           "'<warning-type>:<count>' where <warning-type> is the name "
            "of the warning type to ignore, and <count> is the number of "
            "warning of that type to ignore. "
            "'{value}' is not a valid value.".format(value=value))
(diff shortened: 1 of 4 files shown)


bin/martinize2
```diff
@@ -269,6 +270,31 @@ def _cys_argument(value):
         return result
 
 
+def maxwarn(value):
+    msg = ("Values for the -maxwarn option must be the name of a "
+           "warning type, a number, or following the format "
+           "'<warning-type>:<count>' where <warning-type> if the name "
```
pckroon, line 276: s/if/is/

vermouth/log_helpers.py
```diff
@@ -241,3 +240,39 @@ def get_logger(name):
...
+                deduct_all.add(warning_type)
+            else:
+                specs[warning_type] = max(specs.get(warning_type, 0), count)
+    blanket_ignore = specs.get(None, 0)
+    warning_count = counter.counts[level]
+    total = number_of_warnings
+    for warning_type, count in warning_count.items():
+        type_count = warning_count[warning_type]
+        if warning_type in specs:
+            total -= max(0, min(count, specs[warning_type]))
```
pckroon, line 272: Could use a comment

bin/martinize2
```diff
@@ -269,6 +270,31 @@ def _cys_argument(value):
...
+           "of the warning type to ignore, and <count> is the number of "
+           "warning of that type to ignore. "
+           "'{value}' is not a valid value.".format(value=value))
+    splitted = value.split(':')
+    if len(splitted) == 1:
+        try:
+            count = int(value)
+        except ValueError:
+            return (value, None)
```
pckroon, line 285: Needs a comment

pckroon (approved):
Some nitpicks, but also good enough as-is

pckroon (approved):

jbarnoud:
> Address my own comments, add docstring for maxwarn function

Sorry, I forgot about the PR.

pckroon:
Happens to the best of us :)

I'll leave it to you to press merge once travis is happy

PR #322, opened 2020-10-16
Molecule grouper and water mapping

pckroon:
Implements a constrained Kmeans algorithm to cluster water molecules together to form CG water beads. Probably needs a bit of a polish/field test before it's really ready for deployment...


vermouth/selectors.py
```diff
@@ -48,6 +48,12 @@ def is_protein(molecule):
         for n_idx in molecule
     )
 
+def is_water(molecule):
+    # TODO: different water models.
+    return all(
+        molecule.nodes[n_idx].get('resname') == 'HOH'
+        for n_idx in molecule
+    )
```
fgrunewald, line 56: Can't we make this more general and simply ask the user for a resname and then a resolution (i.e. 1 to 4 or 2 to 4). The method should be general shouldn't it? I'm thinking of material science use cases where this could be very useful and this comes with very little additional code needed.

vermouth/processors/molecule_grouper.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Returns
+    -------
+    list
+    """
+    try:
+        len(num_or_list)
+    except TypeError:
+        num_or_list = [num_or_list] * length
+    return num_or_list
```
fgrunewald, line 47: Is this really needed. This function is only used twice in consecutive lines and very specific to the use case?

vermouth/processors/molecule_grouper.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+        init = np.append(clusters, [[0, 0, 0]], axis=0)
+        results.append([cost, clusters, memberships, niter])
+
+    cost, clusters, memberships, niter = min(results, key=lambda i: i[0])
+    LOGGER.info('Clustered {} water molecules into {} clusters.', len(water_mols), num_clusters)
+    counts = np.count_nonzero(memberships, axis=0)
+    LOGGER.info('Minimum/maximum number of molecules per cluster: {}/{}', counts.min(), counts.max())
+    for mol_idx in sorted(mol_idxs, reverse=True):
+        del system.molecules[mol_idx]
```
fgrunewald, line 268: This also assumes that all water molecules are part of a cluster?

vermouth/processors/molecule_grouper.py
```diff
@@ -0,0 +1,295 @@
...
+    Returns
+    -------
+    None
+        `system` is modified in place.
+    """
+    water_mols = [(mol_idx, mol) for (mol_idx, mol) in enumerate(system.molecules) if selector(mol)]
+    mol_idxs, water_mols = zip(*water_mols)
+    positions = []
+    for mol in water_mols:
+        # TODO: Select CoM vs CoG (current) from FF settings/metavars?
```
fgrunewald, line 238: Probably doesn't even matter for the approximate nature of water and these clusters.

vermouth/processors/molecule_grouper.py
```diff
@@ -0,0 +1,295 @@
...
+        union = Molecule(meta=mols[0].meta, force_field=mols[0].force_field,
+                         nrexcl=mols[0].nrexcl)
+        previous = []
+        for mol_idx, mol in enumerate(mols):
+            added = union.merge_molecule(mol)
+            for n_idx in added.values():
+                union.nodes[n_idx]['mol_idx'] = mol_idx
+            previous.append(added.values())
+        # Create a "cycle" of the mapped molecules, so the mapping for 3 waters
+        # doesn't match 4 waters, etc.
```
fgrunewald, line 282: Does this also influence the type the water gets? Say I have a single water or two coordinated in a pocket, would you be able to say these should be small or tiny waters?

vermouth/processors/molecule_grouper.py
```diff
@@ -0,0 +1,295 @@
...
+        The number of clusters to try.
+    **kwargs
+        Passed on to :func:`constrained_kmeans`.
+
+    Returns
+    -------
+    None
+        `system` is modified in place.
+    """
+    water_mols = [(mol_idx, mol) for (mol_idx, mol) in enumerate(system.molecules) if selector(mol)]
```
fgrunewald, line 234: As a more general note, I think in the future it would be useful to have some correspondence between a molecule and it's index in the topology file to specifically target certain molecules. So far much of the vermouth infrastructure assumes you wanna do something on a single molecule or selection based on name, res type etc. Perhaps this could become an attribute of system.

vermouth/processors/molecule_grouper.py
```diff
@@ -0,0 +1,295 @@
...
+        # TODO: PBC
+        dists = np.sum((data[:, np.newaxis] - clusters[np.newaxis, :])**2, axis=-1)
+        # When clusters are empty their clustroids become nan, which we can't
+        # round to an int.
+        dists[np.isnan(dists)] = 10**2
+        for (data_idx, clust_idx), dist in np.ndenumerate(dists):
+            # Network simplex (and capacity scaling) does not work with float
+            # weights according to nx documentation. So round the distances to
+            # an appropriate int. Magic.
+            flow_graph.edges[data_idx, clust_idx + len(data)]['weight'] = round(precision*dist)
```
fgrunewald, line 193: To what extend to your results relay on selecting a good precision? I'm asking because you only tried charmm tip3 water? If I would use SPC or so that would change anything would it? Density and coordination specifics can be different especially at non-room temperatures.

fgrunewald (commented):
I need to give the algorithm another more in depth look. Overall I couldn't see any technical problems. However, I made some conceptual remarks. Especially important seems to me:

1) In light of the fudge factors that go in this. Does this work for water models different from charmm? Variations in particle density are small but can become significant especially little farther from RT. 

2) What happens to non-clustered water molecules? If less than 3 waters are clustered does the cluster become small? Could the resolution be adjusted dynamically. Especially in light of the protein binding paper this would be important.

3)  Also, I am curious to what extend this  works for other solvents that have another cluster mapping (i.e. methanol is 2:1) for example. This is really more of an edge case, so we don't need to implement.

vermouth/processors/molecule_grouper.py
pckroon: Needed, no. But it's here now anyway, and it was little effort.

vermouth/processors/molecule_grouper.py
pckroon: Every data point will always be part of exactly 1 cluster

vermouth/processors/molecule_grouper.py
pckroon: Indirectly, yes. Since you can have different mappings for different number of water molecules, that mapping can decide on different target blocks.
Note that I actually didn't test this for the cases 1/2/3. Although 3 !<= 4, 2 <= 3.

vermouth/processors/molecule_grouper.py
pckroon: I don't quite understand what you mean here

vermouth/processors/molecule_grouper.py
pckroon: Untested, but I don't expect it to matter too much. The only reason for the precision argument to be there is that the network simplex can only deal with integers, so all floats have to be rounded to an int. The `precision` determines the precision of that rounding (although perhaps it's semantics should be inverted, so that a precision of `1e-8` means up to 8 decimals are taken into account). 
Bear in mind that the input file format also has a limited precision. Going over that will be pointless anyway.

pckroon:
> In light of the fudge factors that go in this. Does this work for water models different from charmm? Variations in particle density are small but can become significant especially little farther from RT.

Answered in the comment

> What happens to non-clustered water molecules? If less than 3 waters are clustered does the cluster become small? Could the resolution be adjusted dynamically. Especially in light of the protein binding paper this would be important.

There are no unclustered water molecules. Every data point will be assigned to exactly one cluster. The mapping afterwards determines how the clustered molecules are mapped to the target FF. What do you mean by dynamically adjusting the resolution?

> Also, I am curious to what extend this works for other solvents that have another cluster mapping (i.e. methanol is 2:1) for example. This is really more of an edge case, so we don't need to implement.
Should work. The mapping files/processor deal with mapping to the output FF> This code only deals with grouping molecules that are close together (based on CoG)

vermouth/selectors.py
pckroon: Possible. The grouping processor takes a selector as argument, which is used to select the molecules to group. I partially agree with setting the resolution+deviations/tolerances, since those will be tied to the available mappings. I think this also hinges on a new FF file format, where it should be possible to set (default) parameters for processors, which should be overwritable from the CLI

jbarnoud: > I think this also hinges on a new FF file format, where it should be possible to set (default) parameters for processors, which should be overwritable from the CLI

Just chipping it quickly: we already use force field variables to set processor values. For instance we use this to set the bond type of the elastic network and to decide if we use CoG or CoM.

pckroon: True, but the processor at the moment only need the origin system, and knows nothing about the target ff. 
How about making the martinize2 bin take the required cluster size+tolerance from the target ff, and pass that on to the processor?

jbarnoud: That would not shock me.

pckroon:
@jbarnoud What's holding back the review?

jbarnoud:
> @jbarnoud What's holding back the review?

Time available is holding back. I'll try to spend time on it later today.

pckroon:
The current implementation is utterly unfeasible :'(
When trying it on a reduced aquaporin system (7594 water molecules) it takes forever, and happily consumes 15+GB of RAM, at which point the OS kill it (*).
I need some help/advise on how to proceed here.

(*) When using pypy at least. Using CPython it takes so long it never even arrives at that point.

PR #331, opened 2021-01-20
Citation

fgrunewald:
Here's a workflow for a citation parser. We parse a .bib file (i.e. bibtex format) but lightweight. That means no checks on allowed fields only entries. We can implement such things later. Here is the main idea:

Each force-field has a citations attribute in which all citations are stored. They are read from a single file in the force-field folder. Each citation has a custom tag that can be put in block, link, modification [citation] directive. Subsequently these tags propagate to the molecule which has a set of tags. When a link is applied the molecule gets the tag. In the end the tags can be used to get the citation information from the force-field and write them out on custom format. 

The PR currently is only a proof-of-principle with no real output workflow. There is one problem that is the tags from blocks don't propagate to the final molecule. I don't know why. @pckroon  can you take a look.

@pckroon @jbarnoud 
- Would you prefer the citations to be written to the top-file, itp-file, only printed to screen or even written to a special citations file? 
- Is there a citation format we should use? 
- Someone needs to add the citations. That cannot be me, because I don't know from where which definitions came. Any volunteers?


bin/martinize2
```diff
@@ -188,6 +188,11 @@ def write_gmx_topology(system, top_path, defines=(), header=()):
             # molecules, without being uninterupted by other moltypes. Even in
             # that case, we want to write the ITP only once.
             with open('{}.itp'.format(moltype), 'w') as outfile:
+                # here we format and merge all citations
+                for citation in molecule.citations:
+                    cite_dict = molecule.force_field.citations[citation]
+                    cite_string = " ".join([field for field in cite_dict.values()])
+                    header.append(cite_string + "\n")
```
pckroon, line 195: Unless we do actual bibtex interpretation and consequent formatting the citations will be hard to read/interpret/ugly. This is one reason why I would prefer to put formatted strings as the citations rather than bibtex in the FF files

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import re
+
+class BibTexDirector():
+    """
+    Lightweight parser for BibTex files. BibTex files
+    in general have an assorment of entries that
+    describe the corresponding sort of publication
+    to refere to and then a number required and optional
```
pckroon, line 21: typo

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+    @<entry>{<some custom ID>, field = {<content>},
+                               field = {<cintent>}}
+
+    Alternatively the {} can be replaced by quotation
+    marks.
+
+    This parser only parses the version with {} as
+    used by google scholar. In addition we do not
+    check for missing fields or invalid fields. All
+    fields are accapted and no fields are required.
```
pckroon, line 35: typo

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+                              "phdthesis",
+                              "proceedings",
+                              "techreport",
+                              "unpublished"
+                             ]
+
+    @staticmethod
+    def prepare_file(lines):
+        """
+        Bibtex is not sensetive to line spacing so we join
```
pckroon, line 58: typo

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+    Lightweight parser for BibTex files. BibTex files
+    in general have an assorment of entries that
+    describe the corresponding sort of publication
+    to refere to and then a number required and optional
+    fields for the different types of entries. A field
+    for example would be Title giving the title of a
+    publication. The syntax in general looks as follows:
+
+    @<entry>{<some custom ID>, field = {<content>},
+                               field = {<cintent>}}
```
pckroon, line 27: typo

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+        the line as one string. Comment characters are not
+        allowed.
+        """
+        return " ".join(line.strip() for line in lines)
+
+    @staticmethod
+    def find_entries(citation_string):
+        """
+        Look in a string where `@` indicate the
+        begnning of a new entry and return the indices.
```
pckroon, line 68: typo

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+        citation_string: str
+
+        Yields:
+        --------
+        idx
+            position of '@' in citation_string
+        """
+        for idx, token in enumerate(citation_string):
+            if token == "@":
+                yield idx
```
pckroon, line 81:
```suggestion
        while True:
            idx = citation_string.find('@')
            if idx == -1:
                break
            yield idx
```

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+        ---------
+        str, str
+            the entry type and the shortened string
+        """
+        for entry_type in self.known_entries:
+            if entry_string.find(entry_type) == 1:
+                # remove @<entry_type>{
+                entry_string = entry_string[len(entry_type)+1:]
+                return entry_type, entry_string
+        return None
```
pckroon, line 104:
```suggestion
        assert entry_string[0] == '@'
        entry_type = entry_string[1:entry_string.find('{')]
        assert entry_type in self.known_entries
        entry_string = entry_string[len(entry_type)+1:]
        return entry_type, entry_string
```

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+        contain the @.
+
+        Parameters:
+        ------------
+            entry_string: str
+
+        Returns:
+        ---------
+        str, str
+            the entry type and the shortened string
```
pckroon, line 97:
```suggestion
        str
            The entry type        
        str
            The shortened string
```

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+        -----------
+        entry_string: str
+
+        Yields:
+        -------
+        str, str
+            the field type, the field content
+        """
+        for field, value in re.findall("(.*?)=(.*?)\}", entry_string):
+            yield field.strip(",").strip(" "), value.strip("{").strip("}")
```
pckroon, line 143: I wonder how many bibtex parsers break if you put an `'='` in your article title... 🤔

vermouth/citation_parser.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def parse_entry(self, entry_string):
+        """
+        Given a string describing a single entry, parse it and
+        then update the force_field citations dict with a field
+        dict.
+        """
+        entry_type, entry_string = self.pop_entry_type(entry_string)
+        cite_key, entry_string = self.pop_key(entry_string)
+        print(list(self.extract_fields(entry_string)))
```
pckroon, line 153: leftover

vermouth/data/force_fields/martini22/aminoacids.ff
```diff
@@ -558,7 +561,9 @@ TRP                1
 resname $protein_resnames
 [ bonds ]
 BB +BB 1 0.350 1250 {"group": "Backbone bonds"}
-
+;[ citation ]
+;M2ProteinModel
+;Martini2
```
pckroon, line 566: ?

vermouth/ffinput.py
```diff
@@ -413,6 +413,13 @@ def _link_features(self, line, lineno=0, context_type=''):
         tokens = collections.deque(_tokenize(line))
         _parse_features(tokens, context, context_type)
 
+    @SectionLineParser.section_parser('moleculetype', 'citation', context_type='block')
+    @SectionLineParser.section_parser('link', 'citation', context_type='link')
+    @SectionLineParser.section_parser('modification', 'citation', context_type='modification')
+    def _parse_citation(self, line, lineno=0, context_type=""):
+        cite_keys = line.split()
+        print(cite_keys)
```
pckroon, line 421: leftover

vermouth/molecule.py
```diff
@@ -528,6 +531,7 @@ def add_or_replace_interaction(self, type_, atoms, parameters, meta=None):
                 break
         else:  # no break
             self.add_interaction(type_, atoms, parameters, meta)
+        self.citations.update(citations)
```
pckroon, line 534: `citations` can be `None` here

vermouth/molecule.py
```diff
@@ -511,6 +512,8 @@ def add_or_replace_interaction(self, type_, atoms, parameters, meta=None):
         meta: collections.abc.Mapping
             Metadata for this interaction, such as comments to be written to
             the output.
+        citations: set
+            set of citations that apply when this link is addded to molecule
```
pckroon, line 516: s/link/interaction/

But it get worse I now realize: the interaction that's being replaced could have a citation that no longer applies. Leave it for now, but open an issue. I'm (mostly) ok with telling people to cite too many papers.

pckroon (changes requested):
Nice start :)
In addition to being able to add citations to specific blocks/links/modifications, it should also be possible to add them to forcefields in their entirety. That should save quite some typing and missing citations.

For adding citations from modifications, see https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/processors/do_mapping.py#L373. Part of that should probably be changed into a `Molecule.apply_modification` method... For later.
For adding citations from Blocks, adding them to `Molecule.copy` should do the trick. Hopefully. https://github.com/marrink-lab/vermouth-martinize/vermouth/molecule.py#L416

> Would you prefer the citations to be written to the top-file, itp-file, only printed to screen or even written to a special citations file?

At least screen (log as info) and itp. We could write our own bibtex with a later PR.

> Is there a citation format we should use?

No preference

> Someone needs to add the citations. That cannot be me, because I don't know from where which definitions came. Any volunteers?

AFAIK the atomistic files come directly from the GMX rtp files, so there should be references there somewhere. I think @jbarnoud and @paulocts wrote the martini files.

bin/martinize2
fgrunewald: this is just to check that the citation actually arrives in the itp file. Formatting is easy based on bibtex which is the entire point of doing it in the first place.

vermouth/citation_parser.py
fgrunewald: @pckroon what is the difference here?

vermouth/citation_parser.py
fgrunewald: I wonder to because our parser is fine

vermouth/data/force_fields/martini22/aminoacids.ff
fgrunewald: just for testing

fgrunewald:
> For adding citations from modifications, see https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/processors/do_mapping.py#L373. Part of that should probably be changed into a `Molecule.apply_modification` method... For later.

Still don't understand where to add stuff. Does this need to go with the links or the molecule or what? I can see what downstream becomes a molecule.

vermouth/citation_parser.py
pckroon: Efficiency. `find` is much faster than enumerating over the string

vermouth/citation_parser.py
pckroon: I missed the trailing } in the regex :)

vermouth/molecule.py
pckroon: Caching links won't solve this I think. You'd need to track citations per interaction. But I'm ok with erring on the side of caution and giving too many citations.

pckroon:
> > For adding citations from modifications, see https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/processors/do_mapping.py#L373. Part of that should probably be changed into a `Molecule.apply_modification` method... For later.
> 
> Still don't understand where to add stuff. Does this need to go with the links or the molecule or what? I can see what downstream becomes a molecule.

`graph_out` is the output molecule at the target resolution. Adding a `graph_out.citations.update(modification.citations)` right after the first log statement *should* do the trick.

> For adding citations from Blocks, adding them to Molecule.copy should do the trick. Hopefully. https://github.com/marrink-lab/vermouth-martinize/vermouth/molecule.py#L416

It may actually be better to add it to the subgraph method. Your choice.

vermouth/citation_parser.py
```diff
@@ -173,3 +173,34 @@ def parse(self, lines):
 def read_bib(lines, force_field):
     director = BibTexDirector(force_field=force_field)
     return director.parse(iter(lines))
+
+def citation_formatter(citation):
+    """
+    Very basic and minimal formatter for citations. It
+    is adopet from basic ACS style formatting. Fields within
```
pckroon, line 180: typo :)

vermouth/citation_parser.py
```diff
@@ -173,3 +173,34 @@ def parse(self, lines):
...
+    || are optionl.
+
+    <authors> |<titel>|. |<journal>| <year>; <doi>
+    """
+    # first we split the author-list
+    citation_string = ""
+    for match in re.findall("(.*?) and", citation["author"]):
+        last_name, first_names = match.split(",", 1)
+        citation_string += last_name.strip().split()[0] + ","
```
pckroon, line 189: Why the split and `[0]`?

vermouth/citation_parser.py
```diff
@@ -173,3 +173,34 @@ def parse(self, lines):
...
+        for name in first_names.strip().split(' '):
+            citation_string += " " + name.strip()[0]
+
+        citation_string += "; "
+
+    if "titel" in citation:
+        citation_string += " " + citation["titel"] +  "."
```
pckroon, line 196: titel or title?

vermouth/molecule.py
```diff
@@ -424,6 +424,7 @@ def copy(self):
         """
         new = self.subgraph(self.nodes)
         new.name = self.name
+        new.citations = self.citations
```
pckroon, line 427: Since these are both in the subgraph method I think they can both be removed here.

pckroon (changes requested):
The modification citations aren't in yet it seems. Forgot to `git add` the file?

vermouth/citation_parser.py
fgrunewald: title

vermouth/citation_parser.py
fgrunewald: leftover from something earlier I guess

vermouth/citation_parser.py
fgrunewald: doesn't work will go with what is there

fgrunewald:
@pckroon if you tell me how to print something with the logger I'll put it in as well. Otherwise I think this is pretty complete and only needs the citation blocks from Paulo/Jon.

vermouth/citation_parser.py
pckroon: Oops. 
```suggestion
        idx = 0
        while True:
            idx = citation_string.find('@', start=idx)
            if idx == -1:
                break
            yield idx
```
Feel free to leave it as is at this point though.

pckroon:
The first 2 lines should already be in /bin/martinize2 somewhere since it also logs things.
```
from ..log_helpers import StyleAdapter, get_logger
LOGGER = StyleAdapter(get_logger(__name__))

...

LOGGER.info(<whatever string you want to log>)
```
One thing to note with logging and string formatting. If you want to use string formatting its better to *not* do:
`LOGGER.info('I want to log {}'.format(42))`, but rather to `LOGGER.info('I want to log {}', 42)`. The formatting (which is expensive) will then be deferred to when the logger decides the string should actually be logged, and not filtered out.

pckroon (approved):

pckroon (dismissed):

fgrunewald:
@jbarnoud @paulocts can you perhaps provide a list of which force-field requires which paper to be cited? I can add them in the appropriate format into the program. I'm mostly unsure of all the Elnadyn stuff.

paulocts:
Hi

I will send you the papers next week.

Best,

Paulo

Em sex., 29 de jan. de 2021 às 12:45, fgrunewald <notifications@github.com>
escreveu:

jbarnoud:
I’ll have a look next week as well.

jbarnoud:
For all martini 2 and elnedyn force fields:
* https://doi.org/10.1021/ct700324x

For everything *22 or *22p:
* https://doi.org/10.1021/ct300646g

For everything elnedyn*:
* https://doi.org/10.1021/ct9002114

I thought I implemented scFix for martini 2 but apparently not, I only did to for martini3. Since the scFix paper does not describe it applied to martini3, I am not sure we should cite them there: yes it would give them credit, but it would unfairly give them the blame as well. If anybody disagrees with that, then we should cite them for each link with the `scfix` feature:
* https://doi.org/10.1021/ct9002114

fgrunewald:
@jbarnoud Do I understand it correctly that if I run martinize2 with `-scfix` using `-ff martini22` I just get plain martini22? I think this is more a data problem than citation issue. In principle I'd argue each link with -scfix needs the citation to the scfix. We should be able to hopefully only get that citation if and only if -scfix is set.

jbarnoud:
If I recall correctly, you get a warning if you ask for a feature that is not implemented in the force field. This is why I included the `[ features ]` section. I do not know what the granularity of your system is, but ideally, you want to be able to add citations for blocks and links individually; then you can cite scFix only if your system included links tagged with the scFix citation.

fgrunewald:
The current implementation should allow citations to be specific to specific links. BUT they get compounded in the apply_links processor. That means if a link is applied and overwritten the citation of the overwritten link will be kept. For scfix that should be fine because it only adds on top of things.

paulocts:
Hi Jonathan and Fabian

Jonathan, thanks for taking care of the references. For version 22p, maybe
is good to include the citation for polarizable water while all martini 2
version need to include 2007 paper from Siewert.

Regarding scfix, another issue is that our implementation is for the whole
protein sequence while the original implementation only considers a
specific secondary structure element. I also remember that you added some
new features related to glycine, which were not considered in the original
scfix implementation. Considering all this modificification, I would
consider to only cite the paper in the martinize2 paper (as the
inspiration) but not add the citation in all itps files. The same
for domainElnedyn, which Fabian also implemented in a different way from
the original paper.

Best,

Paulo

Em qua., 3 de fev. de 2021 às 11:32, fgrunewald <notifications@github.com>
escreveu:

jbarnoud:
I see. You store the citations at the molecule level. When I first thought about the feature, I was thinking about storing the citations in the nodes, edges, and interactions, then walk through them all to get the total list of citation. This would be much more work, though, and likely overkill for our needs. It was the only way I could think of for it to be completely correct. Your current approach is good enough for most cases I think.

fgrunewald:
I added the basic citations to the martini force-fields and added a check in the integration tests. For now I am not including the scfix citation by the reasoning of Paulo. When Martini3 is published and martinize2 as well we'll simply update the protein force-field files accordingly. 

@jbarnoud I've thought about the same thing basically. But I'm not sure it is worth making this effort. Currently worst case martinize asks you to cite more than you should. I think that is safe and the users need to do a little thinking of their own.

vermouth/tests/integration_tests/test_integration.py
```diff
@@ -166,6 +166,12 @@ def test_integration_protein(tmp_path, monkeypatch, tier, protein):
             result.append(token)
     command = result
 
+    # read the citations that are expected
+    citations = []
+    with open(str(data_path/'citation')) as cite_file:
+        for line in cite_file.readlines():
```
pckroon, line 172: `for line in cite_file:`

pckroon (changes requested):
Nitpick.
And could you check why the docs aren't building anymore?

vermouth/citation_parser.py
```diff
@@ -178,9 +178,9 @@ def citation_formatter(citation, title=False):
     """
     Very basic and minimal formatter for citations. It
     is adopted from basic ACS style formatting. Fields within
-    || are optionl.
+    {} are optionl.
```
pckroon, line 181: typo. And I think it's more common to denote optional fields with `[]`

pckroon (changes requested):
Tiny nitpicks left

pckroon (approved):

PR #332, opened 2021-01-26
Implement GH actions

fgrunewald:
Here's my take on GH actions for vermouth. Couple of things:

1) Appears you cannot have global environment/shell variables that are passed to the program calls. You need to feed them directly to the program called. But I don't know what calls the three variables listed atop. 

2) You cannot have allow failures in the same way as travis does it. There is skip upon error but that just tells the other tests to not stop when this particular test fails. In all other regards i.e. depending jobs will still be canceled and the test gets reported as failure. As such the PR will also become a failure. There is a open feature request for this.


.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,111 @@
...
+           WITH_SCIPY: true  
+         - py_version: "3.7-dev"
+           WITH_CODECOV: true
+           WITH_SCIPY: true   
+         - py_version: "3.8-dev"
+           WITH_CODECOV: false
+           WITH_SCIPY: false
+         - py_version: "3.9-dev"
+           WITH_CODECOV: false
+           WITH_SCIPY: false
```
pckroon, line 49: About the build matrix:
- Coverage used to be wrong (too low) for py>38, which is why we disabled coverage reporting for it. Similar for py39. That seems fixed now, so we could remove the variable.
- We need 1 case where we test without scipy. I suggest adding a py39 (not dev) without scipy, and testing 38dev and 39dev with scipy. 

Also, this matrix is actually quite large. How do you feel about removing py36-dev and py37-dev?

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,111 @@
...
+        
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
pckroon, line 81: Could be changed to 39

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,111 @@
...
+   
+  docs:
+      runs-on: ubuntu-latest
+      
+      steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+           python-version: 3.7
```
pckroon, line 101: Also 39

pckroon (changes requested):
Thanks for taking care of this, much appreciated!
Aside from your questions/comments, I have some small remarks about the build matrix (since we're touching it anyway).

> Appears you cannot have global environment/shell variables that are passed to the program calls. You need to feed them directly to the program called. But I don't know what calls the three variables listed atop.

The `SKIP_` variables are for PBR: https://docs.openstack.org/pbr/latest/user/packagers.html#authors-and-changelog
The DSSP variable is used by tests to find the correct DSSP executable: https://github.com/marrink-lab/vermouth-martinize/blob/c5f3980fdaa458a930dd89bfdc47de573e5b16fc/vermouth/tests/test_dssp.py#L38
You could probably add it to the pytest invocation.

> You cannot have allow failures in the same way as travis does it. There is skip upon error but that just tells the other tests to not stop when this particular test fails. In all other regards i.e. depending jobs will still be canceled and the test gets reported as failure. As such the PR will also become a failure. There is a open feature request for this.

Annoying but not an issue. I'm ok with such small behavioural changes when switching CI platforms.

.github/workflows/run_tests.yml
fgrunewald: fine by me also these dev versions are not really a thing anymore. Or at least they should be reasonably in line with 3.7 and 3.6. So I axed them and took care of the scipy thing.

fgrunewald:
@pckroon I think in principle it works but it can't handle something related to hypothesis?!

pckroon:
Correct, this is a change introduced with Hypothesis 6.0. I'll push a fix/workaround to this branch.
And I'll disable the expected travis check

fgrunewald:
Great, so we just have to remove travis files, and then figure out how to release to pypi. But I think this PR can be merged before?

pckroon:
Remove the Travis file, and I think it can be merged to get PRs working again

pckroon (approved):

pckroon:
LGTM. Again many thanks for taking care of this!

PR #333, opened 2021-01-28
Unflake hypothesis tests

pckroon:
Restrict hypothesis strategies generating interaction types and molecule names from using characters in the unicode classes Z (separators) and C (control characters).
This should resolve some intermittent test failures, for example the one in #328

### DIFF
--- a/vermouth/tests/molecule_strategies.py
+++ b/vermouth/tests/molecule_strategies.py
@@ -21,6 +21,13 @@ from vermouth.molecule import Interaction, Molecule, Block, Link, DeleteInteract
 
 # pylint: disable=no-value-for-parameter
 
+# This is a strategy that creates "sane" names. It's currently used to generate
+# interaction type names, and molecule names. The produced strings will not
+# contain characters from the unicode categories C and Z, which contain
+# control characters and separators, respectively.
+SANE_NAME_STRATEGY = st.text(st.characters(blacklist_categories=('C', 'Z')), min_size=1)
+
+
 @st.composite
 def attribute_dict(draw, min_size=0, max_size=None, max_depth=1):
     """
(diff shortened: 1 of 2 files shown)


vermouth/tests/molecule_strategies.py
```diff
@@ -148,7 +148,7 @@ def interaction_collection(draw, graph,
     ninteraction_types = draw(st.integers(min_value=0, max_value=2))
     for _ in range(ninteraction_types):
         ninteractions = draw(st.integers(min_value=0, max_value=2))
-        type_name = draw(st.text())
+        type_name = draw(st.text(st.characters(blacklist_categories=('C', 'Z')), min_size=1))
```
jbarnoud, line 151: Maybe this text strategy can be extracted out so it can be reused in the next test without copy-paste.

vermouth/tests/test_molecule.py
```diff
@@ -1229,7 +1229,13 @@ def test_interaction_sort(interactions, expected):
     assert vermouth.molecule.Molecule.sort_interactions(interactions) == expected
 
 
-@hypothesis.given(moltype=st.one_of(st.none(), st.text()), mol=random_molecule())
+@hypothesis.given(
+    moltype=st.one_of(
+        st.none(),
+        st.text(st.characters(blacklist_categories=('C', 'Z')), min_size=1)
```
jbarnoud, line 1235: Maybe an example case with a character from class C and Z just to know how things explode or not? I would not fight for it, though. If you are using a bell character as an atom name you are looking for problems and you deserve them.

jbarnoud (dismissed):

vermouth/tests/test_molecule.py
pckroon: I had false test failures (because of how I test `molecule.__str__`) where an interaction type is named "0 a" with 0 interactions, which is in "0 atoms", resulting in test failure. It doesn't blow up perse. 
Or a molecule name matching an empty interaction type, something along those lines. 

Easier to sanitize the input a bit than figure out and `asssume` out all possible false failure modes.

jbarnoud (approved):

PR #334, opened 2021-02-01
parse edge attributes

fgrunewald:
Here's a way to set edge attributes and propagate them into the residue graph. To not rewrite an entire set of functions dealing with the parsing of atom attributes, the edge attribute is simply taken to be the "atom" attribute of the second enumerated node. We could raise an IOError for the following case:
```
[edges]
node_A {"some_attr": some_value} node_B
```
Currently this would result in some_attr not being propagated.


vermouth/ffinput.py
```diff
@@ -935,14 +935,12 @@ def _parse_edges(tokens, context, context_type, negate):
...
+    for idx, atom in enumerate(atoms):
         prefixed_reference, attributes = _treat_atom_prefix(*atom)
-        try:
-            apply_to_all_nodes = context._apply_to_all_nodes
-        except AttributeError:
-            apply_to_all_nodes = {}
-        full_attributes = dict(collections.ChainMap(attributes, apply_to_all_nodes))
+        full_attributes = dict(collections.ChainMap(attributes, {}))
         prefixed_atoms.append([prefixed_reference, full_attributes])
+
```
pckroon, line 943: Why is this change in this PR?

vermouth/graph_utils.py
```diff
@@ -216,7 +216,14 @@ def partition_graph(graph, partitions):
     for idx, jdx in graph.edges:
         if mapping[idx] != mapping[jdx]:
             new_idx, new_jdx = mapping[idx], mapping[jdx]
-            new_graph.add_edge(new_idx, new_jdx)
+            edge_attrs = graph.edges[(idx, jdx)]
+            if new_graph.has_edge(new_idx, new_jdx):
+                old_attrs = new_graph.edges[(new_idx, new_jdx)]
+                new_attrs = {key: old_attrs[key] for key in old_attrs.keys() & edge_attrs.keys()}
```
pckroon, line 222: This check is more relaxed than the one used for node attributes: there the values also have to be the same.

vermouth/tests/test_ff_files.py
```diff
@@ -274,9 +291,13 @@ def test_interaction_edges(interaction_lines, edges):
...
         block = ff.blocks['GLY']
+        for edge, attr in zip(edges, edge_attr):
+            assert block.has_edge(edge[0], edge[1]) or block.has_edge(edge[1], edge[0])
+            assert attr == block.edges[edge]
 
-        assert (set(frozenset(edge) for edge in block.edges)
-                == set(frozenset(edge) for edge in edges))
+       # assert (set(frozenset(edge) for edge in block.edges)
+       #         == set(frozenset(edge) for edge in edges))
+
```
pckroon, line 300: leftover

pckroon (changes requested):

vermouth/ffinput.py
fgrunewald: some cleaning up; might as well do it here.

vermouth/graph_utils.py
```diff
@@ -219,7 +219,8 @@ def partition_graph(graph, partitions):
             edge_attrs = graph.edges[(idx, jdx)]
             if new_graph.has_edge(new_idx, new_jdx):
                 old_attrs = new_graph.edges[(new_idx, new_jdx)]
-                new_attrs = {key: old_attrs[key] for key in old_attrs.keys() & edge_attrs.keys()}
+                new_attrs = {key: old_attrs[key] for key in old_attrs.keys() & edge_attrs.keys()\
```
pckroon, line 222: Trailing \ should not be needed.

pckroon (changes requested):
I disagree with the cleanup in ffinput.py. The original code looks like it was put in place very deliberately, and I would not be surprised if it turns out to be used to set e.g. link order attributes, or the attributes that are to be applied to all nodes.

I'm also not sure the required behaviour is properly tested.

fgrunewald:
Sure, but on the one hand it is not covered by any unit test or integral
test. In addition as far as I could tell only Links have that attribute
which gives an empty dict. So what is the point of the try except here? But
if you prefer I'll leave it in.

Peter C Kroon <notifications@github.com> schrieb am Mi., 3. Feb. 2021,
09:22:

vermouth/ffinput.py
jbarnoud: How do you deal with the attribute to apply to all nodes, then? I do not see the feature used, though.

If you want to remove this, why keep the `ChainMap`?

fgrunewald: @jbarnoud Can you explain what the purpose of the "_apply_to_all_nodes" attribute is? I've only removed the except statement because only the Link class has a _apply_to_all_nodes attribute and that just results in an empty dict. I thought this was pointless as that is the same as just giving it an empty dict. However, I am happy to be convinced otherwise.

jbarnoud: Ok, got to dig back deeper in my memory. It may actually be used, even. Look at this example:

```
[ link ]
resname $protein_resnames
cgsecstruct "F"
[ dihedrals ]
-BB BB  +BB ++BB 1 90.7 100 1
[ features ]
collagen
```

This is equivalent to:

```
[ link ]
[ dihedrals ]
-BB {"resname": $protein_resnames, "cgsecstruct": "F"} BB {"resname": $protein_resnames, "cgsecstruct": "F"} +BB {"resname": $protein_resnames, "cgsecstruct": "F"} ++BB {"resname": $protein_resnames, "cgsecstruct": "F"} 90.7 100 1
[ features ]
collagen
```

Nobody wants to type that. Here, the attributes just after `[ link ]` go to the `_apply_to_all_nodes` attribute and is applied to all the nodes described in the subsequent sections, including `[ edges ]`.

fgrunewald: I added it back. No harm done

fgrunewald:
After some more investigation I noticed that apparently the non-edge directive selectively propagates the attributes of the right-hand side node and for some reason that is needed in the way elnedyn and martini3 work for proteins. I think this is extremely problematic but considering the imminent release I wouldn't touch this.

PR #336, opened 2021-02-08
Deploy to pypi-test

fgrunewald:
This should do the trick. I'm not at all sure about the documentation. But perhaps you can add a pypi test token and the test it to see what happens.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,7 +19,7 @@ jobs:
     
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


pckroon:
Could you also add PBR to the setup requirements in pyproject.toml?
We could also consider trimming the test matrix for this to save the cloud some needless CPU hours. Merged code should be tested in PRs anyway. Could you remove the tests againts python -dev versions?

LGTM otherwise.

fgrunewald:
Done! Once it is merged if I were you I'd test releasing a dummy package to the test pypi.

pckroon (approved):

pckroon:
Doesn't seem to work yet: https://github.com/marrink-lab/vermouth-martinize/actions/runs/548049580

Issue #311, opened 2020-09-23
Martinize2 cannot write the output files for debug


jbarnoud:
If an error occurs that causes an exception, the deferred files are never written and there is no way to cause them to be from the command line. This makes debugging more difficult.

A solution would be to have a flag in the `DeferredFileWriter` singleton to cause the files to be written right away. That flag could be set with a `-debug` or `-no-defer` option in the command line.

pckroon:
Related #298 
Yes, good argument. For now though, your files will survive in your temporary directory. 
It may also be a good idea to have the write-{canon,repair,graph} bypass the deferred writer. (Or better, have the writer write them immediately so the backup mechanism works).

PR #338, opened 2021-02-11
Identify and fix edge case in ISMAGS symmetry detection

pckroon:
Fabian found an edge case in the ISMAGS symmetry detection. This should fix it. I'll also see about pushing a fix to the networkx version at some point.
Note that I'd like this reviewed by both of you.

```
5 - 4 \     / 12 - 13
       0 - 3
9 - 8 /     \ 16 - 17
```
Assume 0 and 3 are coupled and no longer equivalent. At that point {4, 8} and {12, 16} are no longer equivalent, and neither are {5, 9} and {13, 17}. Coupling 4 and refinement results in 5 and 9 getting their own partitions, *but not 13 and 17*. Further iterations will attempt to couple 5 to {13, 17}, which cannot result in more symmetries?

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -119,6 +119,7 @@ def pdb_to_universal(system, delete_unknown=False, force_field=None,
     vermouth.MergeNucleicStrands().run_system(canonicalized)
     if write_graph is not None:
         vermouth.pdb.write_pdb(canonicalized, str(write_graph), omit_charges=True)
+        DeferredFileWriter().write()
 
     LOGGER.debug('Annotating required mutations and modifications.', type='step')
     vermouth.AnnotateMutMod(modifications, mutations).run_system(canonicalized)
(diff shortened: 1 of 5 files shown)


jbarnoud (approved):
Looking only at the change, it seems OK. I think the return is good compared to the assertion. I did not spend the weeks it took me for the initial review of the code, though.

pckroon:
Thanks for the quick review. I don't expect anything as thorough as the first round. I'm still brooding on my assumption that the case caught by the if can't result in any more/new symmetries.

fgrunewald:
Let me do this as practical review: With the fix martinize2 works until the mapping processor which just silently deletes the molecule that has raised this issue. Could be because of errors in input could be because of this fix. I'm currently using the following command line:
```
martinize2 -f renamed.gro -ff 2016H66 -ff-dir force_field/ -x CG.pdb -o topol.top -maxwarn 10000 -from 2016H66 -cter none
```
The cter workaround is needed, because it keeps identifying the terminals as C termini.

pckroon:
Curious, but I'm not completely sure it's related. ISMAGS is used only when repairing residues (since afterwards the atom names are taken into account and everything becomes asymmetric).
Using `-from 2016H66 -ff 2016H66` *should* automatically generate mappings...
Also, the mapping processor is rather loud about not knowing how to map *any* atoms in the molecule, causing it to eventually be empty.
I'll get back to you once I figure out what's happening here.

pckroon:
It's an unrelated bug: mapping takes the 'mass' attribute into account when checking node isomorphism. I'll add it to this PR, alongside a fix for #311 (by sprinkling in `write` directives).

pckroon:
This should resolve it for now. I'm also open to change the node equality function to look at atom and res name only, but then I may need to dig a bit deeper to make sure that atoms renamed by modifications are still mapped with the original block as well.

jbarnoud (dismissed):

jbarnoud (approved):

PR #339, opened 2021-02-22
Fix 317 by ensuring repair_graph prefers matching non-modification atoms

### LINKED ISSUE #317: Martinize displays an exception for some DSSP error
If the C-terminus is misrecognized, then the last residue is not sent to DSSP which leads to a mismatch between the martinize topology and the secondary structure. The error message, in that case, is an exception that is not helpful for the end-user:

```
Traceback (most recent call last):
  File "/home/jon/miniconda3/envs/vermouth/bin/martinize2", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/mnt/c/Users/dr19502/PDB/src/vermouth-martinize/bin/martinize2", line 758, in <module>
    entry()
  File "/mnt/c/Users/dr19502/PDB/src/vermouth-martinize/bin/martinize2", line 601, in entry
    AnnotateDSSP(executable=args.dssp, savedir='.').run_system(system)
  File "/mnt/c/Users/dr19502/PDB/src/vermouth-martinize/vermouth/processors/processor.py", line 37, in run_system
    mols.append(self.run_molecule(molecule))
  File "/mnt/c/Users/dr19502/PDB/src/vermouth-martinize/vermouth/dssp/dssp.py", line 458, in run_molecule
    annotate_dssp(molecule, self.executable, self.savedir)
  File "/mnt/c/Users/dr19502/PDB/src/vermouth-martinize/vermouth/dssp/dssp.py", line 291, in annotate_dssp
    annotate_residues_from_sequence(molecule, attribute, secstructs)
  File "/mnt/c/Users/dr19502/PDB/src/vermouth-martinize/vermouth/dssp/dssp.py", line 404, in annotate_residues_from_sequence
    raise ValueError(msg.format(len(sequence), len(residues)))
ValueError: The sequence length does not match the number of residues. The sequence has 372 elements for 373 residues.
```

A better solution would be to log an error telling about the mismatch and suggesting to check the termini.

pckroon:
This is combined by a cleanup of _patch_modification by relabeling all nodes to integers, and unifying the treatment of anchor and non-anchor nodes.

In ISMAGS, when mapping a subgraph node to a candidate, sort the candidates first so that the lowest is taken preferentially.

Fixes #317


jbarnoud:
Would be good to have a minimum test case that makes sure that #317 is actually fixed.

pckroon:
I had the same thought. I'll make an integrative test with a tri-alanine. It should only trigger in the case the PDB is incomplete and there's a terminal atom missing.
It's one of those bugs where the stars aligned... See also my short description in the issue.

jbarnoud (approved):
LGTM

PR #340, opened 2021-02-24
Remove resid from reference blocks in repair graph

pckroon:
Turns out that when repairing your graph you can't use blocks from ITP files as reference. Those contain resid numbers/attributes, which will overwrite the resids from your (e.g.) PDB file, causing all sorts of chaos.
This should fix it. I'll also add a test.

### DIFF
--- a/vermouth/processors/repair_graph.py
+++ b/vermouth/processors/repair_graph.py
@@ -324,10 +324,17 @@ def repair_residue(molecule, ref_residue, include_graph):
             node = molecule.nodes[res_idx]
             if include_graph:
                 node['graph'] = molecule.subgraph([res_idx])
-            node.update(reference.nodes[ref_idx])
+            # We're only really interested in correcting the atomname. Obscure
+            # usecases may utilize e.g. charge, mass, atype. Either way, we need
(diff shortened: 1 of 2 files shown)


fgrunewald (commented):
Looks ok, but just to get that straight: All other attributes in the blocks are used for repairing (i.e. mass, charge-group etc.)? I think I'd prefer a positive inclusion of node attributes (i.e. include everything that you know makes sense). Also a test for this is really needed.

pckroon:
Indeed all attributes of the reference nodes get used (except now resid). 

I'm not completely sure what's wisdom here, can we think of a scenario where you may want to run a processor to do something based on atom type or mass, before mapping to the output resolution? 

Opposite question, can we think of any other attribute that doesn't make sense? "charge_group" (not defined from PDB/GRO)? "index" ("atomid" in PDB/GRO)?

fgrunewald (approved):
Looks good, perhaps you can raise an issue that currently we also use other attributes in the residue graph making. Then we take a second look later. By the way the dendrimer still doesn't work. So there are more issues coming.

PR #344, opened 2021-02-24
Set the warning type when making bonds does not find a block

jbarnoud:
When the nake_bonds processor does not find a block, it sends the
following warning:

```
WARNING - general - Can't add bonds based on atom names for residue F-GDP501 because 'Residue GDP is not known to force field universal'. Falling back to distance criteria.
```

This commit sets the warning type to `unknown-residue` so if
`unknown-residue` is ignored, that warning is ignored.


pckroon:
Could you add 2 tests for this? One that tests the overall behaviour. and one that tests the exception raised by `_bonds_from_names`?

pckroon (dismissed):

pckroon (approved):

PR #346, opened 2021-02-25
propagate edge attributes in res-graph making

fgrunewald:
this is the left-over of #334. This PR makes sure the res-graph making propagates edge-attributes IF they are somehow given. For polyply I now subclass the ff-parser, which seems the only way to make this work.


pckroon (changes requested):
Could you still add a test where multiple attributes should be propagated?

pckroon (approved):

PR #350, opened 2021-03-05
Make cter and nter residue specifications only match proteins

pckroon:
This should make the whole martinize2 experience a bit smoother, since this makes the mutation/modification annotator only affect protein residues for `cter` and `nter` residue specifications.

Most changes are in the tests: renamed residue A to GLY, B to PHE and C to CYS so they're actually protein residues. Also added a non-protein residue to be sure.

### DIFF
--- a/vermouth/processors/annotate_mut_mod.py
+++ b/vermouth/processors/annotate_mut_mod.py
@@ -21,6 +21,7 @@ modifications.
 from .processor import Processor
 from ..log_helpers import StyleAdapter, get_logger
 from ..graph_utils import make_residue_graph
+from ..selectors import is_protein
 LOGGER = StyleAdapter(get_logger(__name__))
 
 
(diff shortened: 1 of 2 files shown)


vermouth/processors/annotate_mut_mod.py
```diff
@@ -148,6 +149,8 @@ def _terminal_matches(resname, residue_graph, res_idx):
     neighbour = list(residue_graph[res_idx])[0]  # Only one neighbour by definition.
     resid = residue_graph.nodes[res_idx].get('resid', 0)
     neighbour_resid = residue_graph.nodes[neighbour].get('resid', 0)
+    if not is_protein(residue_graph.nodes[res_idx]['graph']):
```
jbarnoud, line 152: Just to make sure I understand right: this checks that the specific *residue* is protein. Right? If so I am happy.

jbarnoud (commented):
Ideally I would like to to see the following tests:

* nonprot-prot-prot => only cter is applies
* prot-prot-nonprot => only enter is applied
* prot-nonprot-prot => both termini are applied

vermouth/processors/annotate_mut_mod.py
pckroon: Correct.

fgrunewald (approved):
This one looks all good to me. Just a question out of curiosity, this also means we cannot have any special termini modifications for nonprotein residues?

pckroon:
> This one looks all good to me. Just a question out of curiosity, this also means we cannot have any special termini modifications for nonprotein residues?

Yes and no. You can still manually patch your non-protein termini with the `-modify` flag. And we definitely need to implement something more clever at a later point indeed. I want the FF to be able to define default termini.

PR #358, opened 2021-03-29
Martini 3

pckroon:
Implement the final Martini3 files, and make martini3 the default output force field.


fgrunewald (approved):
Should be fine. But the citations somehow are added multiple times in the output. Problem for later

Issue #175, opened 2019-01-23
ff files do not accept lists as attributes


jbarnoud:
Force field variables or atom attributes cannot be lists while they would be valid python and json constructs. This is because the tokeniser does not recognize lists and split them.

pckroon:
I've been playing with lark-parser, and came up with the following LARK grammar:
```
?value: escape_char
      | dict
      | list
      | int
      | float
      | "true" -> true
      | "false" -> false
      | "null" -> null
      | string

int: SIGNED_INT
float: SIGNED_FLOAT
list: "[" [value ("," value)*] "]"
dict: "{" [pair ("," pair)*] "}"
pair: string ":" value
escape_char: "\\" /./
// Give strings priority 0 so they're matched last. Specifically, after ints.
string.0: QUOTED_STRING | STRING


COMMENT: /[#;][^\n]*/
STRING.0: (LCASE_LETTER | UCASE_LETTER | DIGIT | "_")+
_QSTR_INNER: /.*?/ /(?<!\\)(\\\\)*?/
QUOTED_STRING.0: "'" _QSTR_INNER "'" | "\"" _QSTR_INNER "\""


%import common.WS
%ignore WS
%ignore COMMENT
%import common.LCASE_LETTER
%import common.UCASE_LETTER
%import common.DIGIT
%import common.SIGNED_FLOAT
%import common.SIGNED_INT
```

with this parsing code:
```python3
from lark import Lark, Transformer

class JSONTransformer(Transformer):
    list = list
    dict = dict
    pair = tuple

    def escape_char(self, items):
        return str(items[0])

    def string(self, items):
        return items[0].strip('"\'')

    def float(self, items):
        return float(items[0])

    def int(self, items):
        return int(items[0])

    def null(self, items):
        return None

    def true(self, items):
        return True

    def false(self, items):
        return False

with open('json.lark') as json:
    grammar = json.read()

json_parser = Lark(grammar, start='value', parser='lalr')


text = '{key: [1, +2, -3, 3.14], key2: true, key3: \}, \'key 4\': some_val} ; Hello world'

tree = json_parser.parse(text)

print(tree.pretty())

out = JSONTransformer().transform(tree)
print(out)
```
[...]

PR #359, opened 2021-04-08
(re-)Implement GMX RTP parser using lark-parser

pckroon:
This PR (re) implements the RTP parser using the lark-parser parsing framework, where the file grammar is separated from the interpretation. 
If we like this style we'll use this to parse FF files as well, after we redo that file format.

This PR is a draft because it's main purpose is to foster some discussion. It also requires more extensive testing. On top of that there is one TODO left regarding the generation of 1-4 pairs between hydrogens, and docstrings are missing.

@fgrunewald what's your opinion/take on this?
The main advantages/disadvantages that I see are the following:
\+ The grammar and syntax are explicitly defined, which helps documentation
\+ It's faster than the SectionLineParser (but I haven't quantified this)
\+ When we implement the FF format as well we can start inheriting/including grammar from the other files, which makes the syntax more uniform
\+ It'll allow us to write more complex things for e.g. the JSON parts of the FF format, relaxing the need for quotes and such. See #175
\- We need to redo the parsers. Again.
\- The interpreter logic is not always as clearly depicted as with the sectionlineparser, but that may also be because I'm not good at working with abstract syntax trees.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,7 @@ install-requires =  # ?? requires-dist?
     pbr
     numpy
     networkx ~= 2.0
+    lark
 zip-safe = False
 
 [options.extras_require]
--- a/vermouth/__init__.py
+++ b/vermouth/__init__.py
@@ -32,16 +32,17 @@ LOGGER = StyleAdapter(get_logger(__name__))
 
 
 # Find the data directory once.
+from pathlib import Path
 try:
     import pkg_resources
 except ImportError:
     import os
-    DATA_PATH = os.path.join(os.path.dirname(__file__), 'data')
+    DATA_PATH = Path(os.path.dirname(__file__)) / 'data'
     del os
 else:
-    DATA_PATH = pkg_resources.resource_filename('vermouth', 'data')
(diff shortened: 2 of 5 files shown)


fgrunewald:
@pckroon 

Thanks for sharing this LARK version of the RTP parser. In general it seems more complicated to use than the SectionLineParser, but also I find it more clear in general as a language. However, I think the deciding factor will be if we can do the gromacs specific syntax with LARK as well. That we have to see because RTP ironically is the most easy to parse file format of gromacs in my opinion. 

Having said that, could you run one or two atomsitic FFs on lysozyme using PDB2GMX and compare to what you get with martinize using this LARK parser? I think it is good to get a feeling on how much is missing.

pckroon:
> because RTP ironically is the most easy to parse file format of gromacs in my opinion

I don't agree I think. The semantics of the format are super unclear due to missing documentation. It's also the format that we currently can't do with the SLP.

> Having said that, could you run one or two atomsitic FFs on lysozyme using PDB2GMX and compare to what you get with martinize using this LARK parser? I think it is good to get a feeling on how much is missing.

Good idea, I'll see if I can get around to that this week.

Issue #377, opened 2021-08-20
slow itp parser


fgrunewald:
Turns out the itp parser spents about 67% of time for large graphs dumping nodes into a list for every interaction it parses. This is super inefficient and easily fixed. PR on the way. 

https://github.com/marrink-lab/vermouth-martinize/blob/0af38fefdb655a43bf7a97f051738525af40cbff/vermouth/gmx/itp_read.py#L294

Probably the same applies to the force-field parser. 

https://github.com/marrink-lab/vermouth-martinize/blob/0af38fefdb655a43bf7a97f051738525af40cbff/vermouth/ffinput.py#L516

Issue #362, opened 2021-04-27
`itp_read.py` doens't handle `#else` lines in `itp`


ricalessandri:
One can define the following in an `itp` (used in some of the small molecule itps, for example):
```
#ifdef FLEXIBLE
[bonds]
#else
[constraints]
#endif
; i  j  funct length
  1  2   1     0.293 1000000 ; cog
....
```
and while the `#ifdef `and `#endif` are handled, the `#else` line throws the error:
```
...
...
  File "/Users/riccardo/ric.alessandri/Repos/polyply/vermouth-martinize/vermouth/gmx/itp_read.py", line 140, in parse_pragma
    raise IOError("Don't know how to parse pargma {} at"
OSError: Don't know how to parse pargma #else atline 11.
```

Issue #284, opened 2020-07-14
Reading of #ifdef fails when statments are nested


fgrunewald:
Currently the itp reader fails when #ifdef sections are nested. That is constructs of the following type raise an error:

```
#ifndef NO_RUBBER_BANDS
#ifndef RUBBER_FC
#define RUBBER_FC 500.000000
#endif
Some other stuff
#endif
```

PR #378, opened 2021-08-20
Itp reader issues

fgrunewald:
This fixes a bunch of issues related to the itp parser:

- parse '#else' directives as mentioned in #362
- increase speed of parser by fixing #377 
- parse directives position_restraints, settles, dihedral_restraints
- increase speed of make_residue graph #379 

Open issues are:
- parsing cmap #257 
- nested ifdefs fail #284


fgrunewald:
@pckroon can you give this an initial overview? Do you have any ideas how to deal with nested #ifdefs?

vermouth/gmx/itp_read.py
```diff
@@ -118,14 +120,25 @@ def parse_pragma(self, line, lineno=0):
...
+        elif line.startswith("#else"):
+            if self.current_meta is None:
+               raise IOError("Your #ifdef section is orderd incorrectly."
+                             "At line {} I read #endif but I haven not read"
+                             "a ifdef before.".format(lineno))
+            else:
+               inverse = {"ifdef": "ifndef", "ifndef": "ifdef"}
+               tag = self.current_meta["tag"]
+               condition = inverse[self.current_meta["condition"]]
+               self.current_meta = {'tag': tag, 'condition': condition.replace("#", "")}
```
pckroon, line 140: indentlint. You could also loose the `else` entirely since the if raises.

pckroon (approved):
Tiny nitpick, but LGTM otherwise!

vermouth/tests/test_itp_files.py
```diff
@@ -15,6 +15,7 @@
 Test that force field files are properly read.
 """
 
+from inspect import ismethod
```
pckroon, line 18: Unused import

pckroon (approved):

vermouth/tests/test_itp_files.py
fgrunewald: ahh shoot yes that was my original idea for the last test. I'll remove it.

fgrunewald:
@pckroon can you check the interactions in this list: https://manual.gromacs.org/documentation/current/reference-manual/topologies/topology-file-formats.html 

I think we are just missing pairs_nb and virtual_sites1. Since we're at it, I'd also add those and then be done with it.

pckroon:
Seems that way. It would/might also be nice to (re)order the interactions. Either alphabetically or in the order that's in the GMX manual.

pckroon (approved):

PR #391, opened 2021-11-17
Coords after

fgrunewald:
This is essentially an updated version of the coordinate builder but after the fact. This fixes mutations and everything where at the CG level coordinates are missing. This could also fix modifications like phosphorylation that are not actually present in the mapped pdb. Currently missing is a good routine for rebuilding VS. Luckily there is a very well tested routine in polyply. For now it only rebuilds VSN but that's good enough for proteins in Martini3 our main target. 

@pckroon 
Should we move the vs_builder from polyply to here eventually?

@Tsjerk 
As we discussed can you update the doc-strings with a little more information regarding how the coordinates are built and what quality checks are done. Also we'll need some tests here. Perhaps @pckroon can help with those as well a little. 

@paulocts 
You can use this branch for checking some examples, if you like.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -28,7 +28,7 @@ import sys
 
 import vermouth
 import vermouth.forcefield
-from vermouth.file_writer import open, DeferredFileWriter
+from vermouth.file_writer import deferred_open, DeferredFileWriter
 from vermouth import DATA_PATH
 from vermouth.dssp import dssp
 from vermouth.dssp.dssp import (
@@ -118,8 +118,8 @@ def pdb_to_universal(system, delete_unknown=False, force_field=None,
                        fudge=bonds_fudge).run_system(canonicalized)
     vermouth.MergeNucleicStrands().run_system(canonicalized)
     if write_graph is not None:
-        vermouth.pdb.write_pdb(canonicalized, str(write_graph), omit_charges=True)
-        DeferredFileWriter().write()
+        vermouth.pdb.write_pdb(canonicalized, str(write_graph), omit_charges=True,
+                               defer_writing=False)
 
     LOGGER.debug('Annotating required mutations and modifications.', type='step')
     vermouth.AnnotateMutMod(modifications, mutations).run_system(canonicalized)
@@ -127,14 +127,14 @@ def pdb_to_universal(system, delete_unknown=False, force_field=None,
     vermouth.RepairGraph(delete_unknown=delete_unknown, include_graph=False).run_system(canonicalized)
     if write_repair is not None:
         vermouth.pdb.write_pdb(canonicalized, str(write_repair),
-                               omit_charges=True, nan_missing_pos=True)
-        DeferredFileWriter().write()
+                               omit_charges=True, nan_missing_pos=True, 
+                               defer_writing=False)
     LOGGER.info('Dealing with modifications.', type='step')
     vermouth.CanonicalizeModifications().run_system(canonicalized)
     if write_canon is not None:
         vermouth.pdb.write_pdb(canonicalized, str(write_canon),
-                               omit_charges=True, nan_missing_pos=True)
-        DeferredFileWriter().write()
+                               omit_charges=True, nan_missing_pos=True,
+                               defer_writing=False)
     vermouth.AttachMass(attribute='mass').run_system(canonicalized)
     vermouth.SortMoleculeAtoms().run_system(canonicalized) # was system
     return canonicalized
@@ -191,7 +191,7 @@ def write_gmx_topology(system, top_path, defines=(), header=()):
             # A given moltype can appear more than once in the sequence of
             # molecules, without being uninterupted by other moltypes. Even in
             # that case, we want to write the ITP only once.
-            with open('{}.itp'.format(moltype), 'w') as outfile:
+            with deferred_open('{}.itp'.format(moltype), 'w') as outfile:
                 # here we format and merge all citations
                 header[-1] = header[-1]+"\n"
                 header.append("Pleas cite the following papers:")
@@ -232,7 +232,7 @@ def write_gmx_topology(system, top_path, defines=(), header=()):
     define_string = '\n'.join(
         '#define {}'.format(define) for define in defines
     )
-    with open(str(top_path), 'w') as outfile:
+    with deferred_open(str(top_path), 'w') as outfile:
         outfile.write(
             textwrap.dedent(
                 template.format(
--- a/vermouth/dssp/dssp.py
+++ b/vermouth/dssp/dssp.py
@@ -22,7 +22,7 @@ import os
 import subprocess
 import tempfile
 
-from ..file_writer import open
+from ..file_writer import deferred_open
 from ..pdb import pdb
 from ..system import System
 from ..processors.processor import Processor
@@ -143,7 +143,7 @@ def read_dssp2(lines):
     return secstructs
 
 
-def run_dssp(system, executable='dssp', savefile=None):
+def run_dssp(system, executable='dssp', savefile=None, defer_writing=True):
     """
     Run DSSP on a system and return the assigned secondary structures.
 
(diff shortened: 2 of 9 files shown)


vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+        if all(np.isnan(molecule.nodes[ndx]['position'])):
+            missing.append(ndx)
+    return missing
+
+def get_anchors(molecule, indices):
+    '''
+    Determine particles with known coordinates connected to
+    particles with given indices.
+    '''
+    anchors = []
```
pckroon, line 44: start with anchors as a set and add to that, rather first building a list and turning it to a set in the end

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+def double_cone(n=24, distance=0.125, angle=109.4, axis=None):
+    '''Create positions for two consecutive (SP3) bonded particles'''
+    P = conicring(n, distance, angle, axis)
+    S = np.array([P @ align_z(x) + x for x in P])
+    return P, S
+
+
+def triple_cone(n=24, distance=0.125, angle=109.4):
```
pckroon, line 91: I think we briefly discussed this earlier, but there's no easy way of generalizing the double/triple/quadrupal cone constructions right? Even as a recursive function?

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+    Returns
+    -------
+        None
+    '''
+    a = molecule.nodes[anchor]['position']
+    if control == None:
+        B = [
+            molecule.nodes[ix]['position']
+            for ix, v in molecule[anchor].items()
+            if v != {}
```
pckroon, line 125:
```suggestion
            if not v
```
?

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+    `down`, referring to the direction with respect to the
+    cross-product of the two substituents. If these control
+    atoms are not given explicitly in order, then the control
+    atoms are inferred from the connections of the anchor with
+    known coordinates. In that case, the specific chirality
+    of the center (R or S) cannot be controlled.
+
+    NOTE: This function should probably be able to determine
+          the weights of the substituents to allow placement
+          based on R or S chirality.
```
pckroon, line 156:
```suggestion
    TODO: This function should probably be able to determine
          the weights of the substituents to allow placement
          based on R or S chirality.
```

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+    Returns
+    -------
+        None
+    '''
+    a = molecule.nodes[anchor]['position']
+    if control is None:
+        B = [
+            molecule.nodes[ix]['position']
+            for ix, v in molecule[anchor].items()
+            if v != {}
```
pckroon, line 171:
```suggestion
            if not v
```

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+        self.bases = limbs[0][1]
+        self.limbs = [ s[2] for s in limbs ]
+
+    def __str__(self):
+        return ':'.join([str(self.anchors), str(self.bases), str(self.limbs)])
+
+    def extrude(self, distance=0.125, coords=None, contact=0.5):
+        mol = self.molecule
+
+        for i, (a, b) in enumerate(zip(self.anchors, self.bases)):
```
pckroon, line 210:
```suggestion
        for idx, (anchor, base) in enumerate(zip(self.anchors, self.bases)):
```
It costs nothing to expand these variable names, and it makes the code *so much* clearer...

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+                # Trivial chiral - except for the actual chirality
+                # a simple 'up' or 'down' flag would be helpful
+                # amino acid side chain is 'up'
+                up, down = _chiral(mol, a, target[0], target[1], control=b)
+                # Simply update this segment and return it
+                for k in self.limbs:
+                    k.discard(target[0])
+                self.bases[i] = [a]
+                self.anchors[i] = target[0]
+                return [self]
```
pckroon, line 234: Deduplicate this with the other branches

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+                    #                        self.molecule,
+                    #                        [(t, [[a]], g)],
+                    #                        self.coords,
+                    #                        self.contact
+                    #                    )
+                    #                )
+                    #  This commented-out stuff may be useful for speeding up
+                    #  coordinate generation, because it avoids needing lookup
+                    #  of missing coordinates in subsequent cycles.
+                    #return out
```
pckroon, line 276: This is missing a "but". Why isn't it used?

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+                for i, a in m.atoms
+            ]
+        )
+
+        for molecule in system.molecules:
+            self.run_molecule(molecule)
+
+    def run_molecule(self, molecule):
+        """
+        Process a single molecule. Must be implemented by subclasses.
```
pckroon, line 335: This is not the docstring I am looking for. Move along.

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+    A processor for generating coordinates for atoms without.
+    """
+    def __init__(self, contact=0.5):
+        super().__init__()
+        self.contact = contact
+        self.coordinates = None
+
+    def run_system(self, system):
+        """
+        Process `system`.
```
pckroon, line 314: This is not the docstring you are looking for.

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+        while prev - len(missing) > 0:
+            # Limbs: subgraphs of missing atoms with anchors and bases
+            limbs = []
+            for g in nx.connected_components(molecule.subgraph(missing)):
+                anchors = list(get_anchors(molecule, g))
+                bases = [ list(get_anchors(molecule, [a])) for a in anchors ]
+                limbs.append((anchors, bases, g))
+            limbs.sort()
+
+            # Segment: subgraphs of limbs sharing anchors and bases
```
pckroon, line 359: Some of these comments should be moved to the class

vermouth/processors/generate_coordinates.py
```diff
@@ -0,0 +1,382 @@
...
+            for lim in limbs:
+                if lim[0] != prev:
+                    segments.append([])
+                    prev = lim[0]
+                segments[-1].append(lim)
+            segments = [ Segment(molecule, limbs) for limbs in segments ]
+
+            # Extrude segments one layer at a time
+            while segments:
+                s = segments.pop(0)
```
pckroon, line 371: Nit: Not ideal to be popping from the left all the time. Is it possible to reverse sort the limbs above, and pop from the right here?

pckroon (changes requested):
I reviewed most of this code before, and I still don't like it very much. If this is merged as-is it'll incur a massive maintenance debt. If it breaks I'll have no idea where to look for the issue. The lack of comments and docstrings (the ones that are there are insufficient) acerbates the issue. Coupled to the absence of test this is time-bomb.
I assume I shouldn't get my hopes up regarding the missing tests, failing pylint, and docs?

I would very much welcome the VS routines from polyply here (make sure you can add the required citations, etc), but it will depend a little bit on the dependencies. Martinize2 only has a soft dependency on scipy, and making that a hard dependency needs to be a conscious decision.

fgrunewald:
will take a second look at this in a clean PR that's not based on some other branches.

PR #394, opened 2021-12-06
Deferring

Tsjerk:
This PR changes file_writer.open to file_writer.deferred_open and adds defer_writing=TRUE as boolean keyword argument to writing functions, so as to allow both direct as well as deferred writing (e.g. in write_pdb). Overwriting the function 'open' with file_writer.DeferredWriter().open was never a good idea.


pckroon (changes requested):
Looks good to me. One request though. In `bin/martinize2:pdb_to_universal` there is some logic involving the write_graph/repair/canon that can/should now be simplified.

Tsjerk:
> Looks good to me. One request though. In `bin/martinize2:pdb_to_universal` there is some logic involving the write_graph/repair/canon that can/should now be simplified.

Added changes in martinize2.

pckroon (approved):

PR #397, opened 2021-12-14
Update run_pylint.py

fgrunewald:
Fixes the broken pylint runner.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,7 +7,7 @@ extension-pkg-whitelist=
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=CVS,redistributed
+ignore=CVS,redistributed,tests
 
 # Add files or directories matching the regex patterns to the blacklist. The
 # regex matches against base names, not paths.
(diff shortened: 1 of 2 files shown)


fgrunewald:
This is a quite strange bug. One problem for sure is the usage of 'global_note' as dict key versus class attribute. For polyply that was enough to fix it. Now the exit code is also not 1 anymore but rather 28. As detailed here https://stackoverflow.com/questions/67061702/what-is-pylint-exit-code-28 this should have to do with formatting. However, after copying the pylintrc from polyply it still fails. I also cannot see an open issue that would directly apply to this. Again the fact that pylint works in polyply makes even more strange ...

pckroon:
Is the run_pylint.py still needed? IIRC they added a --fail-under to pylint proper

fgrunewald:
I guess the answer is no we don't need it, but the error we get is not related to that as far as I can tell.

pckroon:
Removing the run_pylint.py reduces the complexity, so I'd say it's a good start. Beyond that I don't know.

fgrunewald:
This PR becomes irrelevant with the update in #407.

Issue #366, opened 2021-05-06
Incorrect residue numbering when merging chains


bisejdiu:
## Issue
When merging chains that have discontinuities in their residue numbering, the output CG file will have incorrect residue numbers. 

## Reproduce
An example file can be downloaded from here: https://gpcrdb.org/structure/homology_models/cxcr1_human_active
Download the model **without loops**, since it will have several breaks/jumps in how residues are numbered. Running: 

```sh
# if we don't specify-merge R then martinize will generate 7 topology files. 
martinize2 -f ClassA_cxcr1_human_Active_6LFO_2021-04-08_GPCRdb_WL.pdb -x cg.pdb -o topol.top -ff martini3001 -scfix -cys auto -dssp -merge R
```

the output `cg.pdb` will have wrong/mismatching residue numbers. E.g., residue SER - 72 in the input pdb has been renumbered to residue SER - 139, residue GLY-106 has been renumbered to GLY - 274, and so on, with the final residue numbers being 1310 in the CG structure instead of 323. 

## Cause of Error 
I am not very familiar with the code, but it seems to me that martinize adds the start number of the new chain to the last residue number of the preceding chain. That's because:
67 + 72 = 139
67 + 101 + 106 = 274
... 
67 and 101 being the last residue numbers of the preceding "chain".  

## Solution
A good solution, at least as a quick fix, would be to have the option to keep residue numbers from the input file, or perhaps have this be the default option when `-merge` has only one argument (i.e. if we only loop once here: https://github.com/marrink-lab/vermouth-martinize/blob/master/bin/martinize2#L710) 
If we don't specify `-merge` at all, then martinize2 will generate 7 itp files (which, at least for this case, is really not ideal), even though all residues belong to the same chain.

pckroon:
Thanks for the detailed report (again!). I'm afraid this one will take a bit longer to resolve than the last one.
Merging chains is done by `vermouth.processors.merge_chains.merge_chains`. This function will call `vermouth.molecule.Molecule.merge_molecule` with the appropriate connected components (i.e. the subchains).

`Molecule.merge_molecule` calculates the residue number offset by taking the residue number of the node with the largest index. I think this explains the behaviour you see.

I would suggest the following solution in `Molecule.merge_molecule`:
1) Find the highest currently used residue number (rather than taking it from the last node)
[...]

PR #399, opened 2022-01-03
keep resids of proteins if different or chains different

fgrunewald:
This is a draft fix for issue #366 . Basically if the -protres flag is given the merge molecule will only compute a residue offset if any of the resids overlap and any of the chain identifiers overlap. In this way you just keep residue offsets in the same chain if there is no overlap and you also keep resids of different chains in the same molecule. This is the expected behavior for most PDBs and proteins.

Code wise this branch needs some improvements. But I think it makes sense to have a flag with merge molecule that dictates protein behaviour, because it is quite specific to proteins. Also I was thinking to API wise if it would make sense to maybe have generic criteria for how to merge molecules.


bin/martinize2
```diff
@@ -360,6 +360,10 @@ def entry():
                             default=[], action='append', type=lambda x: x.split(','),
                             help='Ignore residues with that name: e.g. '
                             '-ignore HOH,LIG (+)')
+    file_group.add_argument('-protres', dest='protres', default=False,
+                            action='store_true',
+                            help="Check if molecules have the different chains "
+                                 " or resids when merging and preserve them if they are")
```
pckroon, line 366: Help is missing an 'else'. I like the functionality, but I don't like the name of the flag. Can we think of something better?

vermouth/molecule.py
```diff
@@ -650,7 +650,7 @@ def __getattr__(self, name):
         else:
             raise AttributeError('Unknown attribute "{}".'.format(name))
 
-    def merge_molecule(self, molecule):
+    def merge_molecule(self, molecule, offset_handling='simple'):
```
pckroon, line 653: New argument missing in docstring, and the CLI option does store_true, but this string implies we'll support multiple strategies, which will cause a change in CLI API at a later point. I'd suggest changing the CLI option to take a string with a sane default

vermouth/molecule.py
```diff
@@ -681,16 +681,51 @@ def merge_molecule(self, molecule):
...
+        # this works well for many molecules but especially not
+        # for proteins where for example gaps in resid are common
+        if self.nodes() and offset_handling == 'simple':
             # We assume that the last id is always the largest.
             last_node_idx = max(self)
             offset = last_node_idx
             residue_offset = self.nodes[last_node_idx].get('resid', 1)
             offset_charge_group = self.nodes[last_node_idx].get('charge_group', 1)
+        # protein is more advanced and checks the intersection of chain and resid
+        # between the two merging molecules. If chains are all different, resids
```
pckroon, line 694: This is also a valid approach for DNA and arbitrary linear polymers, no? Anything for which a sane chain can be identified

vermouth/molecule.py
```diff
@@ -681,16 +681,51 @@ def merge_molecule(self, molecule):
...
+        # are preserved. If the chains are all the same but the resids are different
+        # resids are preserved. If both resid and chain are the same for at least one
+        # node we fall back to the simple scheme.
+        elif self.nodes() and offset_handling == 'protein':
+            current_chains = nx.get_node_attributes(self, 'chain')
+            new_chains = nx.get_node_attributes(molecule, 'chain')
+            same_chains = current_chains.items() & new_chains.items()
+            # all chains are the same
```
pckroon, line 702: /some/ chains are the same

vermouth/molecule.py
```diff
@@ -681,16 +681,51 @@ def merge_molecule(self, molecule):
...
+            if same_chains:
+                current_resids = nx.get_node_attributes(self, 'resid')
+                new_resids = nx.get_node_attributes(molecule, 'resid')
+                same_resids = current_resids.items() & new_resids.items()
+                if same_resids:
+                    last_node_idx = max(self)
```
pckroon, line 708: Add comment here on what this if branch will do (and why)

vermouth/molecule.py
```diff
@@ -681,16 +681,51 @@ def merge_molecule(self, molecule):
...
+                    offset = last_node_idx
+                    residue_offset = self.nodes[last_node_idx].get('resid', 1)
+                    offset_charge_group = self.nodes[last_node_idx].get('charge_group', 1)
+                else:
+                    last_node_idx = max(self)
```
pckroon, line 713: Similar comment here

vermouth/molecule.py
```diff
@@ -681,16 +681,51 @@ def merge_molecule(self, molecule):
...
+                    offset = last_node_idx
+                    residue_offset = 0
+                    # we keep this because GMX might use it and then they shouldn't be equal
+                    offset_charge_group = self.nodes[last_node_idx].get('charge_group', 1)
+            else:
```
pckroon, line 718: And also here

vermouth/molecule.py
pckroon, line 708: You need a last_node_idx for all cases, so move it up

vermouth/processors/merge_chains.py
```diff
@@ -22,7 +22,7 @@
 from ..processors.processor import Processor
 
 
-def merge_chains(system, chains):
+def merge_chains(system, chains, offset_handling='simple'):
```
pckroon, line 25: Missing in docstring

pckroon (changes requested):
Overall looks quite good, but I have a few minor concerns on design. Also, tests and docstrings are missing (but this is still a draft, so fine for now).
The bigger discussion is what attributes we want to use to identify a residue (monomeric repeat unit). Chain + resname + resid sounds reasonable, but then we miss the case where there are no chain IDs, but there is a TER record between two chains. We can consider "fixing" the chain IDs in cases like that, or add a molecule_idx as identifier.

fgrunewald:
a better PR is on the way

PR #401, opened 2022-01-13
Added documentation of file formats

Tsjerk:


doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+this section lists the atoms according to the higher resolution force
+field in the same order as specified in the corresponding
+moleculetypes. Each atom consists of a number, the atom name and a
+series of low-resolution particle names. The latter together specify
+the weighing in the mapping, both forward and backward, but are
+easiest interpreted as specifying the position of the higher
+resolution atom by the weighted average of the lower resolution
+particles as given.
+  
+Geometric modifiers
```
fgrunewald, line 334: Do we need this section? Let's leave it in perhaps better to be complete.

fgrunewald (changes requested):
Overall this documentation is pretty good. I've made some comments, where to add small bits and pieces. I don't like the one paragraph in links talking about what links should and should not be. However, we have established that and I've been overruled so nothing I can do about that. What worries me on the test side is that the build of the documentation appears to fails. Somewhere there is a reference to meta that Sphinx doesn't pick up on. This definitely needs fixing.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+[ variables ]
+elastic_network_bond_type 1
+res_min_dist 3
+
+Citations
+~~~~~~~~~
+
+The citations section lists the citations to be used for the force
+field. Citations are named and refer to an entry in the bibtex file
+`citations.bib` in the force field directory. An example of this is
```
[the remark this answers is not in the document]
Tsjerk: Okay to state it definite even if the code doesn't :) Maybe there's merit/reason to use a definite name, but if it appears that it's better not to have that, the docs can be changed still.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+
+[ citations ]
+Martini3
+
+Citations can also be specified for moleculetypes, links, and
+modifications, in a citation subsection.
```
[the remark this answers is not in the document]
Tsjerk: Okay, adding '(note the singular)' to the last line there.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+The moleculetype describes a :ref:`block <data:block>`, i.e., a
+topological building block, comprising of particles (atoms) with their
+properties and interactions. This can be a separate molecule or a part
+of a larger molecule, typically a monomeric unit in a polymer. The
+moleculetype has a name and the number of bonds to use for exclusions
+as its first content line. This is followed by one or more
+subsections. The subsection atoms is mandatory, optional subsections
+are edges and subsections corresponding to the different types of
+atomic interactions, including bonds, constraints, angles, and
+dihedrals.
```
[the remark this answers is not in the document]
Tsjerk: Adding a note about the unique atom names. The link to the Gromacs manual can be added later (have to sort out the best way (document) to link).

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+* distance_restraints(2)
+* dihedral_restraints(4)
+* orientation_restraints(2)
+* angle_restraints(4)
+* angle_restraints_z(2)
+* cmap(...)
+
+Any of the subsections can be given multiply times, in which case they
+are additive. Do note that in the output topology specifying the same
+interaction several times (the same type and particles) will overrule
```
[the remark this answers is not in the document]
Tsjerk: I recall I checked the statement and found they were additive. Checking the code suggests I'm wrong. I don't like using 'version', because it suggests the interactions have a temporal hierarchy (version 3 is newer than version 2), while the interaction is actually a compound interaction (a dihedral profile consisting of several simple dihedrals). In that context, using the metas is overly convoluted.  I can see how it works now, but I think it would be better to have ordered collections of interactions; then a link/mod that needs to change an interaction will just replace the whole collection, rather than needing to keep track of the numbers given to the individual interactions. It'll probably be wise to issue a warning if  interactions are added on a set of atoms in non-consecutive lines. 

I'll file an issue on this one. Then we can discuss the best approach there. AFAIC the docs describe how this should work (just like it works in Gromacs, btw), but the behaviour is slightly different, so that's a bug.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+
+#meta {"group": "Side chain bonds", "ifdef": "FLEXIBLE"}
+
+Metadata can also be added to a single line by adding an attribute
+statement as the last element.
+
+Atoms
+^^^^^
+
+Each line in the atoms section describes one particle, corresponding
```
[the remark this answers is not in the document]
Tsjerk: That's possible with the current format, no? The only restriction is that names have to be unique. It would be doable to say the combination of atom name and residue id+name needs to be unique, but we can't deal with all equal atom names.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+molecular graph. These metadata can be used (and modified) by
+Vermouth's processors and for provenance.
+
+Edges and non-edges
+^^^^^^^^^^^^^^^^^^^
+
+Within the context of links, the edges and non-edges specify that an
+edge should or shouldn't be present, respectively, for the link
+pattern to match. This behavior will likely be deprecated in the near
+future in favor of filter keywords 'edge' and 'nonedge'. The current
```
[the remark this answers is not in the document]
Tsjerk: Oops. There were discussion points included in the section accidentally. Everything from the sentence starting with 'The current' should be dropped. The four lines describe the state as is and indicate that this part is likely going to be modified slightly (transfer of lines from sections edge/non-edge to the regular filtering section). 

A propos, users are not writing file formats, only writing files (or attempting) using the formats (as they perceive/understand them)

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
-modifications, and block mappings that cross residue boundaries.
+
+Vermouth has two main types of data files:
+
+* Force fields describe :ref:`blocks <data:block>`, :ref:`links <data:link>` and :ref:`modifications <data:modification>` to generate
+topologies.
+
+* :ref:`Mappings <data:mapping>` describe the transformations
+  required for going from one description (force field) to another or
+  vice versa.
```
pckroon, line 11:
```suggestion
* Force fields describe :ref:`blocks <data:block>`, :ref:`links <data:link>`
  and :ref:`modifications <data:modification>` to generate topologies.

* :ref:`Mappings <data:mapping>` describe the transformations
  required for going from one description (force field) to another or
  vice versa.
```

doc/source/file_formats.rst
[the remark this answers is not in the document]
pckroon: It's been a while since I've made datafiles but I'm sure there's some gotchas when you try to add a block from an itp file. For sure there is the requirement that atomnames must be unique (which probably produces an obscure error message). I'm not sure there's more. 
It's not really an issue if you add small molecules/residues (the main usecase). It blows up in your face for polymers of course.
Is it desirable to have the parser generate unique atomnames by adding numbers? I kind of doubt it, since that complicates writing mappings (and links and modifications) since you can no longer lookup atom names in the itp file. An intelligible warning/error would be my preference I think.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+`citations.bib` in the force field directory. An example of this is
+
+[ citations ]
+Martini3
+
+Citations can also be specified for moleculetypes, links, and
+modifications, in a citation subsection.
+
+Moleculetypes
+~~~~~~~~~~~~~
```
pckroon, line 82: Each subsection should have a (short) example I think.

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+* virtual_sites3(4)
+* virtual_sites4(5)
+* position_restraints(1)
+* distance_restraints(2)
+* dihedral_restraints(4)
+* orientation_restraints(2)
+* angle_restraints(4)
+* angle_restraints_z(2)
+* cmap(...)
+
```
pckroon, line 169: virtual_sitesn needs special mention, since it takes an unknown number of atoms, and the syntax is different (is it still?) from the gromacs syntax:
gmx: `<at_id1> <func> <at_ids ...>`
M2: `<at_id1> <at_ids...> -- <func>`

Why is this? So that we can assume (only) the first X elements are atom names. Everything after `--` is parameters. This was added to be able to kind-of support unknown interactions, but that functionality perished later on. This `--` also works for other interactions. If you insist on making a bond with 3 atoms:
`at1 at2 at3 -- 2 0.1 400`

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+The lines following the section tag may list selection statements for
+filtering atoms in which to search for matching patterns. Each line
+specifies a property and the corresponding value. The selection
+statements may include filters based on, e.g., the residue name and
+the secondary structure type, which are used to determine the
+structural properties of protein backbone in the Martini force field.
+
+Note: the link should have a name as overall content, to be consistent
+with moleculetype and modification (and to increase clarity), and the
+filters should probably fall under a subsection [ filters ].
```
pckroon, line 197: Agreed. But should this note be here? I guess you can add comments in rst... Let me look.

pckroon:
```suggestion
..
  Note: the link should have a name as overall content, to be consistent
  with moleculetype and modification (and to increase clarity), and the
  filters should probably fall under a subsection [ filters ].
```

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+link applies. Each line in the subsection describes a pattern. One of
+the patterns must apply for the link to match. A pattern consists of
+atom identifiers. Each atom identifier consists of a name which may be
+preceded by a shift and which may be further specified using an
+attributes statement. The shift indicates a relative position in terms
+of residues:
+
+* +, ++, +++ : 	first, second, third following residue
+* -, --, --- : 	first, second, third previous residue
+* > :		other residue (not in order)
```
pckroon, line 212: `>` means residue with larger resid (smaller also exists).
`*` means residue with different resid (smaller or larger). Beware of symmetry hell. But there's an open PR about this #402

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+
+Thus, +CA in amino acids refers to the C-alpha atom in the C-terminal
+connected neighbor, while >SG in the construction of a disulphide
+bridge will refer to the SG atom in the partner cysteine.
+
+An attributes statement is a JSON style mapping of key/value pairs,
+similar to those described above (see #meta). For a pattern to match,
+the atom names and the attributes need to match.
```
pckroon, line 220: We need to list attributes and their semantics somewhere

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+
+To generate a topology for a polymer or any molecule constructed from
+joining parts, Vermouth connects moleculetypes using links. A link
+describes how blocks are to be joined, what changes are effected in
+the atom lists and which interactions are added, removed, or
+altered. The changes in the atom and interaction lists are specified
+using the corresponding subsections as under moleculetype. Further
+subsections available for links are patterns, features, molmeta,
+edges, and non-edges
+
```
pckroon, line 187: From here on I would restructure the Links section somewhat:
- Links consist of atoms and edges. These are constructed from [atoms] [edges] and some interactions. 
- Links apply where the match the molecule being constructed.
- Atoms know something about (relative) resids (explain order attr here)
- Further filtering using patterns, features, molmeta, and non-edges.
- Atoms, edges and interactions; links can remove and modify atoms and interactions
- patterns, features, molmeta and non-edges

To me this flow makes more sense, because at the moment the order attribute falls from a blue sky in the Patterns subsection

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+connected neighbor, while >SG in the construction of a disulphide
+bridge will refer to the SG atom in the partner cysteine.
+
+An attributes statement is a JSON style mapping of key/value pairs,
+similar to those described above (see #meta). For a pattern to match,
+the atom names and the attributes need to match.
+
+Shift operators can also be used in the interaction subsections.
+
+Features
```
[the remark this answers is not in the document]
pckroon: We also need to list the available features (and semantics)

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+Intuitively, edges and non-edges should specify the state after
+application of the link, similar to the other interactions. I would
+think that selection properties 'edge' and 'nonedge' in the filtering
+would be more intuitive and useful.
+
+Modifications
+^^^^^^^^^^^^^
+
+Modifications can be used to edit molecules or parts thereof (blocks),
+e.g., for specifying protonation states. Each modification starts with
```
pckroon, line 258: blocks only. And I would add termini as example

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+a line with the name. Thereafter may follow subsections as under
+links. A modification may add, remove, or change atoms, interactions
+and/or edges, using the corresponding subsections. Subsections
+patterns, features, molmeta, edges, and non-edges function as under
+links.
```
pckroon, line 263: this should probably be true, but currently isn't. I know for a fact that modifications currently can't remove interactions (and they definitely need to be able to!)

doc/source/file_formats.rst
```diff
@@ -1,23 +1,375 @@
...
+corresponding to the modification. Such changes can also be applied to
+atoms already present in the molecular graph, i.e., the 'non-PTM
+atoms'.`
+
+Interactions
+^^^^^^^^^^^^
+
+Interactions are added or changed under the same interaction
+subsections as used for moleculetype. How are interactions removed?
+
```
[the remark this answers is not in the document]
pckroon: Or maybe just refer to e.g. termini modifications? Those are not feature-complete, but I do think they cover most use cases.

doc/source/file_formats.rst
pckroon: I'd suggest to remove it, or add a big notice that M2 doesn't use them.

pckroon (changes requested):
Nice work! 
I agree with most of the text and Fabian's comments. I do think the Links section could do with a little bit of restructuring (see comment). 
I think it's mostly complete, but I'm missing something on the `.mapping` format. I'll see if I can write that soon-ish.
I'll also leave this here: https://vermouth-martinize.readthedocs.io/en/doc_format/

PR #402, opened 2022-01-14
Cysteine

Tsjerk:
This PR fixes the issue where a cysteine bridge is to be added between chains, where the cysteines have the same residue id. It was tested on 3i40.pdb. Maybe a good idea to add that as test case.


fgrunewald:
I've had a closer look at the PR. At the moment 3 integration tests fail. Reasons I found so far:

- the scfix does not work anymore; I think more general it seems to mess with the features attribute; not sure how and why though. The problem seems to be with the match_order not the other bit
- the first BBS angle seems to be always missing; I think this bit is related to filtering and the versions attribute
-  there is already syntax option that uses the * character. As far as I can it is only present in martini22p though

@pckroon do you have any idea how this messes with features? The problem here seems that cysteine bridges are not added when the resids and all other attributes are the same except for the chain identifiers. The reason is that match_order makes sure the resids cannot be equal. 

Here are some more general thoughts:

- I like the idea of having a '*' syntax attribute that means can have all equivalent attributes just needs to be a different node. But I think we should not put this into the match-order function but evaluate this in the DoLinks based on weather the nodes are the same or not using the networkx property of a node. Also we need to resolve the conflict with the Martini22p '*'. I don't even know what that does. @jbarnoud can you remember this bit?
- We should not use a filtering of interactions in DoLinks, in general add_or_replace_interaction takes care of this even though it's super slow. We could also have a temporary interactions dict like in polyply where keys are simply frozensets. That works well and gives quite some speed-up. 
- Whatever we do we should try to make it backward compatible as much as possible.

pckroon:
The main issue I see is that the whole "order" attribute is a bit of a hack. When we implemented it we only ever considered the resid to the the relevant attribute, but this is (obviously, in hindsight) not true. This PR is also a patch on top of the already festering implementation (but it works! For now).
With this PR, for order = "\*", now both resid and chain id will be taken into account. However, I can think of cases where that still breaks (2 chains without chain id, even though we know they're different molecules). Is that a bad thing? Maybe. Is it worth fixing now? Probably not.
See also #57.

pckroon:
> We should not use a filtering of interactions in DoLinks, in general add_or_replace_interaction takes care of this even though it's super slow. We could also have a temporary interactions dict like in polyply where keys are simply frozensets. That works well and gives quite some speed-up.

This is a good idea. It makes much more sense for add_or_replace_interaction to know which interactions are symmetric (or not). The only issue/problem is that links can do more than add interactions (remove interactions and nodes, for example), so we can't remove /all/ the symmetry filtering from DoLinks, but a middle ground is possible. For example, if a link only adds interactions don't filter anything, and otherwise decide on whether it's symmetric or not based on more complex criteria.
Does require some careful thought. What if Links add multiple interactions (which don't need to overlap)?

vermouth/processors/do_links.py
```diff
@@ -98,7 +98,7 @@ def _interpret_order(order):
     return order_type, order_value
 
 
-def match_order(order1, resid1, order2, resid2):
+def match_order(order1, resid1, chain1, order2, resid2, chain2):
```
pckroon, line 101: Currently you separate chain1, resid1, chain2, resid2 everywhere. Would it make sense to consolidate them into a `residue_identifier`? Could be a tuple (or dict). Does bring a few headaches for dealing with orders +-<>, since that implies you need to be able to subtract and compare them. Or, for those cases, you pull out the resid.
Doesn't make much difference now, but does make adding more criteria later on easier.

vermouth/processors/do_links.py
```diff
@@ -216,10 +220,24 @@ def match_link(molecule, link):
...
     for raw_match in raw_matches:
         # raw_match: mol -> link
         # rev_raw_match: link -> mol
         rev_raw_match = {value: key for key, value in raw_match.items()}
+        
+        # When the 'other' operator (*) is used, links involving
+        # same atoms from different residues will be doubled over
+        # permutations. This needs to be avoided.
+        unique = {a.lstrip('*') if isinstance(a, str) else a for a in raw_match.values()}
+        if len(unique) == 1:
```
pckroon, line 233: This will only work for, e.g. a Link `SG *SG`, and not for `CB SG *SG *CB`. If you want to implement the latter option, you should also take connectivity into account I think though, so that'll become messy.
Actually, what we need/want is to be able to decide whether or not a link is symmetric (automorphism!), but then we're back at the discussion at #57.

vermouth/processors/do_links.py
```diff
@@ -216,10 +220,24 @@ def match_link(molecule, link):
...
+            group = (unique.pop(), set(raw_match.keys()))
+            resnames = {molecule.nodes[idx]['resname'] for idx in group[1]}
```
pckroon, line 235: Why only resname and not chain? Or other attributes?

vermouth/processors/do_links.py
```diff
@@ -232,16 +250,18 @@ def match_link(molecule, link):
...
                 # Assert all orders correspond to the same resid
                 elif order in order_match and order_match[order] != resid:
                     break
         else:  # No break
-            for ((order1, resid1), (order2, resid2)) in combinations(order_match.items(), 2):
+            for ((order1, (resid1, chainid1)), (order2, (resid2, chainid2))) in combinations(order_match.items(), 2):
                 # Assert the differences between resids correspond to what
                 # the orders require.
-                if not match_order(order1, resid1, order2, resid2):
+                print(order1, resid1, chainid1, order2, resid2, chainid2)
```
pckroon, line 263: Some leftover

vermouth/tests/test_links.py
```diff
@@ -23,156 +23,157 @@
...
+        (('**', '>'), (5, 4), ('A', 'A'), True),
+        (('****', '>>>'), (5, 5), ('A', 'A'), True),
         # *, *
-        (('*', '*'), (6, 6), True),
-        (('*', '*'), (6, 5), False),
-        (('**', '**'), (6, 6), True),
-        (('**', '**'), (6, 5), False),
-        (('**', '****'), (5, 7), True),
-        (('**', '****'), (5, 5), False),
+        (('*', '*'), (6, 6), ('A', 'A'), True),
```
pckroon, line 164: Could you add a `(('*', '*'), (6, 6), ('A', 'B'), True),`? Or `((0, '*'), ...)`? One of those should match but didn't before this PR.

pckroon (changes requested):

vermouth/processors/do_links.py
Tsjerk: Yes, this is a hack, and it would be better to have a residue id class that supports comparisons; it wouldn't need a chain ID then, because you can query the corresponding molecule graphs if so needed. That requires a few more lines though, and thus more time than we have now.

vermouth/processors/do_links.py
Tsjerk: This would also be solved with a resid class, making * mean 'different molecule'

vermouth/processors/do_links.py
pckroon: A `Residue` class would indeed fix a few things. But I completely it's out of scope for this PR. Could you open an issue?

vermouth/processors/do_links.py
Tsjerk: It can only happen if the chains are _different_ (using *). However, it doesn't need only apply to same residue names. In practice it does now (CYS-CYS), and it _is_ a symmetry issue (the same pattern applies both ways). But there could be cases where symmetry in atom names exists withouth symmetry in residue names, so this part of the check should probably be removed.

vermouth/processors/do_links.py
pckroon, line 101: A different thing, that should be fixed: this changes the public vermouth API, and that breaks polyply. Could you make the chain arguments optional?

Tsjerk: That requires changing the order to completely illogical. But it's doable.

pckroon: Agreed, but fugly API is preferable to broken API I'd say.

pckroon (changes requested):
I'm still missing a testcase that would fail on the current code, but will pass with this PR. LGTM otherwise!

Tsjerk:
> I'm still missing a testcase that would fail on the current code, but will pass with this PR. LGTM otherwise!

(('*', '*'), (6, 6), ('A', 'B'), True) was added as per suggestion in commit 295596a419760e59f5085e98be93e3d72647e025.

pckroon (approved):
Oops, my bad! I'm a little bit surprised there's still so many uncovered lines... But LGTM :)
Merging is blocked until the docs build, but I suggest we either fix that in the respective PR, /or/ rebase this on master.

fgrunewald:
@pckroon this needs mostly tests and otherwise we're happy with it?

pckroon:
As far as I'm concerned tests are a "nice to have" in this case (but shouldn't be too hard. I think it's mostly a few edge cases). 
I'm blocking this on the docs not building though, and for some reason this PR is based on #401 rather than master. So IMHO this needs to be rebased.

fgrunewald:
This PR will be replaced by #444.

Issue #360, opened 2021-04-20
Make martinize2 accessible in python


OLaprevote:
Couldn't martinize2 exist as a python function too? This way it would be possible to call it directly in python

Ex:
```python
from vermouth import martinize2

input_file = 'my/path/to/inp.pdb'
output_file = 'my/path/to/out.pdb'
topology_file = 'my/path/to/out.top'

martinize2(input_file, output_file, top_path=topology_file, cystein_bridge='auto',
                  neutral_termini=True, to_ff='martini3001', posres='backbone')
```

Everything is already written in `bin/martinize2` and it feels like a loss not to give the choice. I was writing a jupyter notebook and I thought that having it as a python function would be more practical than calling the bash command. Not too hard to implement either: I would gladly spend the time to do these changes.

pckroon:
I strongly agree, and I think it would be a great idea. I would be very open to a PR, if you have the time. I'm currently rather swamped unfortunately, but review shouldn't be too hard.

PR #405, opened 2022-02-25
Make martinize2 accessible in python (see Issue #360)

OLaprevote:


vermouth/martinize2.py
```diff
@@ -0,0 +1,684 @@
...
+        Factor with which Van der Waals radii should be scaled when
+        determining bonds based on distances.
+    
+    Force field parameters
+    ----------------------
+    to_ff: str, default=martini3001
+        Which forcefield to use.
+        To know available forcefields:
+        >>> forcefield.find_force_fields(
+        ...     Path(vermouth.DATA_PATH) / 'force_fields')
```
pckroon, line 302: I wouldn't linewrap this. If it's a little bit too long it's not a problem

vermouth/martinize2.py
```diff
@@ -0,0 +1,684 @@
...
+    leftover_warnings = ignore_warnings_and_count(COUNTER, maxwarn)
+    if leftover_warnings:
+        LOGGER.error('{} warnings were encountered after accounting for the '
+                     '-maxwarn flag. No output files will be '
+                     'written. Consider fixing the warnings, or if you are sure'
+                     ' they are harmless, use the -maxwarn flag.', leftover_warnings)
+        sys.exit(2)
+    else:
+        DeferredFileWriter().write()
+        processors.Quoter().run_system(system)
```
pckroon, line 684: Could you add a newline at the end?

pckroon (changes requested):
This is great! Many many thanks :D

While you were working on this I guess the API of the deferred file writer changed. It no longer exposes an `open`, but rather `deferred_open`. This is also what causes the test failures. 
I like how you formatted the docstring for the `martinize2` function, but I'm not sure what sphinx will think of it. I guess we'll find out. I can deal with it if sphinx doesn't like it.
Beside that a final request: currently there is a lot of duplication in the docstring (and signature) of martinize2 and the CLI. Could you open an issue to generate the signature and docstring from an argparse object? It's out of scope for this PR though.

Again, looks great! And thanks again

OLaprevote:
Changes done :)
Yes, I thought so for Sphinx. I'm curious too, as you mentioned in a worst case scenario it's just removing a few lines.
Yup, the duplicated docstring is a pain. Looking into the problem right now: I have some ideas on how to proceed.

pckroon:
Thanks :)
There's something going wrong with passing filenames, and there's also something fishy with the docs that I can't reproduce locally... I'll see what's going on there.

> Yup, the duplicated docstring is a pain. Looking into the problem right now: I have some ideas on how to proceed.

I was thinking of the inspect module. That has the functionality to create/change signatures and docstrings on the fly. And the argparse object should have all the required info to fill the blanks. But like I said, it's out of scope for this PR. Could be a nice second one though ;)

> One thing: I left the `__init__.py` untouched. That means currently if anyone wants to use the python function they have to write from `vermouth.martinize2 import martinize2` (or `import vermouth.martinize2`). Would you rather have it more direct?

I think this is fine for now. We can always add it to the `__init__.py` at a later point.

pckroon:
Ok, small issue. The docstring style you use requires sphinx > 3.1.2. If I update my sphinx to the latest (4.4.0) I get about a million warnings. I'll open a separate PR to fix those, and then I think I'll come back here...

pckroon:
I'm going to try closing and reopening this PR to get the CI to run. Hopefully.

OLaprevote:
I've found a few mistakes in my docstrings as well while getting the automated documentation to work, e.g. missing colons. That may be one of the reason it doesn't pass the test. The corrections on the docstring and the automated documentation are right now one and only commit on my local computer: as soon as I have a bit of time I split them and push only the docstring changes.

pckroon:
Ok great :)
I'll patiently wait for those commits then!

PR #407, opened 2022-03-02
Upgrade sphinx to latest, and fix doc issues

pckroon:
Also add some napoleon config. Setting preprocess_types to True
might (have been) a good idea, but will currently cause about 200
warnings, since that will break with e.g. `list[str]`. Instead we'd
need to change it to e.g. `list of str`, and IMHO it results in
less pretty html


doc/source/conf.py
```diff
@@ -185,6 +185,13 @@
                            'undoc-members': None,
                            'show-inheritance': None}
 
+napoleon_google_docstring = False
+napoleon_google_docstring = True
```
fgrunewald, line 189: Aren't these two variables exactly the same?

requirements-docs.txt
```diff
@@ -1,8 +1,8 @@
-sphinx == 3.1.2
+sphinx
 sphinxcontrib-apidoc
 pbr
 setuptools >= 30.3.0
 
 numpy
-networkx ~= 2.0
+networkx
```
fgrunewald, line 7: I think defining a minimum networkx version would be better, because there are still quite few people that run older versions. But this is an optional comment

fgrunewald (changes requested):
Looks good! One doubled line in the config file and one optional comment. And of course ideally you'd figure out why the lint fails. As far as I can see the package scores a 9.07 but the fail limit is 8 so should be fine.

pckroon:
Pylint failed on bin/martinize2, which was set to 9.5. The score dropped because it started taking into account places where you should/could use f-strings rather than .format. So I dropped the requirement to a 9.0.

fgrunewald:
yeah that makes sense. Personally I think the fstring should be treated equivalent to the format option in terms of linting. Because for long strings frstring becomes less easy to comprehend.

requirements-docs.txt
fgrunewald: @pckroon how about this one?

pckroon: I think latest is fine tbh, so let's leave it like this

fgrunewald (approved):

Issue #421, opened 2022-05-10
Martinizing multiple proteins fails 


BartBruininks:
When I am running martinize on a homo dimer (e.g. two insulin molecules), the residue index skips the exact model size before numbering continues. This additional shift of one times the model size seems to also be preserved in the ITP. I have a strong feeling that there is a small math error somewhere. Renumbering the residues or atoms in the input file does not solve the issue. Adding a single CONECT between the two molecules does rescue this behavior, but can result in an additional unwanted bond. Plus it is a rather dirty solution.

`martinize2 -f 5BTS_clean_gmxRenumber.pdb -o insulin_dimer.top -x insulin_dimer_cg.pdb -dssp /usr/local/bin/mkdssp -p backbone -ff martini3001 -govs-include -govs-moltype insulin_dimer -scfix -cys auto`

[5BTS_clean_gmxRenumber.pdb.txt](https://github.com/marrink-lab/vermouth-martinize/files/8660701/5BTS_clean_gmxRenumber.pdb.txt)
[insulin_dimer_cg.pdb.txt](https://github.com/marrink-lab/vermouth-martinize/files/8660702/insulin_dimer_cg.pdb.txt)

fgrunewald:
Hi Bart,

Does this only concern the resid numbers in the itp file?
Do you get the dislufide bonds as required?

BartBruininks:
It seems to be present everywhere. The CG PDB has the jump, the ITP has the jump. I checked for 6 disulfide bridges, they were there, however, I do not know if they connect the right residues. I think in theory this file could work but the numbering just is very strange and clearly something is going on. 

Besides, the virtual sites scripts do not like this at all as now the numbering is not consistent with the contact map files.

fgrunewald:
Just to be completely sure: By numbering you mean resids? Could this be a duplicate of #154 and/or #366

fgrunewald:
@BartBruininks can you try this branch on your protein https://github.com/marrink-lab/vermouth-martinize/pull/399

BartBruininks:
I can, but not now. I will do it hopefully later today or otherwise tomorrow.

fgrunewald:
Thanks! Let me know if that works.

BartBruininks:
I used fix/366 and it did not solve the issue of the residue numbering skip.

fgrunewald:
[...]

Issue #154, opened 2018-10-09
resids are not kept


jbarnoud:
The resids are not kept from the input structures into the result file. This is a problem as resids are used by the users to relate their output structure with the input one, and with the literature.

One fix would be to add the resid in the `attribute_keep` argument of `do_mapping` and its corresponding processor. Though, it I recall correctly, we lost the resid because it could break mappings spanning across residues.

NanduTej:
Hey, Can I pick this up?

pckroon:
It may be better if @jbarnoud and me first have a short discussion on what we want this to do in the first place. His comment
> Though, it I recall correctly, we lost the resid because it could break mappings spanning across residues.

does indeed create issues here that require some thought. How good is your polymer chemistry/bioinformatics? ;)

NanduTej:
> It may be better if @jbarnoud and me first have a short discussion on what we want this to do in the first place. His comment
> 
> > Though, it I recall correctly, we lost the resid because it could break mappings spanning across residues.
> 
> does indeed create issues here that require some thought. How good is your polymer chemistry/bioinformatics? ;)

Pretty bad actually. I haven't done any chemistry in 4/5 years. :sweat_smile:

pckroon:
Hey,

after a discussion with @jbarnoud we decided that the mapping processor needs a rather sizable overhaul. It would be a waste of your time and effort if you fix this now since it's going to change anyway. I'll fix this while I deal with #165.

pckroon:
Also, as a note to myself: there are three cases, to be picked by the user:
1) Renumber everything
2) Keep everything the same as much as possible. If we write a file format that does not support insertion codes, and we have ambiguous residues (same name and number) at the end, give a warning.
3) Renumber with an offset based on the first resid.

PR #422, opened 2022-05-10
include region

xikunliu:
I included a new flag -region which can apply rubber bands only to the atoms within the designated regions. One example can be that I know resid 200-300 and 450-500  in my protein is structured, but the rest residues are not. I can use the following command to apply rubber bands only within resid 200-300 and 450-500 among different chains:
martinize2 -f example.pdb -ff martini3001 -dssp -elastic -merge A,B,C,D -scfix -region 200:300,450:500


bin/martinize2
```diff
@@ -726,6 +728,17 @@ def entry():
             message = 'Unknown value for -eunit: "{}".'.format(args.rb_unit)
             LOGGER.critical(message)
             raise ValueError(message)
+        if args.rb_region == 'molecule':
+            domain_criterion = vermouth.processors.apply_rubber_band.always_true
+            regions = []
+        else:
+            regions = [tuple([int(i) for i in apair.split(':')]) for apair in args.rb_region.split(',')]
+            if len([i for i in [len(atuple) for atuple in regions] if i != 2])==0:
```
pckroon, line 736: nit: these lines are rather complex, and might benefit from being split into smaller bits and/or comments

pckroon, line 736: lint nit: spaces around ==

pckroon, line 736:
```suggestion
            regions = [tuple([int(i) for i in apair.split(':')]) for apair in args.rb_region.split(',')]
            if any(len(region) != 2 for region in regions):
```
I think this would have my prefence as far a clarity goes, but I'm pretty nit-picky at the moment.

vermouth/processors/apply_rubber_band.py
```diff
@@ -365,7 +365,43 @@ def same_chain(graph, left, right):
     node_right = graph.nodes[right]
     return node_left.get('chain') == node_right.get('chain')
 
+def same_region(graph, left, right, regions):
+    """
+    Returns ``True`` is the nodes are part of the same chain.
+
+    Nodes are considered part of the same chain if they both have the same value
+    under the "chain" attribute, or if neither of the 2 nodes have that attribute.
```
pckroon, line 373: Docstring needs updating

vermouth/processors/apply_rubber_band.py
```diff
@@ -365,7 +365,43 @@ def same_chain(graph, left, right):
...
+    Parameters
+    ----------
+    graph: networkx.Graph
+        A graph the nodes are part of.
+    left:
+        A node key in 'graph'.
+    right:
+        A node key in 'graph'.
+    regions:
+        [(atom_index_start_1,atom_index_end_1),(atom_index_start_2,atom_index_end_2),...] (atom indexes are in theh first chain; include start and end)
```
pckroon, line 384: typo: theh -> the

vermouth/processors/apply_rubber_band.py
```diff
@@ -365,7 +365,43 @@ def same_chain(graph, left, right):
...
+
+    Returns
+    -------
+    bool
+        ``True`` if the nodes are part of the same region.
+    """
+    node_left = graph.nodes[left]
+    node_right = graph.nodes[right]
+    left_resid =node_left.get('resid')
+    right_resid =node_right.get('resid')
```
pckroon, line 394: lint nit: spaces around =

vermouth/processors/apply_rubber_band.py
```diff
@@ -365,7 +365,43 @@ def same_chain(graph, left, right):
...
+    domain_range = [lower <= left_resid <= upper for (lower, upper) in regions]
+    if True in domain_range:
+        True_index = domain_range.index(True)
+        if regions[True_index][0] <= right_resid <= regions[True_index][1]:
+            return True
+        else:
+            return False
+    else:
+        return False
```
pckroon, line 403: This will break if regions overlap. I would suggest the following:
```suggestion
    for region in regions:
        lower, upper = region
        if lower <= left_resid <= upper and lower <= right_resid <= upper:
            return True
    return False

```

vermouth/processors/apply_rubber_band.py
```diff
@@ -365,7 +365,43 @@ def same_chain(graph, left, right):
     node_right = graph.nodes[right]
     return node_left.get('chain') == node_right.get('chain')
 
+def same_region(graph, left, right, regions):
```
pckroon, line 368: I disagree with this signature since other "criterions" (such as `same_chain`) only accept `graph`, `left` and `right`. I think the best option is to either create a SameRegion class with a `__call__` method, or a function factory:
```py
def make_same_region_criterion(regions):
    regions = copy.deepcopy(regions)
    def same_region(graph, left, right):
        ...
    return same_region
```
You can then do something like this:
```py
criterion = make_same_region_criterion(args.regions)
```
This `criterion` is then a function that has the same signature as `same_chain`. This would also save you from having to pass `regions` to all functions here.

pckroon (changes requested):
Hello hello!

First off: Thanks! I love the functionality, and I'm very happy with your contribution.

I have couple of nitpicks regarding minor style/pep8 issues, but you're free to ignore those. I have also indicated a few points where (IMO) the code can be made a bit clearer. Lastly, there is one thing I don't like, and that is the signature of your `same_region` function, since it differs from the other criterions (see also e.g. the docstring of e.g. `apply_rubber_bands`). It also causes you to have to pass regions through the entire call stack, complicating the API along the way. I have indicated 2 possible solutions for this though. I am also missing a few test-cases.

Again, thanks! I hope you're willing (and have the time) to help finish this PR :)
I think the functionality is really useful, and it would make a very nice addition to martinize2!

vermouth/processors/apply_rubber_band.py
```diff
@@ -381,26 +383,27 @@ def same_region(graph, left, right, regions):
...
+    def same_region(graph, left, right):
+        node_left = graph.nodes[left]
+        node_right = graph.nodes[right]
+        left_resid = node_left.get('resid')
+        right_resid = node_right.get('resid')
+        for region in regions:
+            lower, upper = region
+            if lower <= left_resid <= upper and lower <= right_resid <= upper:
+                return True
+            return False
```
pckroon, line 405:
```suggestion
        return False       
```

pckroon (changes requested):
Awesome! One small issue left, and then there's the need for a few tests.

Also a question: what is the intended behaviour regarding chains/molecules? If there's 2 chains (A and B), with regions=[(200,400)], could there be an elastic bond between resid 200 in chain A and resid 300 in chain B?

vermouth/processors/apply_rubber_band.py
```diff
@@ -383,7 +383,7 @@ def make_same_region_criterion(regions):
     right:
         A node key in 'graph'.
     regions:
-        [(resid_start_1,resid_end_1),(resid_start_2,resid_end_2),...] (resids are in the first chain if there are multiple chains. the resid_start and resid_end are included)
+        [(resid_start_1,resid_end_1),(resid_start_2,resid_end_2),...] resid_start and resid_end are included)
```
pckroon, line 386: Trailing bracket ;)

pckroon (commented):
Based on your commit message I'm still a bit unclear on what happens with multiple chains, and what your intended behaviour is. 
I see the following scenarios (that need tests), all assuming region 200-400. I'll write e.g. A200 to mean resid 200 of chain A.

- A200 - A300: Yes EB
- A200 - A201: No EB (due to minimum distance in connectivity matrix)
- A100 - A150: No EB
- A100 - A250: No EB
- B200 - B300: Yes EB
- A200 - B200: ??? EB
- A200 - B300: ??? EB
- A200 - B201: ??? EB
- A200 - B500: ?????? EB

You can ask the same questions if A and B are connected by a disulfide bridge (and are therefore the same molecule).

I think for now you can ignore anything regarding chains/molecules, and /only/ look at resids. This would mean the answers above become yes, yes, yes, no. At a later point (different PR), we should figure out how to combine different criteria, e.g. same chain and region. 
What do you think?

xikunliu:
I tested the code on a pdb with multiple chains and found that with region [(200-400)]:
A200 - A300: Yes EB
A200 - A201: No EB (due to minimum distance in connectivity matrix)
A100 - A150: No EB
A100 - A250: No EB
B200 - B300: No EB ( if the total number of residues in one chain is 400, the resid of B200 will become 600. So if we want to apply rubber bands on A200-A400 as well as B200-B400, we need to use the region flag [(200-400),(600-800)]
A200 - B200: No EB
A200 - B300: No EB
A200 - B201: No EB
A200 - B500: No EB

pckroon:
Hmmn, that's not supposed to happen. The resids of chain B being wrong is probably related to at least one of these: #421 #154 #366. Did you merge the chains for this test?

And could you add some tests for your code, describing the *desired* behaviour in https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/tests/test_apply_rubber_band.py 
Some of these tests will fail due to the resid renumbering, but that's ok for now. We can mark those as expected failures right before we merge this.

fgrunewald:
@pckroon looks like a nice addition. However, can you wait with merging until we figured out what happens to the resids, because that will probably cause some stuff in the EN code to be changed. I'm investigating at the moment. Also we need to test this for performance. I only had a quick glance and saw a deep copy as well as some interactions, which for large proteins could run out of hand. The EN code is already quite slow.

pckroon (approved):

fgrunewald:
@pckroon fine by me if you vouch for it. CI must pass fully though

pckroon:
CI was an unrelated transient failure (hypothesis is too good). So I'll merge this.

fgrunewald:
@pckroon ok good! Please go ahead with merging then.

Issue #425, opened 2022-06-21
rename elnedyn to elnedyn21


fgrunewald:
Currently the elnedyn force-field which is based on protein force-field 21 is just named elnedyn. For book-keeping purposes changing to elnedyn21 was suggested by Paulo

Issue #426, opened 2022-06-21
Proline bead-type in elnedyn is incorrect


fgrunewald:
The bead type for PRO side chain in the elnedyn force-field is incorrect. It is C2 bout should be AC2

PR #427, opened 2022-06-21
Fix/425

fgrunewald:
This PR fixes issues #425 and #426 and #427.


fgrunewald:
@pckroon can you have a look at the failed test and see if you know what's going on? Locally all tests pass. I suspect some updated dependencies could be the issue. Perhaps networkx?

pckroon:
`E       Not equal to tolerance rtol=1e-07, atol=0`. An absolute tolerance of 0 sounds particularly strict. So probably indeed a change in default values due to a version change. I'll check if I can reproduce this locally.

vermouth/data/mappings/elnedyn/modifications.elnedyn21.mapping
```diff
@@ -2,7 +2,7 @@
 [ from ]
 universal
 [ to ]
-elnedyn22p
+elnedyn21
```
pckroon, line 5: Is this intended?

pckroon (commented):
LGTM, "are you sure?" aside. And how about the filename (e.g. `vermouth/data/mappings/elnedyn21/asp.charmm36.elnedyn22p.map`)

vermouth/data/mappings/elnedyn/modifications.elnedyn21.mapping
fgrunewald: nope will fix it

fgrunewald: this came from cleaning the elnedyn directory. I guess there should be a modifications file for each version because modifications cannot stack force-fields but mapping files can ?! To be honest the mapping directories are a mess

vermouth/data/mappings/elnedyn/modifications.elnedyn22p.mapping
```diff
@@ -0,0 +1,129 @@
...
+
+[modification]
+[ from ]
+universal
+[ to ]
+elnedyn22p
+[ from blocks ]
+ASP0
+[ to blocks ]
+ASP0
```
pckroon, line 115: This is currently causing test failures, since there is (apparently) no ASP0 modification in elnedyn22p.

fgrunewald: yep missed this one apparently it exits for another elnedyn

pckroon (approved):

PR #434, opened 2022-06-24
Fix write_pdb to always define open

### LINKED ISSUE #433: Writing Debug PDBs/GRO broken
Here we define open only if deferred file writing is true. However, that means open is undefined if it is not because open is declared as local variable. 

https://github.com/marrink-lab/vermouth-martinize/blob/16de64e3f391a6ee1d3ed91ddeaa7352eece8e2e/vermouth/pdb/pdb.py#L583-L586

pckroon:
Since open is defined as a local variable (if defer_writing) it won't be looked up in the global namespace (if not defer_writing).

Fixes #433

### DIFF
--- a/vermouth/gmx/gro.py
+++ b/vermouth/gmx/gro.py
@@ -142,6 +142,8 @@ def write_gro(system, file_name, precision=7, title='Martinized!', box=(0, 0, 0)
 
     if defer_writing:
         open = deferred_open
+    else:
+        from builtins import open
     with open(str(file_name), 'w') as out:
         out.write(title + '\n')  # Title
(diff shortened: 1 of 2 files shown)


fgrunewald:
@pckroon the gro file writer has the same issue

fgrunewald (approved):
Even though having a test covering this would be best, I think we can ignore it this time.

pckroon:
I agree with leaving out the test for now. I'll open a new issue with the proposed API change where all writers should accept a file handle rather than a path. *That* should come with tests though.
Do we want the same for readers? I guess so, right?

Issue #432, opened 2022-06-23
modification identification too few anchors


fgrunewald:
Modifications that are "too large" are never identified correctly (unless there's no work to be done?). Too large in this context means more than one layer of anchors. 

Workaround for now, probably stick to 1 anchor layer or take entire residue.

pckroon:
Initial code trace summary so far:
1) annotating modifications results in a 'modification' (singular) attribute on all nodes in the relevant residue.
2) repair graph takes this 'modification' attribute and patches the block, repairs it accordingly. It also sets a 'modifications' (plural, I hate myself) attribute on the nodes defined by the modification, including anchors.
3) Canonicalize modifications starts by taking all nodes that have `'PTM_atom'=True` (not identified by blocks in repair graph), *or* that have a non-empty 'modifications' (plural) attribute. These is expanded (across residue boundaries if needed) until all anchors are found. *All* nodes identified this way are to be covered by modifications in the FF.
4) The PTM atoms are then grouped by the resids of the anchors.
5) For all these resids, collect the corresponding atoms from the molecule. *note: this doesn't use the residue_graph, instead it looks at just resid number...*
6) These residues are used as basis on which all PTMs from the FF are matched, where atomnames must match for atoms where `PTM_atom = False`, and elements must match for atoms where `PTM_atom = True`.
7) Then, all these matches are used to solve a set covering: which minimal set of matching PTMs describes all PTM atoms identified in step 3.

I'm now thinking a solution would be to modify step 7, so that it only aims to cover atoms for which `PTM_atom = True`, so don't include the anchors in the covering goal. I'm not quite convinced this would fix the issue though. Do you have the files for me to play around with @fgrunewald?

pckroon:
Running on 1bta:
`martinize2 -f vermouth/tests/data/1bta.pdb -from amber -x derp.pdb -o derp.pdb` works
`martinize2 -f vermouth/tests/data/1bta.pdb -from amber -x derp.pdb -o derp.pdb -modify HIS17:HIS_PROT` breaks.
Bug hunting season has been opened.

PR #436, opened 2022-06-27
Fix/432 by making sure modifications are connected components

pckroon:
`canonicalize_modificaitons.identify_ptms` assumes modifications are connected components. If this assumption is violated, in combination with the `-modify` flag, this causes hard to debug user facing errors (see #432 for more background).

This PR resolves that by
1) Log an error for any modification that is not a connected component.
2) Modify the maxwarn mechanism to count all messages >= warning, meaning an error will cause no files to be written. Note that the `-maxwarn` flag cannot override errors, only warnings.


fgrunewald (approved):
Looks good to me.

PR #442, opened 2022-07-06
add draft for ff-output

fgrunewald:
An initial draft for a ff-file writer. @pckroon some initial feedback would be appreciated. Once the general layout is good, I write tests.

### DIFF
--- a/vermouth/ffoutput.py
+++ b/vermouth/ffoutput.py
@@ -40,7 +40,7 @@ class ForceFieldDirectiveWriter():
         self.stream = stream
         # these attributes have a specific order in the moleculetype section
         self.normal_order_block_atoms = ["atype", "resid", "resname",
-                                         "atomname", "charge_group", "charge", "mass"]
+                                         "atomname", "charge_group", "charge"] #, "mass"]
 
     def write(self):
         """
@@ -54,7 +54,7 @@ class ForceFieldDirectiveWriter():
             self.write_interaction_dict(block.interactions)
 
         for link in self.forcefield.links:
-            self.write_link_header(link)
+            self.write_link_header()
             self.write_atoms_link(link.nodes(data=True))
             self.write_interaction_dict(link.interactions)
             self.write_edges(link.edges)
--- a/vermouth/gmx/rtp.py
+++ b/vermouth/gmx/rtp.py
@@ -260,21 +260,44 @@ def _complete_block(block, bondedtypes):
     # have some control over these dihedral angles through the bondedtypes
     # section.
     all_dihedrals = []
-    for center, dihedrals in itertools.groupby(
-            sorted(block.guess_dihedrals(), key=_dihedral_sorted_center),
-            _dihedral_sorted_center):
+    had_atoms = []
+    for center, dihedrals in itertools.groupby(sorted(block.guess_dihedrals(), key=_dihedral_sorted_center), _dihedral_sorted_center):
+        #print(center)
         if _keep_dihedral(center, block, bondedtypes):
             # TODO: Also sort the dihedrals by index.
             # See src/gromacs/gmxpreprocess/gen_add.cpp::dcomp in the
             # Gromacs source code (see version 2016.3 for instance).
-            atoms = sorted(dihedrals, key=_count_hydrogens)[0]
-            all_dihedrals.append(Interaction(atoms=atoms, parameters=[], meta={}))
+     #       atoms = sorted(dihedrals, key=_count_hydrogens)[0]
+            for atoms in dihedrals:
+                if atoms[::-1] not in had_atoms:
+                    all_dihedrals.append(Interaction(atoms=atoms, parameters=[], meta={}))
+                    had_atoms.append(atoms)
     # TODO: Sort the dihedrals by index
     block.interactions['dihedrals'] = (
         block.interactions.get('dihedrals', []) + all_dihedrals
     )
 
-    # TODO: generate 1-4 interactions between pairs of hydrogen atoms
+    # Add all the angles
+    all_angles = []
+    had_angles = []
+    for atoms in block.guess_angles():
+        if frozenset(atoms) not in had_angles:
+            all_angles.append(Interaction(atoms=atoms, parameters=[], meta={}))
+            had_angles.append(frozenset(atoms))
+
+    # TODO: Sort the dihedrals by index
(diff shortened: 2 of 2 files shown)


vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+            the force-field object to write
+
+        stream: ``
+            the stream to which to write; must have a write method
+        """
+        self.forcefield = forcefield
+        self.stream = stream
+        # these attributes have a specific order in the moleculetype section
+        self.normal_order_block_atoms = ["atype", "resid", "resname",
+                                         "atomname", "charge_group", "charge", "mass"]
```
pckroon, line 43: Maybe these should be class attributes/variables rather than instance attributes/variables.

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+    def write(self):
+        """
+        Write the forcefield to file.
+        """
+        for name, block in self.forcefield.blocks.items():
+            self.stream.write("[ moleculetype ]\n")
+            excl = str(block.nrexcl)
+            self.stream.write(f"{name} {excl}\n")
+            self.write_atoms_block(block.nodes(data=True))
+            self.write_interaction_dict(block.interactions)
```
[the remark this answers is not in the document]
pckroon: We can move this one further up the chain, give it to ForceField. Note that I'm not convinced it's a good idea though, just an idea.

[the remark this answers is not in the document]
pckroon: Either way, group all these into a write_block function

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+
+        for link in self.forcefield.links:
+            self.write_link_header(link)
+            self.write_atoms_link(link.nodes(data=True))
+            self.write_interaction_dict(link.interactions)
+            self.write_edges(link.edges)
```
pckroon, line 60: Group these to write_link

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+        Writes interactions to `self.stream`, with a new
+        interaction directive per type. Meta attributes
+        are kept and written as json parasable dicts.
+
+        Parameters
+        ----------
+        inter_dict: `class:dict[list[vermouth.molecule.Interaction]]`
+            the interaction dict to write
+        """
+        for inter_type in inter_dict:
```
pckroon, line 73: See also `Molecule.sort_interactions`

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+        Writes edges to `self.stream` into the edges directive.
+
+        Parameters
+        ----------
+        edges: abc.iteratable
+            pair-wise iteratable edge list
+        """
+        self.stream.write("[ edges ]\n")
+        for idx, jdx in edges:
+            self.stream.write(f"{idx} {jdx}\n")
```
pckroon, line 93: Only write edges that are not already in interactions?

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+    def write_atoms_block(self, nodes):
+        """
+        Writes the nodes/atoms of the block atomtype directive to `self.stream`.
+        All attributes are written following the GROMACS atomtype directive
+        style.
+
+        Parameters
+        ----------
+        edges: abc.iteratable
+            pair-wise iteratable edge list
```
pckroon, line 104: pairwise edges?

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        self.stream.write("[ atoms ]\n")
+        for idx, (node, attrs) in enumerate(nodes):
+            idx += 1
+            attr_line = " ".join([str(attrs[attr]) for attr in self.normal_order_block_atoms ])
```
pckroon, line 109: What about any other attributes?

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+
+        Parameters:
+        -----------
+        nodes: abc.itertable[tuple(abc.hashable, dict)]
+            list of nodes in form of a list with hashable node-key and dict
+            of attributes. The format is the same as returned by networkx.nodes(data=True)
+        """
+        self.stream.write("[ atoms ]\n")
+        for node_key, attributes  in nodes:
+            attr_line = " {" + " ,".join([f"\"{key}\": \"{value}\"" for key, value in attributes.items()]) + "}"
```
pckroon, line 126: The attr line is defined as json, so abuse that with `json.dumps`

vermouth/ffoutput.py
```diff
@@ -0,0 +1,140 @@
...
+        ----------
+        inter_dict: `class:dict[list[vermouth.molecule.Interaction]]`
+            the interaction dict to write
+        """
+        for inter_type in inter_dict:
+            self.stream.write(f"[ {inter_type} ]\n")
+            for interaction in inter_dict[inter_type]:
+                atom_string = " ".join(interaction.atoms)
+                param_string = " ".join(interaction.parameters)
+                meta_string = "{" + " ,".join([f"\"{key}\": \"{value}\"" for key, value in interaction.meta.items()]) + "}"
```
pckroon, line 78: `json.dumps`

pckroon (changes requested):
Some comments and design remarks. I don't think sphinx will like your docstrings at the moment, but we can take a look at that later.

What I'm mainly missing is the why. Why do we want/need a FF writer?
As for the functionality of the code, I guess it looks ok, but the proof will be in the pudding. Ideally you'd make some tests where hypothesis creates a ForceField, you write it, you parse it, and you confirm they're the same. Undoubtedly there are corner cases missing here, such as `[variables]`, and all the thinks links can do...

fgrunewald:
@pckroon the why is easy: Because we already use that in 2 projects relating to automatic link generation from template itp-files. Also it is useful in the API as you can read in ff-files manipulate them not manually and then write the links again e.g. look for all links that do this and change attribute there. 

About the corner cases, they are not so bad. The input files are more condensed because of some hacks and tricks but once they are at link level everything gets resolved. The writer will not write the same force-field by line-comparison but a more verbose but also more robust version. For examples all variables should be resolved in the link, they just get written directly to the links.

PR #449, opened 2022-07-29
fix block name HSD

fgrunewald:
residue HSD had the BB resname HIS. This should change.


jan-stevens:
Fixes the problem for me. (Tested it on the OpuAC protein)

jan-stevens (approved):
Fixes the problem for me. (Tested it on the OpuAC protein)

PR #455, opened 2022-08-31
Skip EN generation for no-coordiantes

fgrunewald:
If the selector returns no coordinates the EN code break.

### DIFF
--- a/vermouth/processors/apply_rubber_band.py
+++ b/vermouth/processors/apply_rubber_band.py
@@ -23,7 +23,9 @@ import copy
 from .processor import Processor
 from .. import selectors
 from ..graph_utils import make_residue_graph
+from ..log_helpers import StyleAdapter, get_logger
 
+LOGGER = StyleAdapter(get_logger(__name__))
 
 # the bond type of the RB
 DEFAULT_BOND_TYPE = 6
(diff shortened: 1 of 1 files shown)


vermouth/processors/apply_rubber_band.py
```diff
@@ -303,36 +303,38 @@ def apply_rubber_band(molecule, selector,
...
+            to_key = idx_to_node[selection[to_idx]]
+            force_constant = constants[from_idx, to_idx]
+            length = distance_matrix[from_idx, to_idx]
+            if force_constant > minimum_force:
+                molecule.add_interaction(
+                    type_='bonds',
+                    atoms=(from_key, to_key),
+                    parameters=[bond_type, length, force_constant],
+                    meta={'group': 'Rubber band'},
+                )
```
pckroon, line 337: I would prefer the opposite check: `if not coordinates: return`.
As a side note, should we care about `nan` coordinates?

fgrunewald: I thought you preferred it this way around :rofl: I think nan coordinates are not top priority because they'd be useless as simulation input anyways. In principle we should raise an Error.

pckroon (approved):

PR #456, opened 2022-08-31
implement small molecule DB

fgrunewald:
43 small molecules from Riccardo's small molecule DB together with their charmm files. This one depends on the EN fix


pckroon:
Is there anything specific you need me to take a look at? 100k+ lines is a bit much to go through ;)
Maybe add the appropriate citation?

fgrunewald:
@pckroon I don't think so. It's just a bunch of parameters for the database

fgrunewald:
> Just to make sure, there doesn't need to be citation for these?

I meant you don't need to review; for sure I need to add the citations. But Siewert has a whishlist of data he wants to add. That goes all in here.

vermouth/data/mappings/modifications.mapping
```diff
@@ -199,3 +199,75 @@ N   BB
...
+OH	HH
+O	C
+CD1	CG
+CE1	CZ
+CE2	CD2
+OH P1
+P1 O2
+P1 O3
+P1 O4
+H2 O2
```
pckroon, line 261: This doesn't look correct to me. `[ from nodes ]`  and `[ from edges ]` should only list nodes/edges that are not already described by `[ from blocks ]` (i.e. the modification TRYPHOS), and even then only insofar the mapping is different from the unmodified block.

pckroon (changes requested):

vermouth/data/mappings/modifications.mapping
fgrunewald: BUT modifications need to be connected components and if I start stripping them I get the disconnected component error at some point. If I don't include the anchor to the BB it completly missed the PTM. This one works

fgrunewald: I'll test it and let you know. Could be I made a typo in one of the earlier toys and this worked because I just copied it all from the rtp file.

fgrunewald: seems to have worked

fgrunewald:
Point of order on the phosphorylation. The additional bead is added at the end of the molecule not with the residue

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -478,3 +478,62 @@ def test_apply_rubber_bands_same_regions(test_molecule, regions, chain_attribute
...
+        res_min_dist=3)
+    process.run_molecule(test_molecule)
+    assert test_molecule.interactions['bonds'] == []
+
+def test_bail_out_on_nan(caplog, test_molecule):
+    """
+    Test the the EN processor bails out when nan coordinate
+    is found.
+    """
+    test_molecule.nodes[0]['position'] = np.array([np.nan, np.nan, np.nan])
```
pckroon, line 512: Are we sure this doesn't affect other tests using the test_molecule fixture? Maybe make sure by setting its scope to 'function' (the current default)?
https://docs.pytest.org/en/6.2.x/reference.html#pytest-fixture-api

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -478,3 +478,62 @@ def test_apply_rubber_bands_same_regions(test_molecule, regions, chain_attribute
...
+        domain_criterion=domain_criterion,
+        res_min_dist=3)
+    process.run_molecule(test_molecule)
+
+    required_warning = ("Found nan coordinates in molecule testmol. "
+                        "Will not generate an EN for it. ")
+    record = caplog.records[0]
+    assert record.getMessage() == required_warning
+
+    assert test_molecule.interactions['bonds'] == []
```
pckroon, line 539: `assert len(records) == 1`

pckroon (changes requested):
Small nitpicks about one of the new tests, looks good to me otherwise :)

pckroon:
> Point of order on the phosphorylation. The additional bead is added at the end of the molecule not with the residue

This happens because the new bead is assigned the next available atom ID, and at the end of martinize atoms are sorted by atom id. We *could* aggressively renumber atom ids, but is it an issue?

fgrunewald:
Yeah this will be an issue for many people. It makes the files harder to read and also it messes with the residue completeness. Most analysis programs (falsely) assume atoms in one residue are consecutive. Perhaps it is even enforced by some file formats. But if it is not in the code now it is hard to add. I've been looking into it and this type of renumbering get's exteremly slow even for moderately large proteins. So I suggest since we don't have any modifications for the M3 FF yet, we just ignore this. But be prepared

pckroon (approved):

pckroon:
I just noticed citations are still missing for these. Iff you can't get around to it today could you open an issue so we don't forget?

Issue #384, opened 2021-10-14
modifications parser bug


fgrunewald:
Sometimes in ff input modifications are referred to as 'modification' and sometimes as 'modifications'. The following line breaks when trying to parse interactions directives located in the modifications directive. 

https://github.com/marrink-lab/vermouth-martinize/blob/4261733ffa187cda80ac4f55510d2dc794d781ea/vermouth/ffinput.py#L802

PR #458, opened 2022-09-05
fix modifications

fgrunewald:
This fixes issue #384. However, there is one caveat: Now the modification can have an order attribute. This originates from the way we treat the atoms when making interactions. Even though we maybe don't want this, having order attributes is most consistent in my opinion with the fact that modifications is a subclass of `Link`. I would propose to later down the line disentangle the Modifications object from the Link, because they are really very separate things. As far as I can asses it makes no difference between having order attributes as to not having them in the current code.

### DIFF
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -384,7 +384,7 @@ class FFDirector(SectionLineParser):
     @SectionLineParser.section_parser('moleculetype', 'patterns')
     @SectionLineParser.section_parser('moleculetype', 'features')
     @SectionLineParser.section_parser('moleculetype', 'non-edge')
-    @SectionLineParser.section_parser('modifications', 'non-edge')
+    @SectionLineParser.section_parser('modification', 'non-edge')
     def _invalid_out_of_link(self, line, lineno=0):
         raise IOError('The "{}" section is only valid in links.'
                       .format(self.section[-1]))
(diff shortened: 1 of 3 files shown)


vermouth/ffinput.py
```diff
@@ -409,7 +409,7 @@ def _modification_atoms(self, line, lineno=0):
         tokens = collections.deque(_tokenize(line))
         _parse_link_atom(tokens, self.current_modification,
                          defaults={'PTM_atom': False},
-                         treat_prefix=False)
+                         treat_prefix=True)
```
pckroon, line 412: Why this change?

pckroon (dismissed):
Small question, but lgtm otherwise. I agree that some restructuring is required, but that can be postponed for quite some time.

vermouth/ffinput.py
fgrunewald: because otherwise you end up with some nodes having the order attribute and others not having the order attribute.

fgrunewald: but apparently it messes with some part of the code

pckroon: Ok. I guess that will get fixed at some point in the not-so-near future when we redo the parsers again.

fgrunewald: @pckroon can you see why this breaks the code? We either have the option to see why the order attribute on modifications messes with the code or we have to copy the entire base parser for interactions again only for modifications. The order attribute is so hard-coded into the interactions parser it is not even funny anymore. This is the last PR that has to be done for publication.

PR #460, opened 2022-09-06
Comments

fgrunewald:


pckroon:
As an alternative, should `parse` call the line methods with arguments `line, lineno, comment` instead of creating the current_comment attribute? Bit more invasive, but maybe it makes the resulting API cleaner?

fgrunewald:
@pckroon that can be done! And then to solve the problem of Chris you could think of annotating the ss-sequence in the atom types section of the itp file as comment rather than trying to parse a header string like so:

[ moleculetype]
protein 1
[ atoms ]
1  P2  ALA 1 ... ; ss_code

Citations I'm not in favour of parsing form the itp file

pckroon:
Sounds like a reasonable idea. Also note that annotating beads is not exclusive with putting the SecStruct string in the header. The latter is useful when you want to invoke martinize again with the same secstruct. 
Assuming we want to annotate atoms with comments about the node in question, do we want to pick a sane format (json?) and just write them all?

> Citations I'm not in favour of parsing form the itp file

Sure. And there's indeed a difference between parsing and interpreting.

PR #465, opened 2022-09-07
merge eunit and region flag

### LINKED ISSUE #459: Eunit flag Broken
Someone messed up when implementing #422. The lines below are actually non-sense because they overwrite the e-unit criterion. 

https://github.com/marrink-lab/vermouth-martinize/blob/aa53b3e3ca734032518bd56b3b8dad0831653612/bin/martinize2#L737-L741

fgrunewald:
This fixes #459 at least partially. As stated in the issue there is still some ambiguity as to what exactly should happen if a resid region is defined for a molecule with more than one chain and same resids per chain


bin/martinize2
```diff
@@ -442,14 +442,11 @@ def entry():
...
-                          choices=['chain', 'molecule', 'all'],
                           help=('Establish what is the structural unit for the '
                                 'elastic network. Bonds are only created within'
-                                ' a unit.'))
-    rb_group.add_argument('-region', dest='rb_region', default='molecule',
-                          help=('Apply rubber bands only within the specified regions. '
-                                'The format is <start_resid_1>:<end_resid_1>, <start_resid_2>:<end_resid_2>...'
-                                'Start and end resids are included'))
+                                ' a unit. Options are molecule, chain, all, or a'
+                                'specified region defined by resids, with follwing'
```
pckroon, line 448:
```suggestion
                                'specified region defined by resids, with following'
```

vermouth/tests/data/integration_tests/tier-1/EN_chain/README
```diff
@@ -0,0 +1,11 @@
+### DETAILS: 
+
+# Human Insluin HOMRO dimer
```
pckroon, line 3:
```suggestion
# Human Insulin HOMRO dimer
```

vermouth/tests/data/integration_tests/tier-1/EN_chain/README
```diff
@@ -0,0 +1,11 @@
...
+# 2 chains which are connecte by disulphide bridge
```
pckroon, line 4:
```suggestion
# 2 chains which are connected by disulphide bridge
```

vermouth/tests/data/integration_tests/tier-1/EN_region/README
```diff
@@ -0,0 +1,13 @@
+### DETAILS: 
+
+# Human Insluin HOMRO dimer
```
pckroon, line 3:
```suggestion
# Human Insulin HOMRO dimer
```

vermouth/tests/data/integration_tests/tier-1/EN_region/README
```diff
@@ -0,0 +1,13 @@
...
+# 2 chains which are connecte by disulphide bridge
```
pckroon, line 4:
```suggestion
# 2 chains which are connected by disulphide bridge
```

pckroon (changes requested):
Few typos, nothing major.

I am starting to have doubt to what's the desired behaviour though. Is it not desirable to be able to combine `-eunit chain` with `-region 200:400`? The behaviour would be that EB only get made where both conditions are true.
Also, CLI nitpick, when I provide an `-eunit` I'd argue the `-elastic` is implied, but that can be shelved for a while.

fgrunewald:
As far as the behaviour goes the `eunit=chain` is really just an edgy edge case. It only comes to be important when two chains, which are labelled as such are connected by a disulphide bridge. Any thing else will end as two separate molecules or one molecule. Thus I think it is fine having the eunit flag being either all, molecule, chain, or residrange.

pckroon:
Then maybe chain + region was a poorly chosen example. 
How about molecule + region? You may want an elastic network only within a molecule/chain, but only between resids 200-400. So if you have 2 homodimers that are close together you only want EN between resid 200-400, but not between the two dimers. I think that's not possible with the current CLI suggestion

fgrunewald:
@Tsjerk any thoughts on this

pckroon (approved):
Ok, that works for me. I'll open a fresh issue.

PR #473, opened 2022-10-19
Update aminoacids.ff

fgrunewald:
patch HSP amino acid name & add definitions of HSE


pckroon (commented):
Changes look OK to me. Would it be possible to include the test-case into one of the integrative cases we already have, e.g. 1mj5, by just renaming one of the residues there to HSE?

fgrunewald:
@pckroon no that's not possible because the 1mj5 test-case uses Amber naming convention. In fact the charmm names were not tested at all

pckroon (approved):
Ah ok, sure. 
Going forward I do think there is value in limiting integrative test cases in size though.

PR #474, opened 2022-10-20
Add option to generate log entries from .ff files

pckroon:
This PR adds the option to have martinize2 generate log statements based on .ff files. Blocks, Modifications and Links get debug, info, warning, and error sections, in which each line will become a log statement with the appropriate level.
This is still WIP, since I kind of want to be able to do e.g. 
```
[ info ]
Created a disulfide bridge between residues {SC1['resid']} and {>SC1['resid']}.
```
but I'm not quite sure yet how painful that's going to be (shouldn't be too bad though, `format` will do the heavy lifting). Does require some bookkeeping .

Question, should each line in a section result in a separate log message, or should these be aggregated to a single entry?

### DIFF
--- a/vermouth/ffinput.py
+++ b/vermouth/ffinput.py
@@ -181,9 +181,12 @@ class FFDirector(SectionLineParser):
             self.current_modification.citations.update(self.citations)
             self.force_field.modifications[self.current_modification.name] = self.current_modification
 
-    def get_context(self, context_type):
+    def get_context(self, context_type=''):
+        if not context_type:
+            context_type = self.section[0]
         possible_contexts = {
             'block': self.current_block,
+            'moleculetype': self.current_block,
             'link': self.current_link,
             'molmeta': self.current_link,
             'modification': self.current_modification,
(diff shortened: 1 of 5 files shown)


pckroon:
With the last commit the following *should* be possible. I gave it a quick test, but I probably missed some edge cases...

```
;; Cystein bridge
[ link ]
resname "CYS"
[ constraints ]
SC1 >SC1 1 0.24 {"comment": "Disulfide bridge"}
[ features ]
disulfide
[ info ]
Added a disulfide bridge between {SC1[chain]}-{SC1[resname]}{SC1[resid]} and {>SC1[chain]}-{>SC1[resname]}{>SC1[resid]}.

```
Ditto for Blocks and Modifications. During the formatting SC1 and >SC1 in the example above are replaced by the corresponding dictionaries with node attributes.

vermouth/ffinput.py
```diff
@@ -180,9 +181,12 @@ def finalize_section(self, previous_section, ended_section):
             self.current_modification.citations.update(self.citations)
             self.force_field.modifications[self.current_modification.name] = self.current_modification
 
-    def get_context(self, context_type):
+    def get_context(self, context_type=''):
+        if not context_type:
+            context_type = self.section[0]
```
fgrunewald, line 186: Do we really need this? As this is never called in the tests I'm not sure it makes any sense to set this default. You can just add the context type explicitly to the decorator right? If we keep it I think the context type should also be removed from the other decorators where it is not needed.

vermouth/processors/do_links.py
```diff
@@ -272,6 +272,7 @@ def run_molecule(self, molecule):
         _nodes_to_remove = []
         for link in links:
             matches = match_link(molecule, link)
+
```
fgrunewald, line 275: is the empty line on purpose?

vermouth/processors/do_links.py
```diff
@@ -292,5 +293,10 @@ def run_molecule(self, molecule):
                         interaction = _build_link_interaction_from(molecule, interaction, match)
                         molecule.add_or_replace_interaction(inter_type, *interaction, link.citations)
 
+                for loglevel, entries in link.log_entries.items():
+                    for entry, fmt_args in entries.items():
+                        fmt_args = fmt_args + [match]
+                        molecule.log_entries[loglevel][entry] += fmt_args
```
fgrunewald, line 299: I'm a bit concerned about performance, but if we have not too many debug statements in links for now I think it is fine.

fgrunewald (commented):
Overall looks good to me. Have a look at the two small comments I made. One question that came to my mind. Would it be good/needed to have force-field wide log statements and warnings. Like for example "You choose Martini2, are you sure cuz it sucks for protein interactions?"

vermouth/processors/do_links.py
pckroon: I did some premature optimization, and populate the log_entries attribute only if they'd result in actual log messages being emitted.

pckroon:
> Overall looks good to me. Have a look at the two small comments I made. One question that came to my mind. Would it be good/needed to have force-field wide log statements and warnings. Like for example "You choose Martini2, are you sure cuz it sucks for protein interactions?"

I played around with this, but I'm running into limitations of the SectionLineParser. So good idea, but no.

fgrunewald:
@pckroon what is the status of this one?

pckroon:
Ready as far as I'm concerned. The log statements at FF level we can leave for later I think.

fgrunewald:
How about the test coverage? Do you care to fix it?

pckroon:
Honestly not really. If this feature turns out to not work perfectly it's not too big a deal.

fgrunewald (commented):
@pckroon I'll test this one and then we merge.

fgrunewald:
@pckroon is there a way to line-break a warning ? Just using two lines scrambles the print output

pckroon:
No, there isn't. Multiline statements are something for the parser overhaul as far as I'm concerned. For now just make long lines.

fgrunewald:
@pckroon I'm in principle fine with this PR. Seems to work. Except I wonder if it is a good idea to print the log statements in the end or actually where the processor is called?

pckroon:
Printing them in the end makes deduplication possible/much easier. Also, at the end they're more likely to be noticed I guess.

fgrunewald:
@pckroon I'm fine with merging this. I want to adopt it to polypyl asap as well

PR #475, opened 2022-10-21
Update gh actions

pckroon:
The GH workflow started complaining about soon-to-be deprecated node.js versions, so I figured I'd update the actions we use. In addition, I took the liberty to update the python versions we test against. In particular, I dropped py3.6 and added py3.10

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -68,7 +68,7 @@ jobs:
     - name: Set up Python 3.10
       uses: actions/setup-python@v4
       with:
-        python-version: 3.10
+        python-version: '3.10'
         cache: pip
         cache-dependency-path: |
           **/setup.cfg
@@ -93,7 +93,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v4
         with:
-           python-version: 3.10
+           python-version: '3.10'
            cache: pip
            cache-dependency-path: |
              **/setup.cfg
@@ -120,7 +120,7 @@ jobs:
     - name: Set up Python 3.10
       uses: actions/setup-python@v4
       with:
-        python-version: 3.10
+        python-version: '3.10'
         cache: pip
         cache-dependency-path: |
           **/setup.cfg
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,15 +19,15 @@ jobs:
     
     strategy:
       matrix:
-       py_version: ["3.6", "3.7", "3.8", "3.9"] 
+       py_version: ["3.7", "3.8", "3.9", "3.10"]
        include:
-         - py_version: "3.6"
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -124,7 +128,7 @@ jobs:
         --outdir dist/
         
     - name: Publish package to PyPI
-      uses: pypa/gh-action-pypi-publish@master      
+      uses: pypa/gh-action-pypi-publish@release/v1
```
fgrunewald, line 131: why this change? Does it work?

.github/workflows/deploy.yml
```diff
@@ -59,11 +60,12 @@ jobs:
...
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.7
-      uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
       with:
-        python-version: 3.9
+        python-version: 3.10
+        cache: pip
```
fgrunewald, line 68: what's the reason behind caching pip?

.github/workflows/run_tests.yml
```diff
@@ -21,27 +21,30 @@ jobs:
     
     strategy:
       matrix:
-       py_version: ["3.6", "3.7", "3.8", "3.9", "3.6-dev", "3.7-dev", "3.8-dev", "3.9-dev"] 
+       py_version: ["3.7", "3.8", "3.9", "3.10", "3.7-dev", "3.8-dev", "3.9-dev", "3.10-dev"]
```
fgrunewald, line 24: Here we can actually remove all the dev versions, because GHA will only collect the latest release version not the actual deve versions. This is an actions bug we discussed earlier I believe.

fgrunewald (changes requested):
Looks good. Only thing is we should remove the devs because they are not actually running the devs. See [here](https://github.com/marrink-lab/polyply_1.0/issues/160). Could have been fixed with a new installer, but we should check.

.github/workflows/deploy.yml
pckroon: Because I don't want to run off of their master branch. I'd rather depend on specific, released versions.

.github/workflows/deploy.yml
pckroon: Because I think that should speed things up a little. There is a catch though, I'm not 100% sure we'll ever get the newest versions of our dependencies this way, since we don't list explicit version numbers in our requirements. And those requirements files are used to create a hash-index for the cache. Maybe rip it out again after all?

fgrunewald (approved):

fgrunewald:
@pckroon I think this one is good to go

pckroon:
Feel free to merge when the CI finishes.

fgrunewald (approved):

Issue #472, opened 2022-10-19
Problems with -nt -cter and -nter when Martinizing from Charmm36 to M3


MikkelDA:
### These problems were encountered in the latest development version of martinize2/vermouth which was downloaded 18/10-2022.

I should preface by saying that martinize2 successfully martinizes the terminals if no terminal-relating flags are given (though it results in charged terminals).

**-nt with a proline C-terminal**
Martinization causes the following warning:
"WARNING - unmapped-atom - These atoms are not covered by a mapping. Either your mappings don't describe all atoms (bad idea), or, there's no mapping available for all residues. ['7285A-PRO500:OXT']"

The AA proline C-terminal is shown below though the error persists event if the last three atoms are removed and "7284 NT" is changed to "7284  O" to reflect the standard proline mapping as described in the "pro.charmm36.map" file for martini3001:
ATOM   7271  N   PRO A 500      51.433  37.411  30.113  1.00  0.00           N
ATOM   7272  CD  PRO A 500      50.019  37.684  30.421  1.00  0.00           C
ATOM   7273  HD1 PRO A 500      49.948  38.596  31.013  1.00  0.00           H
ATOM   7274  HD2 PRO A 500      49.572  36.802  30.881  1.00  0.00           H
ATOM   7275  CA  PRO A 500      51.751  37.549  28.681  1.00  0.00           C
ATOM   7276  HA  PRO A 500      52.256  36.651  28.327  1.00  0.00           H
ATOM   7277  CB  PRO A 500      50.381  37.674  28.003  1.00  0.00           C
ATOM   7278  HB1 PRO A 500      50.390  38.528  27.324  1.00  0.00           H
ATOM   7279  HB2 PRO A 500      50.141  36.742  27.492  1.00  0.00           H
ATOM   7280  CG  PRO A 500      49.391  37.913  29.082  1.00  0.00           C
ATOM   7281  HG1 PRO A 500      49.029  38.940  29.021  1.00  0.00           H
ATOM   7282  HG2 PRO A 500      48.547  37.236  28.956  1.00  0.00           H
ATOM   7283  C   PRO A 500      52.596  38.773  28.364  1.00  0.00           C
ATOM   7284  NT  PRO A 500      52.776  38.809  27.381  1.00  0.00           N  
ATOM   7285  HT1 PRO A 500      52.409  38.099  26.780  1.00  0.00           H    
[...]

PR #476, opened 2022-10-21
Proline termini mappings

pckroon:
This PR fixes part of #472, namely the proline termini mappings


fgrunewald:
@pckroon I think this one is fine as well. Even if the names are not 100% aligned I think it should still work. We can worry about Amber implementation later.

PR #477, opened 2022-10-24
Add charmm data for NCAP and CCAP mappings

pckroon:
Fixes the other half of #472, together with #476.


vermouth/data/force_fields/charmm/modifications.ff
```diff
@@ -46,6 +46,38 @@ HN2 {"element": "H", "PTM_atom": true, "replace": {"atomname": "HN2"}}
...
+[ modification ]
+CCAP-ter
+[ atoms ]
+C {"element": "C"}
+O {"element": "O"}
+OC {"element": "O", "PTM_atom": true}
+CC1 {"element": "C", "PTM_atom": true}
+HCC1 {"element": "H", "PTM_atom": true}
+HCC2 {"element": "H", "PTM_atom": true}
+HCC3 {"element": "H", "PTM_atom": true}
```
fgrunewald, line 58: Are these the CHARMM names?

fgrunewald (commented):
@pckroon Couple of questions:

- are the names the CHARMM names? 
- If not does it matter? I think not really does it?
- If the above answer is no, can we dump the same files into the amber modifications?
- Did you test the files?

pckroon:
> are the names the CHARMM names?
Probably not. Where could I find those? Are they standardized to begin with?
> If not does it matter? I think not really does it?
Not really I think.
> If the above answer is no, can we dump the same files into the amber modifications?
Sure.
> Did you test the files?
Small scale, yes. I ran m2 with 1ubq with these termini and checked the itp

fgrunewald (approved):

PR #484, opened 2022-11-17
Correct the version check in Molecule.remove_interaction

### LINKED ISSUE #483: Selectively removing interactions from molecule
### Discussed in https://github.com/marrink-lab/vermouth-martinize/discussions/482

Regarding reading in itps: it would be useful to have read back in header information and meta attributes for different bond groups. The former would be useful for checking aspects of secondary structure assignment, and the latter for itp editing when needed.

<div type='discussions-op-text'>

<sup>Originally posted by **csbrasnett** November 16, 2022</sup>
I need to edit an itp to remove parts of the elastic network, as determined by secondary structure assignment.

As a minimal example, I would have thought something like this should work:

```python
# loop over regions where we want to keep EN
for i in wanted_region:
     # look at all the elastic network interactions
     for j in all_EN_interactions:
         # compare atom indicies of the EN interaction to the boundaries of the region we want to keep
         # if they're within, keep them
         if all_EN_interactions[j] in wanted_region[i]:
             continue
         # if they're outside, remove the interaction
         else:
             ff.blocks[molname].remove_interaction('bonds', atoms = all_EN_interactions[j])
```

However, I keep getting a KeyError
`KeyError: "Can't find interaction of type bonds between atoms ['0', '5'] and with version 0"`
even though I can definitely see that there's an interaction there. I've tried various versions of the atoms argument in remove_interaction (give as an array, a list, a list of strings, etc.) but none of these seem to work.

I guess it might want me to give a version explicitly, but this doesn't seem to work.

Perhaps this should be raised as a separate issue, but on this last point when I read in the itp file that's been generated with martinize2, the meta information isn't read back in by vermouth - I think this might be quite a useful feature to have for this sort of purpose?

</div>

pckroon:
Fixes #483

I need a facepalm emoji


vermouth/tests/test_molecule.py
```diff
@@ -1279,3 +1279,49 @@ def test_to_molecule():
...
                              meta={'b': 1}),]
     assert ref_bonds == test_molecule.interactions['bonds']
+
+
+@pytest.mark.parametrize('atoms, bonds, interactions, removed, expected', [
+    # empty molecule
+    ([], [], [], {}, {}),
+    # interactions that all need to be removed
+    ([1, 2], [(1, 2)], [('bond', (1, 2), {})], {'type_': 'bond', "atoms": (1, 2)}, {}),
+    # Molecule with interactions of different types of which some need to be removed
```
fgrunewald, line 1289: to future proof this I would add a test with populated meta and parameter dict. Just to make sure if someone touches it in future the test will fail.

fgrunewald (changes requested):
Did I get the gist right, that essentially vermouth did not remove interactions based on the specific version but just the first it found?

Fix looks good, except from 1 request and a questions.

pckroon:
> Did I get the gist right, that essentially vermouth did not remove interactions based on the specific version but just the first it found?

Not quite. It would remove the first interaction it found, *IF* it had a version that is not 0.

fgrunewald (approved):

fgrunewald:
@pckroon it is a mystery that this didn't break anything. I wonder if the ff files are written to circumvent this problem. It will be interesting to see if this breaks anything polyply related. I'll test this real quick. Can you hold off with merging for a sec

PR #488, opened 2022-11-25
Add correct zenodo DOI

fgrunewald:
add zenodo DOI for release v0.8.0


pckroon (changes requested):
LGTM apart from the failing tests.
`Get:6 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 dssp amd64 4.0.4-1 [511 kB]`
We need to deal with the DSSP problem. Do you want to adapt to dssp 4.x, or pydssp?

fgrunewald:
@pckroon We're now using DSSP v4? I assume fixing v4 DSSP breaks the other DSSP versions? Can't we make a DSSP version check and raise an error if the version is too new for the time being and get a specific version? Not sure about how it installs DSSP though and how the labs around the world use it

pckroon:
> We're now using DSSP v4?

The default switched in the CI pipeline.
 
>  I assume fixing v4 DSSP breaks the other DSSP versions?

Yarp.

> Can't we make a DSSP version check and raise an error if the version is too new for the time being and get a specific version?

We can add a check without too many issues, but I'm not sure how to get the CI to install a specific version

PR #489, opened 2022-11-28
Install dssp=3.0 during CI

pckroon:
The github CI runner changed to 22.04 (from 20.04), which has DSSP version 4.0 (https://packages.ubuntu.com/jammy/dssp), which doesn't work with vermouth. So try to specify a different version. If this works we can add a version check to the DSSP processor.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,7 +15,7 @@ env:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     
     strategy:
       matrix:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -45,7 +45,7 @@ jobs:
           **/pyproject.toml
     - name: Install dependencies part I
       run: |
-        sudo apt-get install dssp=3.0.0-3build1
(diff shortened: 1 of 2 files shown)


pckroon:
The only version available to Ubuntu22.04 is 4.0.4-1.

pckroon:
Ok, lgtm. Feel free to approve and merge.

pckroon:
Actually, also set the ubuntu version in the deploy.yaml. Otherwise deployment will still be broken

pckroon (commented):
approved. Github thinks this is my PR, so I can't GH approve it.

.github/workflows/deploy.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           **/pyproject.toml
     - name: Install dependencies part I
       run: |
-        sudo apt-get install dssp
+        sudo apt-get install dssp=3.0.0-3build1
```
fgrunewald, line 48: do we want to keep this?

fgrunewald (approved):

.github/workflows/deploy.yml
pckroon: Meh. Don't particularly care either way, but probably not

fgrunewald (approved):

PR #491, opened 2022-12-02
check dssp version before running DSSP

fgrunewald:
Quick check to see if DSSP has the correct version.

### DIFF
--- a/vermouth/dssp/dssp.py
+++ b/vermouth/dssp/dssp.py
@@ -195,7 +195,7 @@ def run_dssp(system, executable='dssp', savefile=None, defer_writing=True, versi
     """
     # check version
     process = subprocess.run(["dssp", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout_UTF8 = process.stdout.decode('UTF8')
+    version_found = process.stdout.decode('UTF8')
     if version not in stdout_UTF8:
         raise DSSPError('Vermouth currently only supports DSSP version 3.0.0.')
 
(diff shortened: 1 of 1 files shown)


vermouth/dssp/dssp.py
```diff
@@ -191,6 +193,12 @@ def run_dssp(system, executable='dssp', savefile=None, defer_writing=True):
     read_dssp2
         Parse a DSSP output.
     """
+    # check version
+    process = subprocess.run(["dssp", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout_UTF8 = process.stdout.decode('UTF8')
```
pckroon, line 198:
```suggestion
    version_found = process.stdout.decode('UTF8')
```

pckroon (approved):
Do we want to parse the version with e.g. `pkg_resources.packaging.version.parse` before comparing?
Optional though.

fgrunewald:
@pckroon I thought looking for "version 3.0.0" in the output string is a bit safer cuz who knows if DSSP will format their output appropriately. Since this is a quick-fix I think it's not needed but as you like

vermouth/dssp/dssp.py
```diff
@@ -191,6 +193,12 @@ def run_dssp(system, executable='dssp', savefile=None, defer_writing=True):
     read_dssp2
         Parse a DSSP output.
     """
+    # check version
+    process = subprocess.run(["dssp", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    version_found = process.stdout.decode('UTF8')
+    if version not in stdout_UTF8:
```
pckroon, line 199:
```suggestion
    if version not in version_found:
```

pckroon (approved):
Good point.

pckroon (approved):

PR #494, opened 2022-12-22
add citation bib text to README

fgrunewald:
Adding Martinize2 citation to README


pckroon (changes requested):
Maybe add the DOI?

fgrunewald:
@pckroon there is no DOI for preprints. they have their own archive ID plus we have the zenodo doi in the header. I think that's enough

pckroon (approved):
Didn't realize the preprint doesn't have a DOI

Issue #364, opened 2021-05-05
What DSSP version does martinize2 expect for -dssp [executable] flag


KasperBuskPedersen:
Hi 

I have problems with the -dssp flag and cannot get it to work properly. I have tried two dssp versions 4.0 and 2.0.4. Both fail because of formatting in the temporary .pdb made by martinize2. 

In the case of version 4.0 dssp fails because the temporary .pdb is missing a dummy CRYST 1 record. Putting one into the temporary file lets me run dssp.

I the case of version 2.0.4, ,my dssp fails because the atom names in the temporary .pdb start at character position 13 (1-indexed). Moving all atom names to position 14 (all other things equal) , I can excecute dssp normally. 

Both dssp versions can run normally on the input pdb to martinize2

Both fixes unfortunately dont allow me to run martinize2 -dssp succesfully. For the moment i have reverted to input the secondary structure manually as a sequence.

I cannot find which dssp version is expected. Would it be an idea to either distribute a dssp executeable with vermouth or clearly state which dssp is a prerequisite on the main readme at this page? :)

Thank you in advance 
Best regards
Kasper

KasperBuskPedersen:
Ok i found inside dssp.py in python3.7/site-packages/vermouth/dssp that version 2 or 3 is expected, sorry about that. However I still get errors for version 2.0.4

pckroon:
Hello,
thanks for the report!
Let me split this out into 3 separate topic:
1) We need to document in the README which versions of DSSP are supported.
2) We need to start writing CRYST 1 records to support DSSP 4. Although it seems the output format for dssp4 is wildly different...
3) We need to take a critical look at how we format atomnames in PDBs. The problem is that atomnames are commonly 4 characters wide, which means we may need to wrap numbers around, and it may also depend on the element of the atom. This will also affect the parsing side of things.

For reference, I can run martinize2 with
```bash
$ dssp --version
mkdssp version 3.0.0

If you use DSSP, please cite: 
Touw WG, Baakman C, Black J, te Beek TA, Krieger E, Joosten RP & Vriend G.
A series of PDB-related databanks for everyday needs.
Nucleic Acids Res. (2015) 43, D364-D368. doi: 10.1093/nar/gku1028.

The original DSSP reference is: 
Kabsch W & Sander C.
Dictionary of protein secondary structure: pattern recognition of hydrogen-bonded and geometrical features.
[...]

PR #506, opened 2023-03-01
[VIP] Bug fix and improvement in DSSP version check

juminlee:


vermouth/dssp/dssp.py
```diff
@@ -194,10 +195,14 @@ def run_dssp(system, executable='dssp', savefile=None, defer_writing=True, versi
         Parse a DSSP output.
     """
     # check version
-    process = subprocess.run(["dssp", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    version_found = process.stdout.decode('UTF8')
-    if version not in version_found:
-        raise DSSPError('Vermouth currently only supports DSSP version 3.0.0.')
+    supported_versions = [2, 3]
```
pckroon, line 198: I think this should be a global immutable constant, e.g. `SUPPORTED_DSSP_VERSIONS = (2, 3)` at the top of the file.

vermouth/dssp/dssp.py
```diff
@@ -194,10 +195,14 @@ def run_dssp(system, executable='dssp', savefile=None, defer_writing=True, versi
...
+    process = subprocess.run([executable, "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    match = re.search('\d+\.\d+\.\d+', process.stdout.decode('UTF8'))
+    version_found = match[0] if match else None
+    if not version_found:
+        raise DSSPError('Failed to get DSSP version information.')
+    if not int(version_found.split('.')[0]) in supported_versions and version != version_found:
+        raise DSSPError(f'DSSP {version_found} is not supported.')
```
pckroon, line 205: I have the feeling using something like https://packaging.pypa.io/en/latest/version.html#packaging.version.parse will result in simpler code. If you feel like over-engineering this, you can consider allowing more specific versions in SUPPORTED_DSSP_VERSIONS, but this is completely optional.

I'm also a bit confused on the meaning of `version` vs `supported_versions`, especially when I read the docstring. It seems they both cover the same use case. I'm not sure what the intended use was for the `version` argument, nor what it adds at the moment. It would be fine with me to remove the argument --- it's not used anywhere anyway. @fgrunewald You added this argument back in the day, is my interpretation correct?

pckroon (changes requested):
Many many thanks!
I'm rather surprised the DSSP output parsing can remain the same for both DSSP 2 and 3! 
I have some small comments about implementation details, and I think the `version` argument for `run_dssp` can be removed, but I'll give @fgrunewald a little bit of time to veto that.

vermouth/dssp/dssp.py
juminlee: I kept `version` for the case that you want to support DSSP 3 only as a default.
In that case, `supported_versions` will contain only '3', and DSSP 2 can be supported by passing specific version with `version` argument.

fgrunewald: We implemented the version check as a quick fix, so I'm not surprised about the bug. In my opinion the executable thing is the most important to fix. 

We actually had a longer - though recorded - discussion about the version check. At the moment, we can only read DSSP 3 format correctly, I believe. Even worse, we only check and also in the CI test against DSSP 3.0.0 specifically. That's why we specifically search for this version string in the output instead of something fancier. Here we decided simpler would be better. We cannot know which versions are installed by users and a wrong dssp parsing is something, which is not easily caught. However, if it goes wrong it is very very bad for the protein model v3 and even more so v2. Also there are some issues with the version printing in some DSSP exe outputs, which is why the string search seemed safer.

As far as I see it, we have two options: 1) Simply fix the bug keep the version check. 2) Fix the version check but then also implement tests for all supported dssp versions incl. minor versions or specify minor versions as well. For sure we need to have tests then. 

The important question is: @juminlee what motivated you in the first place to change the version check? 

Also integration tests are failing, which indicates something goes wrong.

fgrunewald: Update: Locally the CI with this PR works for me. 🤔

fgrunewald: ok that's what I suspected. Can you run the integration tests using the following steps:

- if not installed, install pytest 
- in the GitHub repro go to vermouth-martinize/vermouth/tests/integration_tests
- run `pytest test_integration.py`

If it runs through, it probably means we are good parsing DSSP 2.2.1 files.

juminlee: DSSP 2.2.1 passed all the tests.
[test_vermouth.log](https://github.com/marrink-lab/vermouth-martinize/files/10875121/test_vermouth.log)

juminlee: So, what do we want to do?

pckroon: Unfortunately developing and maintaining martinize2 is no longer part of my day job :'( so I may not always respond as quickly as I would like.

I partly agree with @fgrunewald here, we /do/ need tests, but IMO it would suffice to test 1) the latest 2.x version; 2) the latest 3.x version; and 3) whatever version the CI uses (which is the version distributed with Ubuntu 20.04).
I don't think it's either feasible nor desirable to test much more. Would it be an option to generate the dssp output for those versions of dssp on a not too simple protein, and check whether the parsing works/produces the expected output?

The CI failure is *very* disturbing, it looks like there's missing backbone bonds? If it doesn't go away with the next commit one of us will have to dig into that, my guess would be a change in scipy's kdtree.

vermouth/dssp/dssp.py
```diff
@@ -461,13 +468,14 @@ def convert_dssp_annotation_to_martini(
...
-    def __init__(self, executable='dssp', savedir=None):
+    def __init__(self, executable='dssp', savedir=None, version='3.0.0'):
         super().__init__()
         self.executable = executable
         self.savedir = savedir
+        self.version = version
 
     def run_molecule(self, molecule):
-        annotate_dssp(molecule, self.executable, self.savedir)
+        annotate_dssp(molecule, self.executable, self.savedir, self.version)
```
fgrunewald, line 478: this line is incorrect, because self.versoin is not the 4th positional argument. This should be replaced by:

```
annotate_dssp(molecule, self.executable, self.savedir, 'secstruct', self.version)
```

Otherwise the dssp string doesn't get annotated correctly, which causes the CI fail. Ohh boy am I happy we have those tests.

vermouth/dssp/dssp.py
fgrunewald: Adding onto this discussion: I agree with @pckroon that we can have some simple tests, which only read the dssp output from two major versions. Instead of one big protein, however, I suggest we collect a bunch of fragments that cover the existing sec. structure assignments by dssp. 

@juminlee The fact that your CI tests worked, means that your executable is not linked against the modified vermouth-library as defined in this PR. I had the same problem initially. You need to link against the editable files using `pip install -e ./`. I'm not sure what the conda equivalent is. Can you try and see if you can reproduce, the failing CI with the current PR.

vermouth/dssp/dssp.py
pckroon: I'm happy you found this :)

vermouth/dssp/dssp.py
pckroon: Ok, executive decision:
1) We will extend dssp support to the latest bugfix releases of versions 2 and 3.
2) We will test the parsing of the output files of those versions
3) We will issue *warnings* for any other version of dssp people will try to use, rather than simply refusing to work altogether.
@juminlee Is this something you can work with?

juminlee: Are you suggesting to have a specific versions in `supported_versions`?
I can make it to contain ("2.2.1", "3.0.0"), but I do not have time to test other versions.

vermouth/dssp/dssp.py
juminlee: Thank you! Fixed in commit 8793110397c4f5bbc84e5dbab7a447a7be58abca.

vermouth/dssp/dssp.py
juminlee: For #3, I can change following line:
`raise DSSPError(f'DSSP {version_found} is not supported.')`

to:
`LOGGER.warning("Vermouth is tested only with DSSP versions {}. "`
`               "The specified DSSP version {} may cause unexpected results.",`
`               SUPPORTED_DSSP_VERSIONS, version_found,`
`               type='DSSP-version')`

However, does this mean that the `version` argument is no longer needed?

juminlee: I updated the code to issue a warning instead of raising an error 77c6a9933fd1d3b9e106ecae19bcd8d855ac9804.
In addition, I removed all version arguments.

> For point 2, it would entail running DSSP (verions 2.2.1 and 3.0.0) on a carefully selected PDB file, and recording the output. Then we can add a testcase that runs that output through read_dssp2, and making sure the secondary structure string is what we expect it to be.
The biggest issue would be getting an appropriate pdb file.

It would be great if someone could take over this.

pckroon: @fgrunewald It's been way too long since I touched a protein, can you (or someone in the lab) create an appropriate PDB file for me to play around with?
I think I prefer the idea of a crystal structure with a RCSB/PDB accession ID, that will make regenerating the whole thing easier later on, since that means we won't have to store the PDB file.

I can turn that into a test case or 2.

pckroon:
@juminlee Many thanks for your help in finding and fixing this :)
With a bit of help from @fgrunewald (if he gets me a pdb file) I can take over making tests for this.

fgrunewald:
@pckroon I'll dig out some files. We should also make sure to fix #364 or at least consider it again.

pckroon:
Excellent plan. I hope I can take a look at it later this week, but I won't make any promises...

vermouth/dssp/dssp.py
```diff
@@ -194,10 +194,17 @@ def run_dssp(system, executable='dssp', savefile=None, defer_writing=True, versi
...
-    if version not in version_found:
-        raise DSSPError('Vermouth currently only supports DSSP version 3.0.0.')
+    process = subprocess.run([executable, "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    match = re.search('\d+\.\d+\.\d+', process.stdout.decode('UTF8'))
+    version = match[0] if match else None
+    if not version:
+        raise DSSPError('Failed to get DSSP version information.')
+    if not version in SUPPORTED_DSSP_VERSIONS:
+        LOGGER.warning("Vermouth is tested only with DSSP versions {}. "
+                       "The specified DSSP version {} may result inaccurate "
```
fgrunewald, line 204: `result in inaccurate`

fgrunewald: perhaps add that one can manually give a ss string using the ss flag?

fgrunewald (dismissed):
Looks good. One nitpick and a suggestion

fgrunewald (approved):

PR #516, opened 2023-04-25
add boxes to parsers

fgrunewald:
For multiple reasons it would be very useful, when the coordinate parsers would also store the box. I've added this parsing feature to the gro / pdb parser in the least invasive way. The gro solution is not so neat but then again the parser is also entirely out of line with the rest. 

For the PDB parser I stored the box in the parser object itself. In my opinion this would be safer and then we can decide when and if to use it with molecules. Unfortunately this doesn't work for the gro file reader so it then becomes a molecule class variable. I'm open to other suggestions, but would rather have it fast than redoing any major part of the parsers

### DIFF
--- a/vermouth/gmx/gro.py
+++ b/vermouth/gmx/gro.py
@@ -107,6 +107,11 @@ def read_gro(file_name, exclude=('SOL',), ignh=False):
 
             molecule.add_node(idx, **properties)
             idx += 1
+        box = np.array(line.strip().split(), dtype=float)
+        # not pretty but the parser is out of touch with the rest of the parsing
+        # infrastructure already
+        molecule.box = box
+
     return molecule
 
 
--- a/vermouth/pdb/pdb.py
+++ b/vermouth/pdb/pdb.py
@@ -279,13 +279,12 @@ class PDBParser(LineParser):
             ('beta', float, 7),
             ('gamma', float, 7),
             ('space_group', str, 11),
+            ('', str, 1),
             ('z_value', int, 4),
             ]
         start = 0
         field_slices = []
         for name, type_, width in fields:
-            if name == "space_group":
-                start+=1
             if name:
                 field_slices.append((name, type_, slice(start, start + width)))
             start += width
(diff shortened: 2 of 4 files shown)


vermouth/pdb/pdb.py
```diff
@@ -265,6 +265,38 @@ def _atom(self, line, lineno=0):
...
+            ('beta', float, 7),
+            ('gamma', float, 7),
+            ('space_group', str, 11),
+            ('z_value', int, 4),
+            ]
+        start = 0
+        field_slices = []
+        for name, type_, width in fields:
+            if name == "space_group":
+                start+=1
```
pckroon, line 288: lint.
Also, why this +1? If there's space in the format, take a look at how I solved it with the ATOM records. IIRC I put a negative width

vermouth/gmx/gro.py
```diff
@@ -107,6 +107,11 @@ def read_gro(file_name, exclude=('SOL',), ignh=False):
 
             molecule.add_node(idx, **properties)
             idx += 1
+        box = np.array(line.strip().split(), dtype=float)
+        # not pretty but the parser is out of touch with the rest of the parsing
+        # infrastructure already
+        molecule.box = box
```
pckroon, line 113: What's the problem with this?

pckroon (changes requested):
I like it. Tiny bit of lint and a small question/comment.
Also, could you hack these back into the writers as well?

vermouth/pdb/pdb.py
fgrunewald: yeah there is a gap in the fields between space_group and the next item.

vermouth/gmx/gro.py
fgrunewald: Ideally I'd not like to add "secret" class variables to molecule. Makes it a bit hard to know when this attribute is set and what possible values are. Having said that the only other option would be to put it into the meta dict. Both things are possible. Also box should ideally propagate to system but that's perhaps something for when we deal with coordinate building etc.

fgrunewald: well we can go either way; if you like it we can also just add it to molecule and put a doctoring with it

pckroon:
Keep it as is, and please open an issue complaining about the gro parser not returning a System; and to add something box-like to System

PR #520, opened 2023-05-15
do not sort when Go

fgrunewald:
Do not sort atoms when a Go model is requested. This is for sure a quick patch but seeing that the Go model makes pretty strict assumptions about what order beads and Go sites can be in, I think it will be not too much of a problem.


pckroon (approved):
LGTM. Why does pylint fail?

fgrunewald:
@pckroon I think because we now have 8.999 on pylint and somehow the added code brings it down. Note exactly sure why though. to me the added lines are lint conform

Issue #348, opened 2021-02-26
Mis handling of some atlernate residues


jbarnoud:
Some PDB specify alternate conformation by specifying one conformation without the alternate indicator and the following ones with the indicator. An example of this is 1amh:

```
ATOM   1230  N   GLY A 184       2.557  -3.590  23.433  1.00 15.72           N
ATOM   1231  CA  GLY A 184       3.626  -3.224  24.374  1.00 15.63           C
ATOM   1232  C   GLY A 184       3.345  -3.716  25.795  1.00 14.94           C
ATOM   1233  O   GLY A 184       2.536  -4.633  26.006  1.00 17.82           O
ATOM   1234  N   PHE A 184A      4.054  -3.065  26.704  1.00 15.27           N
ATOM   1235  CA  PHE A 184A      4.000  -3.345  28.144  1.00 17.29           C
ATOM   1236  C   PHE A 184A      3.605  -2.082  28.915  1.00 19.37           C
ATOM   1237  O   PHE A 184A      3.770  -0.956  28.424  1.00 18.07           O
ATOM   1238  CB  PHE A 184A      5.376  -3.804  28.631  1.00 17.67           C
ATOM   1239  CG  PHE A 184A      5.977  -4.906  27.758  1.00 15.44           C
ATOM   1240  CD1 PHE A 184A      6.772  -4.573  26.654  1.00 14.38           C
ATOM   1241  CD2 PHE A 184A      5.731  -6.247  28.066  1.00 11.27           C
ATOM   1242  CE1 PHE A 184A      7.320  -5.585  25.857  1.00 13.12           C
ATOM   1243  CE2 PHE A 184A      6.279  -7.260  27.269  1.00 10.40           C
ATOM   1244  CZ  PHE A 184A      7.073  -6.929  26.164  1.00 13.43           C
```

Because we use the atoms that do not have an alternate indicator OR that have it set to `A`, the example above takes both the conformations which results in DSSP failing:

```
raise DSSPError(message.format(err=process.stderr, file=tmpfile_name))
vermouth.dssp.dssp.DSSPError: DSSP encountered an error. The message was DSSP could not be created due to an error:
inconsistent residue types in atom records for residue 164 (PHE != GLY)
```

There are different ways of solving the issue:
* when reading an atom with the alternate indicator set to `A`, check that we did not already read a residue without an indicator, if so drop the atom
* after reading the PDB, check the residue graph for redundant residues
[...]

PR #524, opened 2023-06-20
Fix a typo in pdb.py that causes DSSP to fail

juminlee:
There is a typo in pdb.py which causes DSSP to fail. #348


pckroon:
Alright! CI had a bit of a hickup yesterday, but nothing a restart didn't fix. I also notice we accumulated some lint in the code, but that's someone else's problem.

Could I still ask you to add an insertion_code attribute to one of the nodes in `vermouth/tests/pdb/test_write_pdb.py:dummy_system`, and modify the expected outputs for `test_conect` and `test_write_success_missing_pos` in the same file? That way it can't go wrong again :)

pckroon (approved):
Awesome, thank you so much!

Issue #525, opened 2023-06-21
redundant link in martini3001


fgrunewald:
This link is never applied because it get's overwritten by another link. This is the intended behavior and just a remainder of development. 
https://github.com/marrink-lab/vermouth-martinize/blob/b450b6793e85e4f31542d9232b79a12c229a43ec/vermouth/data/force_fields/martini3001/aminoacids.ff#L1091-L1099

pckroon:
I don't see which one overwrites it?

fgrunewald:
yep really hard to see. Took me like 1h to figure it out. But this one:


https://github.com/marrink-lab/vermouth-martinize/blob/b450b6793e85e4f31542d9232b79a12c229a43ec/vermouth/data/force_fields/martini3001/aminoacids.ff#L1128-L1137

fgrunewald:
yes but the bottom patterns include all above patterns the order in the ff file is incorrect

pckroon:
I don't think it does?
`-BB BB {"cgsecstruct": "C", "resname": "PRO|HYP"} +BB` <!= 
`-BB BB {"cgsecstruct": "C", "resname": $protein_resnames_non_pro} +BB`
Note that it says `protein_resnames_non_pro`, so anything but PRO and HYP.

fgrunewald:
Let's say we have the sequence `ALA-PRO-ALA`. What we expect is to have an angle `127 25` due to the first link. That matches because we have a Proline in the middle. But this match gets overwritten by the second pattern of the other link (i.e. `-BB {"cgsecstruct": "C", "resname": $protein_resnames_non_pro} BB +BB  `) This pattern only requires the first BB bead to be non-proline, which is true. Thus we get `127 20`. This is true for ALL cases. Because we never noticed, all itp files ever produced did not apply the first link, which therefore makes it redundant.

pckroon:
Ooooooh you're right. Good catch :)

fgrunewald:
@pckroon do we remove it or leave it?

pckroon:
Remove it, since it slows down the FF

PR #526, opened 2023-06-26
Update aminoacids.ff

fgrunewald:
remove redundant link in ff martini3001

### DIFF
--- a/vermouth/data/force_fields/martini3001/aminoacids.ff
+++ b/vermouth/data/force_fields/martini3001/aminoacids.ff
@@ -1091,15 +1091,6 @@ resname $protein_resnames
 [ link ]
 [ angles ]
 ; (Modified by Paulo)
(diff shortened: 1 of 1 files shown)


fgrunewald:
@pckroon what about this lint stuff?

pckroon:
The lint is an issue. I would like to improve it, but I don't have the time. If you *do* have the time I would be very grateful. Otherwise I'm seriously considering running `black` to take care of it all.

pckroon:
Or we stop running pylint. But that sounds wrong.

pckroon (approved):

fgrunewald:
@pckroon I guess we keep it as implicit default then?

pckroon:
I guess so. Does it mean issue #525 is invalid?

fgrunewald:
@pckroon I the issue an discussion. If I understand the people involved correctly this is not the intended use-case for this link, but now it is.

PR #529, opened 2023-06-29
add modifications for all protonation states

fgrunewald:
In this PR are all missing modifications for the different common protonation states in the context of the Martini3 FF.

Modifications
-------------------------
ASP -> ASP-HD1 
ASP -> ASP-HD2 
GLU -> GLU-HE1
GLU -> GLU-HE2
LYS  -> LSN (deprotonated Lysine)
 
Amber and Charmm named histidines, in this case the resname remains HIS
HIS -> HD
HIS -> HP
HIS -> HE

Issues
----------
- the Lysine modification does not get recognized. I think it has to do with the common issue we encountered before but can you have a look and see if I made a typo
- when using for example elnedyn with the current set of files on a protonated ASP for example, a warning is raised that the modification is unknown. One still get's an itp just with the incorrect protonation state as was the case with martinize1. 

To Do
--------
- [x]  fix elnedyn integration test for ASP modification
- [x] add charmm histidine parameters
- [x] add amber GLU modifications
- [x] add amber ASP modifications
- [x] add amber LYS modifications
- [ ] fix HIS for gromos
- [x] fix LYS for modifications
- [ ] clean up mapping directory
- [ ] add all modifications for all martini FFs


pckroon (dismissed):
Looks good to me, thanks. I didn't check the mappings with a fine-toothed comb, but I assume you did.

> the Lysine modification does not get recognized. I think it has to do with the common issue we encountered before but can you have a look and see if I made a typo

No typos, working as intended. It's hard/impossible to automatically detect modifications that only remove atoms, since then it becomes unclear whether these atoms were simply not resolved in the X-ray, or if it's really a deprotonated amino acid.

> when using for example elnedyn with the current set of files on a protonated ASP for example, a warning is raised that the modification is unknown. One still get's an itp just with the incorrect protonation state as was the case with martinize1.

Working as intended then.

fgrunewald:
@pckroon thanks for the review! 

About Lysine: We could do the same as I did now for Histidine. We remove all hydrogen from the NH3 of Lysine for the base mapping and treat the two protonation states as possible modifications. The only minor disadvantage is that you get the modification notification in case all hydrogen are present. If no hydrogen are present or not resolved you will get the default charged Lysine.

pckroon:
Hmmmn, I'm not sure I like that, for the HIS as well.
How common is it to end up with a PDB without /any/ hydrogens? In those cases you don't want to *quietly* end up running a simulation with the wrong protonation, so IMHO the default protonation state should always be at physiological pH.
You could argue you could add a warning to the base (unprotonated) blocks, but those would also trigger after the modification gets applied, so there'd be no way to silence them (at least not without a code change).

So my question really is, which case would result in the least amount of quietly wrong simulations?

fgrunewald:
> Hmmmn, I'm not sure I like that, for the HIS as well. How common is it to end up with a PDB without /any/ hydrogens? 
In this case you end up with the default HIS / LYS anyways. The modification only applies if you are exactly having the correct amounts of hydrogen corresponding to your protonation state in which case I'd argue that is the correct default.

pckroon:
So my point is that the default HIS/LYS according to our data files (=deprotonated) is not the HIS/LYS you want by default (=physiological).
I'm not arguing the modifications work as intended. I´m arguing that if your PDB/X-ray is iffy you end up with a wrong model because we didn't issue a warning

fgrunewald:
Yes I understand but that is also not the case. The default data file will and is the one corresponding to the physiological pH. BUT the mapping file does not define hydrogen, which allows to have a modification that says IF hydrogen are present corresponding to a known protonation state that state is used. That's the whole point of the Custom Protonation state section.

pckroon:
Excellent. In that case I misunderstood.

fgrunewald:
@pckroon we need to priorities this next week. Apparently the paper is already out for review. I missed it because the email notifications are broken or just terrible.

pckroon:
Alright! What do you need from me?

fgrunewald:
We need to fix the integration tests. Essentially, some modifications are not recognized for the Martini 2 based tests. I suggest to simply switch them to Martini3 for now. Unless there is a better way.

pckroon:
Ok. If you implement the last modifications and cleanup (the todo list on top) I'll see if I can find why this run of tests failed.

pckroon:
Ok, I'm back on track. It's rather annoying I can't run the integration tests locally since I can't for the live of me get DSSP to run. I'm looking at `tier-0-mini-protein2_helix` because that one doesn't need dssp.

Anyway, the integration tests for martini2 fail since we don't want to invest the time to add those mapping files. You already knew, but I finally caught up :p My suggestion is to add (e.g.) `-maxwarn unmapped-atom:3` to the afflicted tests. I'm not sure why, but atom coordinates are  also affected. (
```
E         -                'position': approx([0.4283 ± 1.0e-03, 0.2895 ± 1.0e-03, -2.3345 ± 1.0e-03]),
E         +                'position': array([ 0.4222,  0.2876, -2.3258]),
```
)
Martini2 (or elnedyn) mappings are not affected, so I don't know why those coordinates changed; do you have any idea?

vermouth/data/mappings/elnedyn/modifications.elnedyn22.mapping
```diff
@@ -104,26 +104,61 @@ CA  BB
...
-CG SC1
-OD1 SC1
-OD2 SC1
-HD2 SC1
+CG   SC1
+OD1  SC1
+OD2  SC1
+HD1  SC1
+CB   SC1
+CG   SC1
```
pckroon, line 164: Also here

vermouth/tests/integration_tests/test_integration.py
```diff
@@ -125,23 +125,23 @@ def _interaction_equal(interaction1, interaction2):
...
+   ['tier-1', 'villin'],
+   ['tier-1', '3i40'],
+   ['tier-1', '6LFO_gap'],
+   ['tier-1', '1mj5'],
+   ['tier-1', '1mj5-charmm'],
+   ['tier-1', 'EN_chain'],
+   ['tier-1', 'EN_region'],
+#   ['tier-2', 'barnase_barstar'],
+#   ['tier-2', 'dna'],
+#   ['tier-2', 'gpa_dimer'],
```
pckroon, line 144: The fourth indentation space got lost here

pckroon (changes requested):
Looks mostly good to me. There's still a vmdscene.ppm file which probably shouldn't be there.

pckroon (approved):

PR #532, opened 2023-07-26
small lint

fgrunewald:


pckroon (approved):
I'm guessing you ran an autoformatter like black? I personally don't really agree with all changes, but not enough to request changes.

Note that I didn't go through /everything/ in detail

fgrunewald:
@pckroon indeed you're right. I just wanted to see what black would score in terms of the real lint. This PR is a test-ballon but at least in medium time fixes the linting issue.

fgrunewald:
@pckroon so we merge?

PR #537, opened 2023-08-17
add box attribute to molecule and propagate from pdb parser

fgrunewald:
small fix to reading boxes. The PDB reader did not propagate the box attribute

### DIFF
--- a/vermouth/molecule.py
+++ b/vermouth/molecule.py
@@ -360,6 +360,8 @@ class Molecule(nx.Graph):
         # {loglevel: {entry: [fmt_args]}}
         self.log_entries = defaultdict(lambda: defaultdict(list))
         self.max_node = None
+        # box of the system the molecule is in
+        self.box = None
 
     def __eq__(self, other):
         return (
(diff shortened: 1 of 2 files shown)


pckroon (changes requested):
Could you add a test so this doesn't happen again?

fgrunewald:
@pckroon jup will do

pckroon:
Alright, stupid question. We probably already covered this with the original implementation, but is there any reason why this is a molecule attribute and not a system?
Looks fine otherwise

fgrunewald:
@pckroon indeed we discussed a little. In essence, adding it to the molecule keeps the API a little more consistent. For example, the `read_gro` function only returns a molecule. So there is no way to propagate the box except for restructuring this parser. Personally, I also prefer molecules because polyply works with molecules but not with vermouth systems. In the long run, could be smart to refactor `read_gro` and expose the attribute to both molecule and system. For now I think the current implementation is fine.

pckroon (approved):

PR #542, opened 2023-09-01
quick fix doc

fgrunewald:
quick fix on the doc-string. Apparently it has nothing to do with versions of netwrokx or sphinx. I tested exectly the same state of packages as when the tests worked and now the fail. Might have to do with some ubuntu setup, I don't know. Bumping to nx version 3 doesn't help either and also would require to change all downstream packages. It works but I think it is too early for that. Hence here a hopefully working quick fix.


pckroon (approved):
Alright, this works for me. Not sure if the sphinx linking will work like this, but I don't really care for this one I think.

I'll have a look at bumping the NX version somewhere soon-ish.

PR #543, opened 2023-09-15
Networkx 3

pckroon:
Update the requirements to allow networkx 3 and higher.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ setup-requires =
 install-requires =  # ?? requires-dist?
     pbr
     numpy
-    networkx ~= 2.0
+    networkx ~= 3.0
     scipy
 zip-safe = False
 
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -33,7 +33,7 @@ setup-requires =
 install-requires =  # ?? requires-dist?
     pbr
     numpy
-    networkx ~= 2.0
+    networkx ~= 3.0
```
fgrunewald, line 36: seeing that there is only one test, which was changed, would it be an option to also allow networkx v2 and raise a deprecation warning for some time until we cement the change? I'm a bit worried about the projects using martinize2 such as the MAD and Go script etc.

pckroon: How about `>= 2.0` and no further warnings or checks?

fgrunewald: sounds good

fgrunewald (approved):

PR #549, opened 2023-10-05
Make sure DATA_PATH is always a pathlib.Path

pckroon:
Small followup to #546, which introduced a small issue

### DIFF
--- a/vermouth/__init__.py
+++ b/vermouth/__init__.py
@@ -37,9 +37,9 @@ try:
     import atexit
     from contextlib import ExitStack
 except ImportError:
-    import os
-    DATA_PATH = os.path.join(os.path.dirname(__file__), 'data')
-    del os
+    from pathlib import Path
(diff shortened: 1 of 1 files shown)


vermouth/__init__.py
```diff
@@ -37,9 +37,9 @@
     import atexit
     from contextlib import ExitStack
 except ImportError:
-    import os
-    DATA_PATH = os.path.join(os.path.dirname(__file__), 'data')
-    del os
+    from pathlib import Path
+    DATA_PATH = Path(__file__).parent / 'data'
```
fgrunewald, line 41: Pathlib works os independent right?

fgrunewald (approved):

vermouth/__init__.py
pckroon: Yes, that's kind of the whole point. `os.path` does as well, but that API is less nice since you need to do `os.path.join` all the time rather than inlining a `/`

Issue #553, opened 2023-10-17
Raise warning if mutation not found


csbrasnett:
At the moment, if a -mutate flag is given, but the target residue is not found, then martinize will process the input structure as given without raising a warning.

Eg, taking the [martinize tutorial](http://cgmartini.nl/index.php/2021-martini-online-workshop/tutorials/564-2-proteins-basic-and-martinize-2):

`martinize2 -f 181L_clean.pdb -o t4l_only.top -x t4l_cg.pdb -dssp /usr/bin/dssp -p backbone -ff martini3001 -mutate A-MET1:LYS`

will correctly mutate the first residue from MET to LYS, but 

`martinize2 -f 181L_clean.pdb -o t4l_only.top -x t4l_cg.pdb -dssp /usr/bin/dssp -p backbone -ff martini3001 -mutate A-MET2:LYS`

will do nothing (ie. keep residue 1, 2 as MET, ASN as in the input structure). If a user is expecting the mutation (eg the above could easily be a typo) and uses the topology naively (even though they should really check...) then that's a problem. 

It's probably about time I put in a PR here, so I can try and do something about it this week, just thought I'd open the issue first, or in case I don't get to it as soon as I'd like.

vermouth 0.9.6

pckroon:
I think this would be a great first PR :)
The best place is probably in `vermouth/processors/annotate_mut_mod.py`. Beyond that, good luck, have fun, and feel free to ask for help/advice/whatever!

csbrasnett:
Noticed one minor issue with how this was fixed. In the case where we have >1 chain in the protein, the mutation is looked for across all chains and raises a warning for each of the chains where the mutation isn't found. Then nothing is written, even though the mutation has been specified correctly. I'll come up with a fix asap...

pckroon:
Good find. I don't think the issue is chains per se, but rather molecules, since the `AnnotateMutMod` processor has a `run_molecule` method rather than a `run_system` method.

csbrasnett:
Thanks, not sure I have time to address this week but I'll try to start next. It would be good to try and address #512  at the same time, but I'm less sure about where to start on that one.

csbrasnett:
ha, understood! I'll address this soon then, and maybe we could work something out for the rebuilding issue in the new year?

pckroon:
IMNSHO the rebuilding is one of the biggest missing features we have, so yes please!

csbrasnett:
[...]

PR #556, opened 2023-10-20
add warning when mod not found

csbrasnett:
Addresses #553

### DIFF
--- a/vermouth/processors/annotate_mut_mod.py
+++ b/vermouth/processors/annotate_mut_mod.py
@@ -210,10 +210,12 @@ def annotate_modifications(molecule, modifications, mutations):
                     (mutations, 'mutation', molecule.force_field.blocks)]
 
     residue_graph = make_residue_graph(molecule)
-    for res_idx in residue_graph:
-        for mutmod, key, library in associations:
-            for resspec, mod in mutmod:
+    for mutmod, key, library in associations:
+        for resspec, mod in mutmod:
+            mod_found = False
+            for res_idx in residue_graph:
                 if residue_matches(resspec, residue_graph, res_idx):
+                    mod_found = True
                     if mod != 'none' and mod not in library:
                         raise NameError('{} is not known as a {} for '
                                         'force field {}'
@@ -223,7 +225,11 @@ def annotate_modifications(molecule, modifications, mutations):
                                  _format_resname(res), key, mod)
                     for node_idx in res['graph']:
                         molecule.nodes[node_idx][key] = molecule.nodes[node_idx].get(key, []) + [mod]
-
+            if mod_found == False:
+                LOGGER.warning('{} with resid {} not found. '
+                                 'No modification made.'
+                                 ''.format(resspec['resname'], resspec['resid']))
+    
 
 class AnnotateMutMod(Processor):
     """


csbrasnett:
Thank you Peter! And thanks to @jan-stevens for help with fixing the tests :)

vermouth/processors/annotate_mut_mod.py
```diff
@@ -223,7 +225,11 @@ def annotate_modifications(molecule, modifications, mutations):
                                  _format_resname(res), key, mod)
                     for node_idx in res['graph']:
                         molecule.nodes[node_idx][key] = molecule.nodes[node_idx].get(key, []) + [mod]
-
+            if mod_found == False:
+                LOGGER.warning('{} with resid {} not found. '
+                                 'No modification made.'
+                                 ''.format(resspec['resname'], resspec['resid']))
```
pckroon, line 231: Small problem actually. Not all resspecs need to have a resname or resid attribute. Might be better to format the respec like the CLI accepts it. You should be able to use `_format_resname` for that. Also that function has a bad bad name. Pretty sure that's my fault though.

vermouth/tests/test_annotate_mut_mod.py
```diff
@@ -17,6 +17,8 @@
 
 import networkx as nx
 import pytest
+from contextlib import nullcontext as does_not_raise
```
pckroon, line 20: This is not actually used, right?

vermouth/tests/test_annotate_mut_mod.py
```diff
@@ -289,7 +291,8 @@ def test_nter_cter_modifications(node_data, edge_data, expected):
     mol = Molecule(force_field=ForceField(FF_UNIVERSAL_TEST))
     mol.add_nodes_from(enumerate(node_data))
     mol.add_edges_from(edge_data)
-    modification = [({'resname': 'cter'}, 'C-ter'), ({'resname': 'nter'}, 'N-ter')]
+    modification = [({'resname': 'cter', 'resid': 3}, 'C-ter'), 
+                    ({'resname': 'nter', 'resid': 1}, 'N-ter')]
```
pckroon, line 295: Right, as in the comment before, not all resspecs need a resid or resname. These for example don't need a resid; the 'nter' and 'cter' already have enough information to not need the resid or sane resname.

pckroon (changes requested):

vermouth/processors/annotate_mut_mod.py
csbrasnett: Thanks for the heads up on `_format_resname`, will update using that.

vermouth/tests/test_annotate_mut_mod.py
csbrasnett: yep, will remove that and next line, used when trying to work out how to sort out the tests.

vermouth/tests/test_annotate_mut_mod.py
csbrasnett: weirdly, adding those resids seemed to be what fixed the tests before, but I just tried again with them removed and it looks like it works without? I think we thought it was overwriting the whole dictionary now, and that's why they were needed. Not sure what's happened there.

pckroon (approved):

pckroon:
This looks good to me!
The tests failed since your first warning assumed a resid (and resname) would always be available in a resspec. Since the _format_resname doesn't make that assumption, no more test failure.

PR #559, opened 2023-10-31
-el fix

csbrasnett:
Currently, if elastic network rubber bands are applied and are 1) below the cutoff for -el and 2) not otherwise eliminated by the -ermd threshold for the force field, they are kept in. Ie. the -el cutoff is not actually currently applied. This PR applies the cutoff.

### DIFF
--- a/vermouth/tests/test_apply_rubber_band.py
+++ b/vermouth/tests/test_apply_rubber_band.py
@@ -543,19 +543,31 @@ def test_bail_out_on_nan(caplog, test_molecule):
                           ([1, 2, 0, 0, 500, 400, np.array([[  0,500,500,  0],
                                                             [500,  0,500,500],
                                                             [500,500,  0,500],
-                                                            [  0,500,500,500]])], # no decays, return the base constant within the bounds
+                                                            [  0,500,500,  0]])], # no decays, return the base constant within the bounds
                            [1, 2, 0, 0, 500, 600, np.array([[  0,  0,  0,  0],
                                                             [  0,  0,  0,  0],
                                                             [  0,  0,  0,  0],
                                                             [  0,  0,  0,  0]])], # all forces less than the minumum force
                            [1, 0.5, 0, 0, 500, 600, np.array([[  0,  0,  0,  0],
-                                                            [  0,  0,  0,  0],
-                                                            [  0,  0,  0,  0],
-                                                            [  0,  0,  0,  0]])], # all distances larger than the upper bound
+                                                              [  0,  0,  0,  0],
+                                                              [  0,  0,  0,  0],
+                                                              [  0,  0,  0,  0]])], # all distances larger than the upper bound
                            [4, 2, 0, 0, 500, 600, np.array([[  0,  0,  0,  0],
                                                             [  0,  0,  0,  0],
                                                             [  0,  0,  0,  0],
                                                             [  0,  0,  0,  0]])], # all distances less than the lower bound
+                           [1, 3, 0.1, 0, 500, 400, np.array([[  0.        , 452.41870902, 452.41870902, 452.41870902],
+                                                              [452.41870902,   0.        , 452.41870902, 452.41870902],
+                                                              [452.41870902, 452.41870902,   0.        , 452.41870902],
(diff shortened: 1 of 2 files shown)


fgrunewald:
@csbrasnett do you dare to add a test for this behaviour? Editing 'tests/test_apply_rubber_bands' would be the place. Probably would make sense to add a new test for the compute_force_constants function. To get started you can ask chatGPT to write a test for example.

csbrasnett:
@fgrunewald I was worried you'd ask... Have had a go now, that'll cover the testing of the different bounds. I haven't added anything to test the decay factors/powers because that gets in to testing floats and I didn't want to get in to that.

vermouth/tests/test_apply_rubber_band.py
```diff
@@ -537,3 +537,38 @@ def test_bail_out_on_nan(caplog, test_molecule):
...
+                                [2, 1, 0, 1],
+                                [3, 2, 1, 0]])
+
+    # Call the compute_force_constants function with the constant distance_matrix and varied parameters
+    result = vermouth.processors.apply_rubber_band.compute_force_constants(distance_matrix, lower_bound, upper_bound, 
+                                                                           decay_factor, decay_power, base_constant, 
+                                                                           minimum_force)
+
+    # Assert the result against the expected output
+    assert result.all() == expected_output.all()
```
fgrunewald, line 574: If you want to test floats you can use `np.allclose(array1, array2)` together with setting the absolute tolerance for the float. Or Peter seems to prefer the pytest build in using pytest.approx.

fgrunewald (changes requested):
For extendability purposes I suggest to use np.allclose and then you could also add a test for the decay factor but totally optional here.

fgrunewald (approved):

pckroon:
I'm not sure the behaviour you propose here is desired actually. The semantics between upper and lower bounds are annoyingly different in this context. The code you propose to change has been there since #54, the initial implementation for the EN.

If this change goes through we will produce a *different* elastic network than martinize1

fgrunewald:
@csbrasnett did you compare this to martinize1? Also @pckroon I don't really care if we diverge here. The implementation in martinize1 seems just incorrect or am I missing something?

csbrasnett:
I haven't checked against martinize1, I can try and find time today. 

But actually @pckroon is correct in that if the distance is < lower then it needs to just be set as the constant, rather than being deleted as this currently behaves. Which still wasn't happening before - eg. if you applied a decay and a lower cutoff, then the forces below the cutoff would be larger than the given constant. But that's easy enough to change. I can fix these and check against martinize1 if we agree that's the expected behaviour?

csbrasnett:
Right, so this would change the behaviour against martinize1, which currently reproduces what the main branch does. However, I 1) cannot find where the concept of a _lower_ cutoff has emerged from, as in both the papers and the citation trail that follows, only the _upper_ cutoff is ever discussed.

That said, I still think that the expected behaviour is currently wrong based on the descriptions. This PR now actually only fixes the side case where decays are used. In the fig here, where ea = ep = 0, the force is constant below el for both old and new. But with decays (and with line 99 now reading `constants[distances < lower_bound] = base_constant` as it should) the force below the el is correctly set to the given force constant.


fig

pckroon (dismissed):
Excellent. Thanks for digging up the details!

csbrasnett:
Thanks! nb don't merge yet, need to add the final condition, which isn't as trivial as I thought it'd be.

fgrunewald (changes requested):
@csbrasnett let us know once you're done.

csbrasnett:
@fgrunewald now corrected. Based on the discussion and graph above, I realised the second condition wasn't actually necessary, because the deviation is constant > base_constant for distance < el, so it's simple enough to set anything larger as the base_constant.

fgrunewald (dismissed):

pckroon (approved):
I'd be very happy if you could expand the docstring of compute_force_constants now that you know exactly what it does and is supposed to do.
LGTM otherwise, nice work :)

pckroon (approved):

PR #565, opened 2023-12-14
Multiple mutate fixes

csbrasnett:
re-addresses #553 in the case that mutations are given across multiple chains. This PR makes sure that in the case that multiple chains are targeted, the check for the target residue existing is not present across all chains. So warnings aren't raised when the 'missing' residue is encountered.


vermouth/processors/annotate_mut_mod.py
```diff
@@ -210,25 +210,48 @@ def annotate_modifications(molecule, modifications, mutations):
...
+                                            ''.format(mod, key, molecule.force_field.name))
+                        res = residue_graph.nodes[res_idx]
+                        LOGGER.debug('Annotating {} with {} {}',
+                                     _format_resname(res), key, mod)
+                        for node_idx in res['graph']:
+                            molecule.nodes[node_idx][key] = molecule.nodes[node_idx].get(key, []) + [mod]
+                if mod_found == False:
+                    LOGGER.warning('Mutation "{}" not found. '
+                                   'Check target resid!'
+                                   ''.format(_format_resname(resspec)))
```
pckroon, line 238:
```suggestion
                if not mod_found:
                    LOGGER.warning('Residue specified by "{}" for mutation "{}" not found. ', _format_resname(resspec), mod)
```

vermouth/processors/annotate_mut_mod.py
```diff
@@ -210,25 +210,48 @@ def annotate_modifications(molecule, modifications, mutations):
...
+            # If instead we're targeting all residues in the chain
+            elif resspec.get(chain) == None:
```
pckroon, line 240:
```suggestion
            elif resspec.get(chain) is None:
```

vermouth/processors/annotate_mut_mod.py
```diff
@@ -210,25 +210,48 @@ def annotate_modifications(molecule, modifications, mutations):
...
+                        res = residue_graph.nodes[res_idx]
+                        LOGGER.debug('Annotating {} with {} {}',
+                                     _format_resname(res), key, mod)
+                        for node_idx in res['graph']:
+                            molecule.nodes[node_idx][key] = molecule.nodes[node_idx].get(key, []) + [mod]
+                if mod_found == False:
+                    LOGGER.warning('Mutation "{}" not found. '
+                                   'Check target resid!'
+                                   ''.format(_format_resname(resspec)))
+            # If instead we're targeting all residues in the chain
```
pckroon, line 239:
```suggestion
            # If instead we're targeting residues in any chain
```

vermouth/processors/annotate_mut_mod.py
```diff
@@ -259,3 +282,9 @@ def __init__(self, modifications=None, mutations=None):
     def run_molecule(self, molecule):
         annotate_modifications(molecule, self.modifications, self.mutations)
         return molecule
+    def run_system(self, system):
```
pckroon, line 285:
```suggestion
        return molecule
    
    def run_system(self, system):
```

vermouth/processors/annotate_mut_mod.py
```diff
@@ -259,3 +282,9 @@ def __init__(self, modifications=None, mutations=None):
...
+        mols = []
+        for molecule in system.molecules:
+            mols.append(self.run_molecule(molecule))
+        system.molecules = mols
```
pckroon, line 289: What does this add over `super().run_system()`?

pckroon (changes requested):
Thanks for picking up this issue.
I'm not very happy with how much code duplication this solution produces though...
Maybe you can abstract the duplicated functionality in a separate function (maybe that would return the affected nodes?)? As alternative, make the resspecs-to-find attributes of the AnnotateMutMod Processor, and make sure you find each at least once when running run_system.
```py
def run_system(system):
    self.resspec_counts = ...
    # Modify run_molecule to increment the resspec_counter(s)
    super().run_system()
    if any(self.resspec_counts == 0): warn
```

vermouth/processors/annotate_mut_mod.py
```diff
@@ -176,6 +176,22 @@ def _format_resname(res):
     out += res.get('insertion_code', '')
     return out
 
+def _resiter(mod, residue_graph, resspec, library, key, molecule):
+    mod_found = False
+    for res_idx in residue_graph:
+        if residue_matches(resspec, residue_graph, res_idx):
+            mod_found = True
+            if mod != 'none' and mod not in library:
```
pckroon, line 184:
```suggestion
            if mod != 'none' and mod not in library[key]:
```
?

vermouth/processors/annotate_mut_mod.py
```diff
@@ -176,6 +176,22 @@ def _format_resname(res):
     out += res.get('insertion_code', '')
     return out
 
+def _resiter(mod, residue_graph, resspec, library, key, molecule):
```
pckroon, line 179: Needs a docstring, especially with the number of arguments it takes

pckroon (changes requested):
Nice bit of cleanup :) I do think I spotted a (small) mistake though, and there are some small comments left over from before.
I forgot why the different behaviour for whether a chain is specified or not?

pckroon:
Updating the codecov action seems to have fixed the failing test thingie

csbrasnett:
just to check, are you expecting any more changes from me?

pckroon (commented):
I think this still gives erroneous warnings for chemically disconnected things. For example, protein with 2 chains (A and B) which are not connected by a chemical bond (and are thus separate Molecules), combined with a valid resspec that defines a chain.

pckroon (changes requested):
One more testcase: I think `{'resname': 'GLY', 'resid: 1'}` will not print a warning even if it doesn't match anything.
After that I'm happy :)

csbrasnett:
So, I think those tests do what you expect them to, but I'm not sure they should work, as they don't pick up on the error identification correctly? Eg, GLU4:GLY should mutate any GLU at resid 4 on every chain, but if resid 4 is not GLU then there should be a warning, and there currently isn't. I'll have a think about how to address this.

I'll also update the martinize2 help to add some more detail about what's possible with the mutate/modify flags?

pckroon:
>  Eg, GLU4:GLY should mutate any GLU at resid 4 on every chain, but if resid 4 is not GLU then there should be a warning, and there currently isn't.

Exactly :)
I think you'll need to adapt the run_system method to track whether at the end there are any resspecs that didn't match anything

> I'll also update the martinize2 help to add some more detail about what's possible with the mutate/modify flags?

This is always welcome. Documentation is not my strong suite

pckroon:
That warning works for single molecules, not the entire system, which is what's going wrong if you have a resspec that does not mention any chains but does not match anywhere

csbrasnett:
> That warning works for single molecules, not the entire system, which is what's going wrong if you have a resspec that does not mention any chains but does not match anywhere

I'm pretty sure that possibility's covered by this test though: 

```
[
    {'chain': 'A', 'resname': 'ALA', 'resid': 1},
    {'chain': 'A', 'resname': 'ALA', 'resid': 2},
    {'chain': 'A', 'resname': 'ALA', 'resid': 3},
    {'chain': 'B', 'resname': 'ALA', 'resid': 1},
    {'chain': 'B', 'resname': 'ALA', 'resid': 2},
    {'chain': 'B', 'resname': 'ALA', 'resid': 3}
],
[(0, 1), (1, 2), (3, 4), (4, 5)],
[({'resname': 'GLY', 'resid': 1}, 'ALA')],
True
```

ie. two molecules in the system, target any GLY with resid 1 (of which there are none), and expect a warning to be raised

pckroon:
Since the test was always running on a single Molecule (rather than a System of multiple molecules) it was hiding the issue I was trying to dig up.
The problem is that  if a resspec does not get applied to /all/ molecules, a warning gets issued, even if it applies to at least 1.

csbrasnett:
thanks for that additional test, I see what you mean now about what wasn't captured. Will fix in due course!

vermouth/tests/test_annotate_mut_mod.py
```diff
@@ -394,7 +394,7 @@ def test_nter_cter_modifications(node_data, edge_data, expected):
         ],
         [(0, 1), (1, 2), (3, 4), (4, 5)],
         [({'resname': 'GLY', 'chain': 'A'}, 'ALA')],
-        True
+        False
```
pckroon, line 397: This test should raise a warning, since there is no GLY in chain A?

pckroon:
No worries. I know how these things can happen.
My suggestion is to simplify the `annotate_modifications` function again, to simply apply yes or no depending on whether the resspec was matched anywhere. Being able to warn whether a modification applied to just one or all chains is nice to have, but not necessary.  The most relevant message is when a modification doesn't apply anywhere. I'll leave it up to you whether the more detailed message is worth the added complexity.

pckroon (changes requested):

vermouth/processors/annotate_mut_mod.py
```diff
@@ -176,8 +176,41 @@ def _format_resname(res):
...
+                                ''.format(mod, key, molecule.force_field.name))
+            res = residue_graph.nodes[res_idx]
+            LOGGER.debug('Annotating {} with {} {}',
+                         _format_resname(res), key, mod)
+            for node_idx in res['graph']:
+                molecule.nodes[node_idx][key] = molecule.nodes[node_idx].get(key, []) + [mod]
+    return mod_found
 
-def annotate_modifications(molecule, modifications, mutations):
+def annotate_modifications(molecule, modifications, mutations, resspec_counts):
```
pckroon, line 213: Final Nitpick, could you add the new argument to the docstring, and emphasize there that it is modified in place?

pckroon (changes requested):
One last nitpick, LGTM otherwise!

pckroon (approved):
Nice work! Thanks for following through and dealing with all the small nitpicks!

Issue #570, opened 2024-01-25
parameterizing ATP and MG+


Nancyarvin:
Hi,
I am trying to parameterize ATP and mg+ in order to coarse grain it from charmm to martini2. Pasting down the error below:
```
Traceback (most recent call last):
  File "/home/arvin/miniconda3/envs/python3.6.3_env/bin/martinize2", line 1110, in <module>
    entry()
  File "/home/arvin/miniconda3/envs/python3.6.3_env/bin/martinize2", line 761, in entry
    Path(DATA_PATH) / "mappings", known_force_fields
  File "/home/arvin/miniconda3/envs/python3.6.3_env/lib/python3.6/site-packages/vermouth/map_input.py", line 426, in read_mapping_directory
    new_mappings = read_backmapping_file(infile, force_fields)
  File "/home/arvin/miniconda3/envs/python3.6.3_env/lib/python3.6/site-packages/vermouth/map_input.py", line 87, in read_backmapping_file
    full_mapping = _read_mapping_partial(lines, line_number + 1)
  File "/home/arvin/miniconda3/envs/python3.6.3_env/lib/python3.6/site-packages/vermouth/map_input.py", line 338, in _read_mapping_partial
    _, from_atom, *to_atoms = cleaned.split()
ValueError: not enough values to unpack (expected at least 2, got 1)
```
It would be great if someone would help me in this regard. Zipped all the required files

Regards,
Arvin

[ATP and MG.zip](https://github.com/marrink-lab/vermouth-martinize/files/14050531/ATP.and.MG.zip)

pckroon:
I have the idea this is only a partial traceback/error. Could you provide the full screen output?

pckroon:
Hmmn, I was expecting more useful screen output from martinize, but alas.

Your MG mapping looks off. The atoms section should look like:

```
[ atoms ]
MG MG
```
Indicating that you want to map the charmm atom called MG (left) to the Martini22 bead MG (right)

Nancyarvin:
I just had this much of output error on the screen. 
Here is the edited charmm36 file of MG as per your suggestion:
![edited mg_charmm](https://github.com/marrink-lab/vermouth-martinize/assets/77006462/42499b2e-8123-42d5-aed0-1ee1fed973e3)
It is giving error:
![error2](https://github.com/marrink-lab/vermouth-martinize/assets/77006462/2abc99b7-0932-43f7-a676-3c8d45a3071c)

[...]

PR #571, opened 2024-01-29
Log a helpful message when an element is not present in the ATOM_MASS dict

pckroon:
… and then use a relatively high mass since it's probably a metal.

Side-effect of #570. I currently set the loglevel to INFO, since I don't think it usually results in a faulty structure or topology.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,0 @@
-[run]
-branch = True
-omit = 
-    vermouth/tests/*
-    vermouth/redistributed/*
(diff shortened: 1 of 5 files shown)


fgrunewald (approved):
Why default mass 30? And shouldn‘t we just add the correct masses for the main groups?

pckroon:
I picked 30 mostly because "Why not".
And yes, it's probably a good idea to at least add all single-letter elements (since those we can detect correctly). Even better would be to add the whole PTE, but for that we also need to fix element derivation from atomnames.

PR #576, opened 2024-02-28
SEQRES parsing

csbrasnett:
First go at adding parsing of SEQRES section of PDB files to check the residues present with coordinates in the pdb vs. the SEQRES entry. The check will also do basic sequence alignment to try and advise which residues are missing, and raise a warning to say which ones are.

### DIFF
--- a/vermouth/pdb/nwalign.py
+++ b/vermouth/pdb/nwalign.py
@@ -1,0 +1,238 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 University of Groningen
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+'''
+
+Sequence alignment with the Needleman-Wunsch algorithm
+
+Designed to check the SEQRES entry of an input PDB versus the residues present
+
+Adapted from
+https://github.com/zaneveld/full_spectrum_bioinformatics/blob/master/content/08_phylogenetic_trees/needleman_wunsch_alignment.ipynb
+
+'''
+import numpy as np
+from ..log_helpers import StyleAdapter, get_logger
+
+LOGGER = StyleAdapter(get_logger(__name__))
+
+ONE_LETTER_AA = {"G": "GLY",
+                 "A": "ALA",
+                 "V": "VAL",
+                 "C": "CYS",
+                 "P": "PRO",
+                 "L": "LEU",
+                 "I": "ILE",
+                 "M": "MET",
+                 "W": "TRP",
+                 "F": "PHE",
+                 "S": "SER",
+                 "T": "THR",
+                 "Y": "TYR",
+                 "N": "ASN",
+                 "Q": "GLN",
+                 "K": "LYS",
+                 "R": "ARG",
+                 "H": "HIS",
+                 "D": "ASP",
+                 "E": "GLU",
+                 }
+
+THREE_LETTER_AA = {ONE_LETTER_AA[key]: key for key in ONE_LETTER_AA.keys()}
+
+def OLA_codes(res_list):
+    '''
+    convert continuous string of three letter AA codes into list of single letters
+    '''
+    n = 3
+    return ''.join([THREE_LETTER_AA[i] for i in res_list])
+
+def traceback_alignment(traceback_array, seq1, seq2, up_arrow= 1,
+                        left_arrow=2, up_left_arrow=3, stop=0):
+    """Align seq1 and seq2 using the traceback matrix and return as two strings
+
+    traceback_array -- a numpy array with arrow characters indicating the direction from
+    which the best path to a given alignment position originated
+
+    seq1 - a sequence represented as a string
+    seq2 - a sequence represented as a string
+    up_arrow - the unicode used for the up arrows (there are several arrow symbols in Unicode)
+    left_arrow - the unicode used for the left arrows
+    up_left_arrow - the unicode used for the diagonal arrows
+    stop - the symbol used in the upper left to indicate the end of the alignment
+    """
+
+    row = len(seq1)
+    col = len(seq2)
+    arrow = traceback_array[row, col]
+    aligned_seq1 = ""
+    aligned_seq2 = ""
+    alignment_indicator = np.zeros(len(seq2))
+    while arrow != 0:
+        arrow = traceback_array[row, col]
+        if arrow == up_arrow:
+            # We want to add the new indel onto the left
+            # side of the growing aligned sequence
+            aligned_seq2 = "-" + aligned_seq2
+            aligned_seq1 = seq1[row - 1] + aligned_seq1
+            # alignment_indicator = " " + alignment_indicator
+            row -= 1
+
+        elif arrow == up_left_arrow:
+            # Note that we look up the row-1 and col-1 indexes
+            # because there is an extra "-" character at the
+            # start of each sequence
+            seq1_character = seq1[row - 1]
+            seq2_character = seq2[col - 1]
+            aligned_seq1 = seq1[row - 1] + aligned_seq1
+            aligned_seq2 = seq2[col - 1] + aligned_seq2
+            if seq1_character == seq2_character:
+                # alignment_indicator = "|" + alignment_indicator
+                alignment_indicator[col-1] = 1
+            # else:
+
+                # alignment_indicator = " " + alignment_indicator
+            row -= 1
+            col -= 1
+
+        elif arrow == left_arrow:
+            aligned_seq1 = "-" + aligned_seq1
+            aligned_seq2 = seq2[col - 1] + aligned_seq2
+            # alignment_indicator = " " + alignment_indicator
+            col -= 1
+
+        elif arrow == stop:
+            break
+        else:
+            msg = (f"Traceback array entry at {row},{col}: {arrow} is not recognized as an up arrow"
+                   "({up_arrow}),left_arrow ({left_arrow}), up_left_arrow ({up_left_arrow}), or a stop ({stop}).")
+            raise ValueError(msg)
+
+    return aligned_seq1, alignment_indicator, aligned_seq2
+
+
+def nw(seq1, seq2, match = 1, mismatch = -1, indel = -1):
+
+    n_rows = len(seq1)+1
+    n_columns = len(seq2)+1
+
+    scoring_array = np.full([n_rows, n_columns], 0)
+    traceback_array = np.full([n_rows, n_columns], np.nan)
+    up_arrow = 1
+    left_arrow = 2
+    up_left_arrow = 3
+
+    for row in range(n_rows):
+        for col in range(n_columns):
+            if row == 0 and col == 0:
+                # We're in the upper right corner
+                score = 0
+                arrow = 0
+            elif row == 0:
+                # We're on the first row
+                # but NOT in the corner
+
+                # Look up the score of the previous cell (to the left) in the score array
+                previous_score = scoring_array[row, col - 1]
+                # add the gap penalty to it's score
+                score = previous_score + indel
+                arrow = left_arrow
+            elif col == 0:
+                # We're on the first column but not in the first row
+                previous_score = scoring_array[row - 1, col]
+                score = previous_score + indel
+                arrow = up_arrow
+            else:
+                # Calculate the scores for coming from above,
+                # from the left, (representing an insertion into seq1)
+                cell_to_the_left = scoring_array[row, col - 1]
+                from_left_score = cell_to_the_left + indel
+
+                # or from above (representing an insertion into seq2)
+                above_cell = scoring_array[row - 1, col]
+                from_above_score = above_cell + indel
+
+                # diagonal cell, representing a substitution (e.g. A --> T)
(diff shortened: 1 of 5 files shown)


vermouth/pdb/nwalign.py
```diff
@@ -0,0 +1,238 @@
...
+# limitations under the License.
+
+'''
+
+Sequence alignment with the Needleman-Wunsch algorithm
+
+Designed to check the SEQRES entry of an input PDB versus the residues present
+
+Adapted from
+https://github.com/zaneveld/full_spectrum_bioinformatics/blob/master/content/08_phylogenetic_trees/needleman_wunsch_alignment.ipynb
```
pckroon, line 23: This is a problem. That code is CC BY-NC-SA 4.0 licensed, which is /way/ more restrictive than the Apache license we carry. See https://creativecommons.org/licenses/by-nc-sa/4.0/

pckroon (changes requested):
We need a different alignment implementation, since you based your implementation on code that is way stricter licensed than the apache license we use.

vermouth/pdb/nwalign.py
fgrunewald: I don't see the issue here; what Chris referenced is an educational resource and if I'm not mistaken but that's for Chris to confirm he did not copy the code in verbatim and / or large chunks of it. The algorithm itself is not restricted by the Licence

pckroon: Any derivative of that code has to be either CC BY-NC-SA or GPL licensed. The algorithm itself is indeed not covered, but derivative implementations are. The fact that it's an educational resource is irrelevant.

fgrunewald: I guess the question is more at what point is it a derivative code?

fgrunewald: could we ask for special permission?

pckroon: > I guess the question is more at what point is it a derivative code?

IANAL, so no idea. I err on the side of caution here

> could we ask for special permission?

We could actually. However, don't you think it'll be less work to build off of an MIT/BSD/Apache implementation? I also sent an old ab-initio implementation of my own to @csbrasnett earlier today.

csbrasnett: I'm looking at the code @pckroon sent earlier now, it won't be too much effort to adapt so I'll sort it out in the next commit

fgrunewald: I'm fine with either way; I didn't look closely enough at the code to know how difficult that is. In general any work is always more work than just sending a mail/request. But in the end it's the choice of @csbrasnett

Tsjerk: It's not their algorithm. Just simple, somewhat convoluted Needleman-Wunsch. I have an (own) implementation of it that could be put in instead. Can dig that up. No licensing issues then.

csbrasnett:
A solution to make everyone happy, remove the NW alignment altogether and just check the seqres directly against the pdb entries.

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
             mol.add_edge(atomidx0, atomidx, distance=dist)
 
+    def do_seqres(self):
+        """
+        Get the AA sequences of the chains in the molecules
+        """
+        chains = list(list(np.unique(list(mol.nodes[idx]['chain'] for idx in mol)))
+                      for mol in self.molecules)
+
+        properties = {a: [] for a in [x for xs in chains for x in xs]}
```
pckroon, line 469: At the very least this needs a comment with what `chains` looks like.

vermouth/pdb/pdb.py
```diff
@@ -130,7 +131,20 @@ def _skip(line, lineno=0):
...
+        """
+        Parse a SEQRES record. Store for processing later.
+
+        Parameters
+        ----------
+        line: str
+            The line to parse.
+
+        """
+        self._seqres.append(line)
```
pckroon, line 145: IMHO it would be better to deal with the syntax of the SEQRES line here, and indeed deal with the semantics in `do_seqres`

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+        Get the AA sequences of the chains in the molecules
+        """
+        chains = list(list(np.unique(list(mol.nodes[idx]['chain'] for idx in mol)))
+                      for mol in self.molecules)
+
+        properties = {a: [] for a in [x for xs in chains for x in xs]}
+
+        for line in self._seqres:
+            chain = line[11]
+            resnames = line[19:].split()
```
pckroon, line 473: As mentioned earlier, I suggest moving this to the method `seqres`. This also makes dealing with malformatted seqres lines easier

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+            #     assert list(set(len(i) for i in resnames)) == 3
+            # except AssertionError:
+            #     #we don't really need to worry about this
+            #     #we just need to not use it any further
+            #     pass
+            properties[chain].append(resnames)
+
+        # this to make a flat list for each chain in the file
+        for chain in properties.keys():
+            properties[chain] = [x for xs in properties[chain] for x in xs]
```
pckroon, line 487: Honest question: do we want a list or str?

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+        self._check_seqres(properties)
+
+
+    def _check_seqres(self, properties):
+        '''
+        In future this could just be done by parsing REMARK 465.
```
pckroon, line 493: ??

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+        '''
+        LOGGER.info("Checking pdb SEQRES entry for missing residues", type="step")
+        for mol in self.molecules:
+            resids = np.array([mol.nodes[idx]['resid'] for idx in mol],dtype = int)
```
pckroon, line 497:
```suggestion
            resids = np.array([mol.nodes[idx]['resid'] for idx in mol], dtype = int)
```

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+            chain = list(set([mol.nodes[idx]['chain'] for idx in mol]))[0]
```
pckroon, line 498: Why do you select only the first item?

```suggestion
            chain = next(iter(set([mol.nodes[idx]['chain'] for idx in mol])))
```

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+
+            missing_res_nos = np.setdiff1d(np.arange(len(properties[chain]))+1,
+                                       np.unique(resids)[np.unique(resids) < len(properties[chain])+1])
```
pckroon, line 501: Needs a comment :)
Does this assume resid numbers are sane?

vermouth/pdb/pdb.py
```diff
@@ -443,6 +459,58 @@ def _do_single_conect(self, conect_record, id2idxs):
...
+
+            #find consecutive sequences of missing residues
+            nums = sorted(set(missing_res_nos))
+            gaps = [[s, e] for s, e in zip(nums, nums[1:]) if s + 1 < e]
+            edges = iter(nums[:1] + sum(gaps, []) + nums[-1:])
+            series = list(zip(edges, edges))
```
pckroon, line 507: I guess this gets the TODO: alignment?
In addition, do we want to check that the resnames are correct?

vermouth/tests/pdb/test_read_pdb.py
```diff
@@ -264,3 +264,39 @@ def test_cryst1(caplog, pdbstr, cryst_dict):
...
+    ATOM      5  CB  SER A   1      26.459  10.418  75.878  1.00 32.29           C  
+    ATOM      6  OG  SER A   1      25.340  11.168  76.321  1.00 31.34           O  
+    ATOM      7  N   SER A   2      26.091  11.817  73.405  1.00 33.30           N  
+    ATOM      8  CA  SER A   2      25.418  12.662  72.416  1.00 33.24           C  
+    ATOM      9  C   SER A   2      24.171  13.333  72.984  1.00 32.59           C  
+    ATOM     10  O   SER A   2      23.457  14.042  72.271  1.00 32.68           O  
+    ATOM     11  CB  SER A   2      25.020  11.833  71.196  1.00 33.13           C  
+    ATOM     12  OG  SER A   2      24.049  10.863  71.549  1.00 34.53           O  
+''', True)))
+def test_seqres(caplog, pdbstr, status):
```
pckroon, line 298: Add a test without a SEQRES entry, and one where there are enough residues, but they have different resnames

pckroon (changes requested):
I like it :)
I do have some questions and comments, but in general I'm a fan. 
I also approve of the decision of just ripping out the alignment for now.

vermouth/pdb/pdb.py
csbrasnett: the list should only contain a single item, it could equally say `mol.nodes[0]['chain]`

vermouth/pdb/pdb.py
csbrasnett: see page 130 of the [pdb format file](https://files.wwpdb.org/pub/pdb/doc/format_descriptions/Format_v33_Letter.pdf) - missing residues are explicitly listed in the REMARK section of pdb files, but getting that will require parsing the whole REMARK section, which I think is well beyond anyone's interest atm...!

vermouth/pdb/pdb.py
csbrasnett: see page 130 of the [pdb format file](https://files.wwpdb.org/pub/pdb/doc/format_descriptions/Format_v33_Letter.pdf) - missing residues are explicitly listed in the REMARK section of pdb files, but getting that will require parsing the whole REMARK section, which I think is well beyond anyone's interest atm...!

vermouth/pdb/pdb.py
csbrasnett: added. The < in the pdb residues array ensures that any residues that are larger (eg. for solvent molecules attached to the chain) are excluded, if that's what you mean by sanity?

vermouth/pdb/pdb.py
pckroon: Ah, right. And yes. Definitely out of scope for now!

vermouth/pdb/pdb.py
pckroon: Alright. Might be worth adding an `assert` then

vermouth/pdb/pdb.py
pckroon: Does this mean that the residnumbers have to be in order? Do solvents etc *always* have to be last? I guess the real solution is to do the alignment?

vermouth/pdb/pdb.py
csbrasnett: keeping this as a list makes element comparison/indexing by resid easier later on

vermouth/pdb/pdb.py
csbrasnett: I don't think the resids have to necessarily be in order, they just have to exist. I agree that it's a problem if there are solvent residues listed first though, not sure how to get around that... 

I think increasingly there may have to be some kind of alignment included. I've now found a problem if there are eg. expression tags, which will be listed in the seqres, but not actually be part of the 'proper' sequence. For this, either we bring back some kind of alignment, or I parse SEQADV for this purpose. I think the latter should be easy enough to do given it will only be used internally for SEQRES comparison. Thoughts?

pckroon: Your call. I don't enough about the pdb format (and not really time to look it up) to say anything about the SEQADV. I do think the alignment is (going to be) important, but that can be medium-long term. For now I see 2 options: 1) SEQADV; 2) if it doesn't match, complain. You have to see whether that would give too many false positives, in which case you may want to make it an INFO message rather than warning.

Tsjerk: Why/when would alignment be important? In files that have sequence, it should correspond in a simple, literal way, so can be done without fancy alignment. But those files will also have REMARK 465, i dicating which residues in the sequence are not resolved. For files without sequence, there is nothing to align to, fancy or not.

pckroon: Iff the seqres does not match the found residues, where is it wrong? And can you guarantee that all pdb files that have a seqres have remark 465? I don't dare make *any* assumptions about pdb files, based on my experience so far.

Tsjerk: If the seqres does not match, someone has deliberately changed the coordinates section, and did not update SEQRES. I guess a deliberate change is meant to be processed. If there is a SEQRES, it's probably good to use it for sanity check, but not using NW or other alignment, because breaks may be assigned wrongly. Use it only for simple verbatim checking, preferably in conjunction with REMARK 465, and if it doesn't match, it makes sense to issue an error; the user can then discard the SEQRES record if that's what was meant.

pckroon: The question is more about what you do when the seqres doesn't match. 
1) You warn: "Seqres is wrong, good luck!"
2) You assume Remark 465 is correct and info after checking: "Seqres is wrong, but the deviations are specified in R465"
3) You warn: "Seqres is wrong and this is not described in R465. Probably, residues X-Z are missing, residue ALA53 was changed to PHE. Sort out your mess."

As user, I would like warning 3 most. But I'm completely fine with just 1 or 2.

Partly this also tied in with how much we trust CHAIN identifiers, but I guess it would be sane to assume they're at least consistent within the pdb file (i.e. between seqres and the coordinate section)

PR #581, opened 2024-03-27
Fix 580

### LINKED ISSUE #580: Slient dropping of edge attributes
Several functions, that operate on Molecule, do not carry over edge attributes. For now, those are:

- to_molecule
- merge_molecule 

@pckroon I believe this is a bug and the edge attributes should be carried over or am I missing something.

fgrunewald:
small fix to the issues in #580

Fixes #580


pckroon (approved):
Nice work.

PR #583, opened 2024-04-05
Fix #558 by inverting the link logic generating charged backbone charge dummy exclusions for polarizable forcefields

### LINKED ISSUE #558: Wrong exclusions for charged termini in 2.2P
Hello,

I have charged terminal beads: N-terminal is Qd, with a +1 charge and index 1, while C-terminal is Qa, with a -1 charge and index 2044. I would expect these to interact with anything except for their own side chains. 

The exclusions are exactly opposite what I would expect. The N-terminal (and C-terminal) bead has exclusions for the virtual particles (SCP/SCN) for all other residues, but not for the virtual particles within that same residue. 

For the N-terminal residue, during my vacuum minimization, the side chain reorients itself such that the negatively-charged bead gets inserted into its own backbone bead. The resulting forces are very high and the system crashes immediately in the next simulation step. I think this interaction should be excluded.

Am I missing some logic for why we wouldn't want the termini interacting with the polarized side chains? I think it would be strange to have the termini interact with water/lipids/ions etc. but not other protein. 

FYI, I do have elastic network activated, though I'm not sure whether it's relevant to this problem.

Maybe this was missed because tests weren't done on residues with polarized side chains at patched termini?

pckroon:
Turns out those exclusions were missing from elnedyn22p completely as well...

### DIFF
--- a/vermouth/data/force_fields/elnedyn22p/aminoacids.ff
+++ b/vermouth/data/force_fields/elnedyn22p/aminoacids.ff
@@ -687,3 +687,8 @@ SC1 >SC1 1 0.24 {"comment": "Disulfide bridge"}
 ;[ features ]
 ;disulfide
 
+; Exclusions between the charged terminii and the charge dummies
+[ link ]
+resname $protein_resnames
+[ exclusions ]
+BB {"atype": "Qd|Qa"} XX {"atomname": "SCP|SCN"}
--- a/vermouth/data/force_fields/martini22p/aminoacids.ff
+++ b/vermouth/data/force_fields/martini22p/aminoacids.ff
@@ -894,4 +894,4 @@ SC1 >SC1 1 0.24 {"comment": "Disulfide bridge"}
 [ link ]
 resname $protein_resnames
 [ exclusions ]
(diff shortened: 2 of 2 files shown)


vermouth/data/force_fields/elnedyn22p/aminoacids.ff
```diff
@@ -687,3 +687,8 @@ SC1 >SC1 1 0.24 {"comment": "Disulfide bridge"}
 ;[ features ]
 ;disulfide
 
+; Exclusions between the charged terminii and the charge dummies
+[ link ]
+resname $protein_resnames
+[ exclusions ]
+BB {"atype": "Qd|Qa"} XX {"atomname": "SCP|SCN"}
```
fgrunewald, line 694: @pckroon uhhh this works? I didn't know you could make the atomname attribute a variable as will. You know that this changes the whole game of making links?

pckroon: Apparently it works. There's no magic code to make this possible even, so yes, it's general. Not sure how you would deal with the case where *any* atomname is acceptable though.

fgrunewald (approved):

Issue #587, opened 2024-04-16
Lazy merging?


csbrasnett:
If I have a protein complex with many chains and just want a single merged itp out, the current input is slightly annoying in having to make list of all of them (although, I often think this of even something with 2 chains... more of a reflection on me than anything else...) 

Is there any reason in principle implementing something like `-merge all` is a bad idea? Happy to do the PR myself but just wanted to check if this has been discussed and rejected before!

pckroon:
By all means go ahead, I think it's a fine feature :)
I can't think of any reason why this would be a particularly bad idea.

PR #589, opened 2024-04-23
Lazy merge

csbrasnett:
Lazy merging of chains as per #587. Instead of specifying groups of chains to merge, users can just give `-merge all` and all chains get merged.


vermouth/processors/merge_chains.py
```diff
@@ -42,8 +42,15 @@ def merge_chains(system, chains):
         The system to modify.
     chains: list[str]
         A container of chain identifier.
+    all_chains:
```
pckroon, line 45:
```suggestion
    all_chains: bool
        If True, all chains will be merged.
```

vermouth/processors/merge_chains.py
```diff
@@ -42,8 +42,15 @@ def merge_chains(system, chains):
...
     """
-    chains = set(chains)
+    if not all_chains:
+        chains = set(chains)
+    else:
+        l = []
+        for molecule in system.molecules:
+            l.append([node.get('chain') for node in molecule.nodes.values()][0])
+        chains = set(l)
```
pckroon, line 53:
```suggestion
        chains = set()
        for molecule in system.molecules:
            chains.update(node.get('chain') for node in molecule.nodes)  # Molecules can contain multiple chains
```
Do we need to deal with `None` and `''` chains?

vermouth/processors/merge_chains.py
```diff
@@ -42,8 +42,15 @@ def merge_chains(system, chains):
         The system to modify.
     chains: list[str]
         A container of chain identifier.
+    all_chains:
     """
-    chains = set(chains)
+    if not all_chains:
+        chains = set(chains)
+    else:
```
pckroon, line 49: What happens if both all_chains=True and chains is passed? IMO that should probably raise an error of some sorts.

pckroon (changes requested):

csbrasnett:
If/once you've merged the codecov update in #565 I'll merge main into here and I guess that'll fix the tests

bin/martinize2
```diff
@@ -972,8 +973,14 @@ def entry():
...
         # Merge chains if required.
         if args.merge_chains:
-            for chain_set in args.merge_chains:
-                vermouth.MergeChains(chain_set).run_system(system)
+            if "all" not in args.merge_chains:
+                input_chain_sets = [i.split(",") for i in args.merge_chains]
+                for chain_set in input_chain_sets:
+                    vermouth.MergeChains(chains=chain_set, all_chains=False).run_system(system)
+            elif "all" in args.merge_chains and len(args.merge_chains) == 1:
+                vermouth.MergeChains(chains=None, all_chains=True).run_system(system)
```
pckroon, line 981: `args.merge_chains` cannot be 'all' and have a length of 1 right? Since the split only happens in the `if`

bin/martinize2
```diff
@@ -972,8 +973,14 @@ def entry():
...
+            else:
+                LOGGER.warning("Multiple conflicting merging arguments given. Please check input arguments.")
```
pckroon, line 983: I think this can be stronger, and raise an ArgumentError explaining the conflict

vermouth/processors/merge_chains.py
```diff
@@ -42,15 +44,32 @@ def merge_chains(system, chains):
...
+    if not all_chains and len(chains)>0:
+        _chains = set(chains)
+    elif all_chains and chains is None:
+        _chains = set()
+        for molecule in system.molecules:
+            # Molecules can contain multiple chains
+            _chains.update(node.get('chain') for node in molecule.nodes.values())
+    else:
+        LOGGER.warning('Conflicting merging arguments have been given. Will abort merging.')
+        return
```
pckroon, line 59: Here too, just `raise ValueError` would be appropriate.

vermouth/processors/merge_chains.py
```diff
@@ -42,15 +44,32 @@ def merge_chains(system, chains):
...
+
+    try:
+        assert ''.join(list(_chains)).isalnum()
+    except TypeError:
+        LOGGER.warning('One or more of your chains does not have a chain identifier in input file.')
```
pckroon, line 64:
```suggestion
    if any(not c for c in _chains):
        LOGGER.warning('One or more of your chains does not have a chain identifier in input file.')
```

vermouth/processors/merge_chains.py
```diff
@@ -65,8 +84,9 @@ def merge_chains(system, chains):
 class MergeChains(Processor):
     name = 'MergeChains'
 
-    def __init__(self, chains):
+    def __init__(self, chains=None, all_chains=False):
         self.chains = chains
+        self.all_chains = all_chains
```
pckroon, line 89:
```suggestion
    def __init__(self, chains=None, all_chains=False):
        self.chains = chains or []
        self.all_chains = all_chains
```

pckroon (changes requested):

bin/martinize2
csbrasnett: this is defending against multiple arguments though, eg. if you give both `-merge all -merge A,B` then `args.merge_chains = ['all, 'A,B']`, so really you need to make sure there really is only one argument

pckroon: Good catch! Needs a comment though.

vermouth/processors/merge_chains.py
```diff
@@ -84,7 +81,7 @@ def merge_chains(system, chains, all_chains):
 class MergeChains(Processor):
     name = 'MergeChains'
 
-    def __init__(self, chains=None, all_chains=False):
+    def __init__(self, chains=[], all_chains=False):
         self.chains = chains
         self.all_chains = all_chains
```
pckroon, line 86: It's generally a bad idea to give a mutable type as default argument, since it will be created only once and shared among all calls/instances. Here it's *probably* fine since `chains` is not modified anywhere, but still, something like this is preferred:
```py
def __init__(self, chains=None, all_chains=False):
    self.chains = chains or []
    self.all_chains = all_chains.
```

As an example of what can go wrong:
```py
def a(x=[]):
    x.append(1)
    print(x)

a()
a()
```

vermouth/tests/test_merge_chains.py
```diff
@@ -118,4 +92,37 @@ def test_merge(caplog, node_data, edge_data, merger, expected):
...
+
+    mols = nx.connected_components(mol)
+    for nodes in mols:
+        system.add_molecule(mol.subgraph(nodes))
+
+    merger = {"chains": ["A", "B"], "all_chains": True}
+
+    processor = MergeChains()
+    processor.chains = merger["chains"]
+    processor.all_chains = merger["all_chains"]
```
pckroon, line 124:
```py
processor = MergeChains(**merger)
```

vermouth/processors/merge_chains.py
```diff
@@ -47,20 +47,17 @@ def merge_chains(system, chains, all_chains):
...
-    elif all_chains and chains is None:
+    elif all_chains and len(chains) == 0:
         _chains = set()
         for molecule in system.molecules:
             # Molecules can contain multiple chains
             _chains.update(node.get('chain') for node in molecule.nodes.values())
     else:
-        LOGGER.warning('Conflicting merging arguments have been given. Will abort merging.')
-        return
+        raise ValueError
```
pckroon, line 58: Do we want an informative message here?

pckroon (changes requested):
Small nitpicks left :)

vermouth/processors/merge_chains.py
csbrasnett: Thanks!

vermouth/processors/merge_chains.py
csbrasnett: Have put one, but probably not necessary. Martinize2 will raise its ArgumentError before it gets here if both have been given, so this would have to be someone using the processor on its own for some reason!

pckroon (approved):

vermouth/processors/merge_chains.py
pckroon: Exactly that, but that *is* a possibility we want to enable/encourage/support

pckroon:
Excellent work, thanks!

Issue #585, opened 2024-04-12
Expose Go model `backbone` and `atomname` in `add_virtual_sites`


Lp0lp:
Would be useful to expose the `backbone` and `atomname` variables in [`add_virtual_sites`](https://github.com/marrink-lab/vermouth-martinize/blob/dbe0d76e3f4f36c005cc5c274ee219ecb22355f5/vermouth/rcsu/go_vs_includes.py#L63), to more easily control where to place the GoVirt bead and how to name it, via the CLI. 

https://github.com/marrink-lab/vermouth-martinize/blob/dbe0d76e3f4f36c005cc5c274ee219ecb22355f5/vermouth/rcsu/go_vs_includes.py#L63

pckroon:
Sure, PR welcome. I also see no reason why not to expose the charge argument to the CLI as well then.

csbrasnett:
Close with release of version 0.13.0

PR #590, opened 2024-04-23
GO IDP bugfix

csbrasnett:
Need to add secstruc to the water bias application, otherwise can't apply across a whole protein (starting resid has to be > 1). this should fix that. And add preprint citation.


fgrunewald (changes requested):
Needs at least one integration test in my opinion. 

@csbrasnett do you think the citation should also be added in the water bias processor?

csbrasnett:
I'm not certain, but I think adding again there would be degenerate anyway? VirtualSiteCreator gets called both in the Go model and if `args.water_bias` is used, so I think just adding it there should be fine.

Happy to write an integration test. It strikes me this might be a good opportunity to deal with #585 at the same time, so I can do that as well?

fgrunewald:
@csbrasnett I'd rather have 2 separate PRs, so that we can release this bug fix in near time

fgrunewald (changes requested):
looks good but waiting for @csbrasnett to complete all changes

csbrasnett:
This *should* bring things in line with polyply for generating topologies of idps. Maybe this is the right approach, maybe it isn't, but here's some kind of implementation either way. Happy to be advised whatever.

To do:

1) properly check against polyply
2) make some tests
3) maybe propagate secondary structure for automatic id/suggestions of IDRs?

PR #593, opened 2024-04-29
Idp go

csbrasnett:
Adding -idp-tune to allow for addition of extra bonded potentials in IDRs


vermouth/processors/tune_idp_bonds.py
```diff
@@ -0,0 +1,315 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Copyright 2018 University of Groningen
```
pckroon, line 3:
```suggestion
# Copyright 2024 University of Groningen
```

vermouth/processors/tune_idp_bonds.py
```diff
@@ -0,0 +1,315 @@
...
+        if not self.system:
+            raise IOError('This processor requires a system.')
+
+        if not molecule.meta.get('moltype'):
+            raise ValueError('The molecule does not have a moltype name.')
+
+        if hasattr(molecule, 'res_graph'):
+            res_graph = molecule.res_graph
+        else:
+            res_graph = make_residue_graph(molecule)
```
pckroon, line 291:
```suggestion
        if hasattr(molecule, 'res_graph'):
            res_graph = molecule.res_graph
        else:
            res_graph = make_residue_graph(molecule)
            molecule.res_graph = res_graph
```
?

vermouth/processors/tune_idp_bonds.py
```diff
@@ -0,0 +1,315 @@
...
+                                                          "version": 1}
+                                                    )
+                molecule.add_or_replace_interaction('dihedrals',
+                                                    atoms=bbbs_atomlist,
+                                                    parameters=['1', '0', '-2.0', '1'],
+                                                    meta={"group": "idp-fix", "comment": "BB-BB(GLY)-BB-SC1-v1",
+                                                          "version": 1}
+                                                    )
+
+    def run_molecule(self, molecule):
```
pckroon, line 278: Don't forget to attach the appropriate citation reference to the molecule

vermouth/processors/tune_idp_bonds.py
```diff
@@ -0,0 +1,315 @@
...
+                                                              resid=resid,
+                                                              chain=chain,
+                                                              prefix=molecule.meta.get('moltype')))
+                all_cross_pairs.append(np.where(all_go_pairs == vs_go_node)[0]) #just need the first one
+
+        # make sure we only have one entry in case a site has more than one interaction
+        all_cross_pairs = np.unique([x for xs in all_cross_pairs for x in xs])
+        # delete the folded-disordered Go interactions from the list going backwards.
+        # otherwise list order gets messed up.
+        for i in all_cross_pairs[::-1]:
```
pckroon, line 79:
```suggestion
        for i in reversed(all_cross_pairs):
```

pckroon (changes requested):
Some small code comments, and a conceptual suggestion.
You want to add some interactions to specific disordered regions of a protein, right? My suggestion would be to have a simple processor annotate nodes/residues that are in an IDR, and add the interactions using links

csbrasnett:
This is a much nicer solution, thanks for suggesting!

bin/martinize2
```diff
@@ -1067,13 +1069,13 @@ def entry():
...
-                                         ).run_system(system)
+        if not target_ff.has_feature("idr"):
+            LOGGER.warning('Improved IDR potentials are not implemented '
+                           'for this force field',
+                           type="missing-feature")
         else:
-            raise argparse.ArgumentError(idr_tuning,
-                                         message=("You are trying to apply IDR biasing to a force field which "
-                                                  "isn't martini3001, which this is not designed for."))
+            vermouth.IDRBonds(idr_regions = [(int(start), int(stop)) for start, stop in args.water_idrs]).run_system(system)
```
pckroon, line 1077:
```suggestion
            vermouth.IDRBonds(idr_regions=[(int(start), int(stop)) for start, stop in args.water_idrs]).run_system(system)
```

bin/martinize2
```diff
@@ -1067,13 +1069,13 @@ def entry():
...
+            vermouth.DoLinks().run_system(system)
```
pckroon, line 1078: I don't think you actually want to run DoLinks a second time. Instead, move the IDRBonds annotator to just after the mapping

vermouth/processors/tune_idp_bonds.py
```diff
@@ -21,14 +21,34 @@
...
 
+def make_disorder_string(molecule, idr_regions):
+    """
+    make a string for annotating order and disorder along a molecule
+    """
+
+    order_disorder_seq = []
+
+    for key, node in molecule.nodes.items():
+        if select_backbone(node):
```
pckroon, line 40: I wouldn't tag just the  backbone (or SC1), but rather just all nodes with the appropriate resid

vermouth/processors/tune_idp_bonds.py
```diff
@@ -21,14 +21,34 @@
...
+            _old_resid = node['_old_resid']
+            if _in_resid_region(_old_resid, idr_regions):
+                order_disorder_seq.append("D")  # D for disordered
```
pckroon, line 43: Why first build the list/string, and not just do `molecule.nodes[key]['is_idr'] = True` here?

pckroon (changes requested):
Much cleaner! <3
Still a few small comments

bin/martinize2
```diff
@@ -170,7 +170,7 @@ def pdb_to_universal(
     return canonicalized
 
 
-def martinize(system, mappings, to_ff, delete_unknown=False):
+def martinize(system, mappings, to_ff, delete_unknown=False, idrs=False, disordered_regions = []):
```
pckroon, line 173:
```suggestion
def martinize(system, mappings, to_ff, delete_unknown=False, idrs=False, disordered_regions=(,)):
```
As before, no mutable types as default arguments

bin/martinize2
```diff
@@ -1066,16 +1076,16 @@ def entry():
...
+    #
+    # # apply idr specific bonded potentials if required
+    # if args.idr_tune:
+    #     if not target_ff.has_feature("idr"):
+    #         LOGGER.warning('Improved IDR potentials are not implemented '
+    #                        'for this force field',
+    #                        type="missing-feature")
+    #     else:
+    #         vermouth.IDRBonds(idr_regions=[(int(start), int(stop)) for start, stop in args.water_idrs]).run_system(system)
+    #         vermouth.DoLinks().run_system(system)
```
pckroon:
```suggestion
```

vermouth/processors/annotate_idrs.py
```diff
@@ -0,0 +1,81 @@
...
+from ..graph_utils import make_residue_graph
+from ..rcsu.go_utils import _in_resid_region
+from ..log_helpers import StyleAdapter, get_logger
+LOGGER = StyleAdapter(get_logger(__name__))
+
+
+def annotate_disorder(molecule, idr_regions):
+    """
+    annotate the disordered regions of the molecule
+    """
```
pckroon, line 32: Docstring is incomplete.
Also, is it worth making the attribute used an argument (with default "cgidr")?

vermouth/processors/annotate_idrs.py
```diff
@@ -0,0 +1,81 @@
...
+    Processor to annotate intrinsically disordered regions of a molecule.
+
+    This processor is designed primarily for the work described in the reference
+    M3_GO, but is generally applicable for such circumstances where extra
+    addition/removals are necessary.
+
+    """
+
+    def __init__(self, idr_regions = None):
```
pckroon, line 49:
```suggestion
    def __init__(self, idr_regions=None):
```

vermouth/processors/annotate_idrs.py
```diff
@@ -0,0 +1,81 @@
...
+        is always molecule specific i.e. no two different
+        vermouth molecules can have the same bias.
+
+        Parameters
+        ----------
+        system: :class:`vermouth.system.System`
+        """
+        if not self.idr_regions:
+            return system
+        self.system = system
```
pckroon, line 79:
```suggestion
```
Why do you need this?

pckroon (changes requested):

csbrasnett:
btw, in trying to write some tests to address the missing coverage I've found an error in the test here:

https://github.com/marrink-lab/vermouth-martinize/blob/e9bf1c928efde188d70f9d9c710032a8a2f81672/vermouth/tests/test_water_bias.py#L89-L98

in that the if statement is testing an atomtype against an atom index. Easy enough to fix, but it does now cause this test to fail once the the eps and sig assertions are actually reached! I'll fix here and you can have a look

pckroon (dismissed):
LGTM! Was there anything still missing in this PR?

csbrasnett:
Thanks! Making a test for the folded/disordered go bond removal, should be done soon!

csbrasnett:
you mean in addition to what I've done here now?

pckroon:
Yes :)

csbrasnett:
This is proving more difficult than anticipated... I'm guessing in general you have something in mind like this. Ie. you have two folded domains connected by some linker, and you want to 1) keep intra-domain go bonds, 2) remove domain-linker bonds, and 3) remove inter-domain bonds? I this should be possible but I need to make some big edits to the removal functionality 

image

pckroon:
What I mean is that in `test_water_bias.py:test_cross_go_bond_removal` you either remove all or none. I'm not sure that's going to be the same case as you describe tbh

csbrasnett:
ok, I think that addresses your point?

pckroon (approved):
Yes, perfect

PR #596, opened 2024-05-08
External EN bonds itp

csbrasnett:
Add a flag, `-ew`, to write out bonds for elastic networks to an external en_bonds.itp file, in the style of Go or Olive bonds.

This has been motivated by some work on visualising cg molecules at [martini_vis](https://github.com/csbrasnett/martini_vis) for now. If the bonds are written out externally, then they can be used (after some editing elsewhere) to visualise the elastic network separately to the bonds inherent in the molecule. Maybe this is useful, maybe it isn't, it's a feature at least!

pros:
- might be able to visualise elastic networks better

cons:
- relies on users actually writing `#include "en_bonds.itp"` in their `[ bonds ]` directive in their molecule's itp file


bin/martinize2
```diff
@@ -536,6 +536,13 @@ def entry():
             "format: <start_resid_1>:<end_resid_1>, <start_resid_2>:<end_resid_2>..."
         ),
     )
+    rb_group.add_argument(
+        "-ew",
+        dest="rb_write",
+        default=False,
+        action="store_true",
+        help=("Write the elastic network to an external .itp file"),
```
pckroon, line 544:
```suggestion
        help="Write the elastic network to an external .itp file",
```

vermouth/processors/apply_rubber_band.py
```diff
@@ -349,12 +354,22 @@ def apply_rubber_band(molecule, selector,
...
+                                        fc=force_constant,
+                                        meta={'group': 'Rubber band'})
+                ebs.append(rubber_band)
+            else:
+                molecule.add_interaction(
+                    type_='bonds',
+                    atoms=(from_key, to_key),
+                    parameters=[bond_type, length, force_constant],
+                    meta={'group': 'Rubber band'},
+                )
```
pckroon, line 370: Fewer if statements is better, so I suggest always gathering the rubber band interactions into a list, and at the end decide whether or not you add the interactions to the molecule or not

pckroon (changes requested):
I like the feature. 
As for the implementation, I have the following suggestion: Don't have the RubberBands processor return the bonds, instead just add them to the molecule. The ITP writer can then select which interactions to write where based on some sort of selector. This would also allow it to write the appropriate `#include` statement. This will at some point also allow you to easily write other bonds or I don't know, dihedrals to external files. It'll also make the `BondParam` namedtuple superfluous.

vermouth/processors/apply_rubber_band.py
pckroon: Just add them to the molecule directly, and find them again later based on the meta attributes.

csbrasnett:
ok, I think this is closer to how you want this to be handled? But, I don't think either of us are happy with this solution, so I'll keep working at something cleaner

pckroon:
Jup, definitely in the right direction. My original idea was to not add an extra interaction type ('el_bonds'), but rather, iterate  over interactions and grouping them using some form of selection function. For example, all interactions that have `meta={'group': 'Rubber bands'}` go in one file, and all interactions that have a meta classifying them as scFix interactions to go somewhere else, all interactions involving an SC4 go to a third, and everything else stays in the main file. Or something like that. This probably involves some sort of `itertools.groupby` and some juggling with selector functions.

Also, the `#include` doesn't need to be in the `[bonds]` directive, does it? Just put them at the end of the file. That way you can also cram multiple types of interactions (bonds, angles) in the same external file

PR #602, opened 2024-06-10
fix error raising for mod not found

csbrasnett:
need to reset extra for every resspec, otherwise 'success': False gets appended for every non-find and 'success': True is never added. This only actually affected the error raised, the annotation was actually applied.

another fix for #553


pckroon (approved):
Awesome, thanks!

PR #611, opened 2024-09-06
Rework the deploy pipeline

pckroon:
Based on https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -125,10 +125,11 @@ jobs:
     permissions:
       id-token: write
     steps:
+    - uses: actions/checkout@v4
     - name: Set up Python 3
       uses: actions/setup-python@v5
       with:
-        python-version: '3.x'
+        python-version: '3.11'
         cache: pip
         cache-dependency-path: |
           **/setup.cfg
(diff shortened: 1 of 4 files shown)


.github/workflows/run_tests.yml
```diff
@@ -116,3 +115,39 @@ jobs:
         run: | 
           mkdir -p doc/source/_static
           sphinx-build -EnW -b html doc/source/ doc/build/html
+
+  deploy:
```
fgrunewald, line 119: Isn't this double now i.e. defined in deploy.yml and here?

fgrunewald (commented):
looks ok to me; one comment

.github/workflows/run_tests.yml
pckroon: No, this one deploys to test.pypi.org rather than actual pypi.org. I would like to deduplicate these 2 files somewhat, but we only want to properly deploy on release, rather than for every PR.

fgrunewald: ok!

fgrunewald (approved):

PR #616, opened 2024-10-09
first draft of martinize2 command tutorials

csbrasnett:
as per #613, some additional pages for the docs about how martinize2 can be used

PR #621, opened 2024-10-21
scfix behaviour change

csbrasnett:
opening this as a PR for changing the default for `-scfix`.

Obviously atm this makes a bunch of tests fail. Is the best way to fix the test to:

1) change the command being used as the input
2) change the expected output of the test

or both?

Do we change the flag description too, to something like -[no]scfix, ie. when the flag is applied scfix gets turned off?


csbrasnett:
Unfortunately, I think we'll have to change (at least some of) the tests. I've noticed a couple of the tests don't actually reflect the default applications. I don't think it should take too long to do :crossed_fingers:

bin/martinize2
```diff
@@ -628,9 +628,9 @@ def entry():
     prot_group.add_argument(
         "-scfix",
         dest="scfix",
-        action="store_true",
+        action="store_false",
         default=True,
-        help="Apply side chain corrections.",
+        help="[don't] apply side chain corrections.",
     )
```
pckroon, line 634: So, the name of the flag remains the same, but the behaviour is inverted? I'm not sure that's very insightful

vermouth/tests/data/integration_tests/tier-0/mini-protein1_betasheet/martinize2/command
```diff
@@ -5,3 +5,5 @@ martinize2
 -ff martini22
 -dssp
 -cys auto
+-ef 2500
```
pckroon, line 8: Typo?

pckroon (changes requested):

bin/martinize2
csbrasnett: change name to `-noscfix`?

pckroon: That would work for me.

pckroon (changes requested):
Yes, excellent!
Just some teenie tiny nitpicks

pckroon:
Probably because 9.0 is not more than 9.0 or something dumb. We need to update the pylint config file as well it seems, lots of deprecated options.
Doesn't really matter at the moment though.
Excellent work!

csbrasnett:
aha thanks! I'll have a look at #616 to check if that's ready to merge too and then we're set

Issue #35, opened 2018-01-17
Naming molecules


jbarnoud:
Writing an ITP file requires the molecules to have names. The current version of martinize call them `Protein_{chain}`, the version with DNA may call them `DNA_{chain}`. But it is not possible to assume we deal with a protein anymore.

Single residue molecules are the simplest case: we can just call them as their residue. Proteins can be identify as collections of residues from a known list; we can have a "detect_protein" processor. Polymers are the worst. Multi-residue lipids are awful too. 

We may need to have something in the force field file, or a generic mechanism to recognize some types of molecules.

PR #623, opened 2024-10-28
Custom molname

csbrasnett:
add a flag to allow users to specify the molname prefix that gets written into files.

This probably solves the problem where you have multiple go models in your system as discussed [here](https://github.com/Martini-Force-Field-Initiative/Martini-Force-Field-Initiative.github.io/issues/41), and potentially #35 too.


csbrasnett:
took longer sorting out the tests than I'd like, but I think this should work.

in short, add `-name` to allow users to specify the prefix name of the molecule to be more customisable rather than just 'molecule_{0..n}'. 

The main other change is then that the `-go-moltype` argument is removed so that the atomnames are taken from the molname automatically. This will then allow users with multiple Go models of different proteins to have them in the same system, so that the additional `[ atomtype ]` and `[ nonbond_params ]` directives don't have overlapping definitions.

vermouth/tests/data/integration_tests/tier-1/lysozyme_GO/martinize2/command
```diff
@@ -8,3 +8,4 @@ martinize2
 -go ../map.map 
 -noscfix
 -cys none
+-name molecule_0
```
pckroon, line 11: shouldn't this be 'molecule'? Also for the second  test

csbrasnett: No, this is what the test expects and I decided to be lazy about it. Before, the molecule name was defaulting to 'molecule_0', so the go_atomtypes get called molecule_0_{1...n}. Because the name now just defaults to 'molecule', giving it as 'molecule_0' here is what's needed to give the same output in the test

pckroon: Does this mean that if I have multiple proteins (=molecules) in my system, named `molecule_{0..n}` that I still end up with overlapping go atomtypes?

csbrasnett: No, because in a Go model the molecules are merged, so you just end up with a single molecule in the end. I guess in a system with multiple proteins that you've martinized with a go model you do, but with this you'd be able to avoid it by having the flexibility to give them different names

pckroon: Ok...
Maybe this is going a bit deep for this PR, but why the merge? Why not make a go model per protein? I realize the merge needs to be there *as an option*, but as default it feels rather heavy handed

csbrasnett: Maybe we're getting terms mixed up? I'm thinking of when the go pipeline is run, the molecules in the system get merged [here](https://github.com/marrink-lab/vermouth-martinize/blob/53c991df953d1e2f2b9763f5c165eab10c32f217/vermouth/rcsu/go_pipeline.py#L40) because that's how the Go model has to work. AFAICT when that happens at the moment that means _all_ molecule names go from `molecule_{0..n}` -> `molecule_0`. 

So, when you then take another protein you want in some complex system, you end up with `go_atomnames.itp` for multiple proteins, with associated `go_nbparams.itp`. However, because in each individual system the atomtypes and nonbonded parameters are notated the same, you end up with conflicting definitions. So if we allow users to define their own prefix for the name, then they can merge systems externally fine

pckroon: > that's how the Go model has to work

I think this is the bit that strikes me, because I don't see this. If I have a pdb with multiple proteins, not in contact, I don't see why they should all get merged. On the other hand, for hetero-dimers the merging does make sense.

In general I may have a different view of how people use martinize2. My idea is that you first produce your atomistic coordinates, containing all your proteins, and then you run the whole system through martinize2.

csbrasnett: ok, I think that explains the confusion! Everyone I've spoken to who makes these massive martini systems first makes their cg models to validate them individually, then combines them in their large system after.

csbrasnett:
Did we reach any consensus on what we wanted to do with this in the end?

pckroon:
You're the closest to the actual users, what do *you* think they want/need?

csbrasnett:
> You're the closest to the actual users, what do _you_ think they want/need?

perfect timing for a group meeting poll tomorrow!

csbrasnett:
The consensus is that this is a good thing!

pckroon:
"this" is a bit ambiguous ;)

csbrasnett:
aha ok, the idea of a `-name` flag to 1) name your molecules as you want and 2) avoid multiple definitions from multiple Go models in your system would be appreciated by the user community

pckroon:
I'm not sure I quite understand what you mean by 2), and how does this relate to the merge all that's currently in the pipeline?

csbrasnett:
If you have 2 different proteins that you've made and validated go models for separately, then at the moment they'll both be called `molecule_0`, with associated `go_atomtypes.itp` that run `molecule_0_1...molecule_0_n`.

If you then build a system that contains both proteins, when it comes to preparing the system for simulation, your top file will at some point need to include _both_ files, which means that you have 1) atom definitions multiply defined, and 2) some apparent nonbonded interactions that shouldn't be there/can't be defined, because there'll be a different number of BB virtual sites on the different proteins.

This PR tries to fix this, so that you give your own prefix for the protein that you're CGing, then for Go models, the virtual site atoms are at least called different things so they can be included multiple times over without error at preprocessing.

For any other protein, it just means that the [ moleculetype ] name is something vaguely more memorable than just `molecule_0`.

I realise this clashes with what you said earlier about CGing a whole atomistic system in one go, but I don't think that's how most people are using martinize

pckroon:
Alright, sounds good.
Is this already implemented/ready for review?

csbrasnett:
yep, should all be contained here!

pckroon (approved):
LGTM!

Issue #620, opened 2024-10-18
write cysteine bridge identification to log


csbrasnett:
currently if a cysteine bridge is identified and the parameters are applied, users have to check for this by hand in the output itp file. It'd be useful to write out which residues have been linked to the log

PR #625, opened 2024-10-28
Log disulfide bridges

csbrasnett:
Small useful feature to write where disulfide bridges have actually been found as per #620. Can't see any effect on time. message might need some work.


pckroon (changes requested):
I like the idea, but I think you can already achieve this by adding a `[ info ]` section to the link:
```
;; Cystein bridge
[ link ]
resname "CYS"
[ constraints ]
SC1 >SC1 1 0.24 {"comment": "Disulfide bridge"}
[ features ]
disulfide
[ info ]
Disulfide bridge found between residues {SC1['chain']}-{SC1['resname']}{SC1['resid']} and {>SC1['chain']}-{>SC1['resname']}{>SC1['resid']}
```
Note that this is an educated guess and YMMV

csbrasnett:
Nice, that worked with a small adjustment thanks!

pckroon (commented):
Awesome. I love preempting requests this way.
Also, could you add it to the other force fields as well?

pckroon (approved):
LGTM!

PR #628, opened 2024-11-04
IDP go interactions and secondary structure

csbrasnett:
This PR fixes two problems with Go models/water biasing as we currently have it, particularly for multi domain proteins.

1) If you assign secondary structure using `-dssp`, and have `-id-regions`, then there's a chance that you'll get folded links written into disordered domains. This is fixed by overwriting the cgsecstruct to be coiled everywhere in disordered domains.

2) Currently if you have disordered and folded regions, the go interactions between them are removed based on the fact that you have water biasing in the disordered region. This requires the `-water-bias` flag to be given. However, if you were to have some disordered domain annotation that didn't require water biasing, then these bonds remain. So, we call the ComputeWaterBias processor in this case, and the bonds are removed through the functionality there.


pckroon:
This PR also needs to be moved to the marrink-lab repo

PR #629, opened 2024-11-04
Pylint update

pckroon:
This updates the pylint rc to the newer version. It's probably easier to review per commit: the first adds the autogenerated new file; I think it took the values from the old file and did the translation. The second is where I make some small changes to the config file. Finally I do some light linting in `bin/martinize2`.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,252 +1,141 @@
-[MASTER]
+[MAIN]
+
+# Analyse import fallback blocks. This can be used to support both Python 2 and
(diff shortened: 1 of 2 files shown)


pckroon:
Oh, this is required before we can mint new releases

csbrasnett (approved):
lgtm, can't see any glaring omissions

fgrunewald (approved):
looks reasonable; I can't find any issue

Issue #642, opened 2024-12-11
handle missing chain indicators in go model better


csbrasnett:
as per [this discussion](https://github.com/orgs/Martini-Force-Field-Initiative/discussions/23), if you have a pdb which doesn't have chain IDs written into it (usually as a result of trjconv?), then the contact map server writes all chains as "Z", which then raises an issue when writing the go model. 

This can usually be solved by going and writing the chains in yourself, regenerating the contact map, etc. But we should probably handle the issue better to indicate this. I'll add it to my 'sorting out the go model list'

csbrasnett:
fixed in v0.13.0

PR #643, opened 2024-12-12
internal go contact map

csbrasnett:
Calculate the contact map for the go model without an externally generated input file.

CLI changes: 
`-go` now accepts _either_ a file in the standard form, or just `-go` to calculate internally.

implementation of the contact map:
- I have tried to do some smart numpy things to speed up the calculation, but I imagine there is still room for improvement.
- Unfortunately I don't think using KDTree for distance cutoffs is possible, because there isn't actually a single cutoff distance, but one that varies atom-atom depending on atomtype. Otherwise I imagine we could speed things up that way.
- This implementation is actually more robust cf the server, because we have a consistent set of atom names we use to calculate the contacts. The server uses default atom names expected from pdbs, which breaks down when people e.g. use AA sim files as inputs, because then these get changed.


vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+def get_vdw_radius(resname, atomname):
+    """
+    get the vdw radius of an atom indexed internally within a serially numbered residue
+    """
+    res_vdw = PROTEIN_MAP[resname]
+    try:
+        atom_vdw = res_vdw[atomname]['vrad']
+    except KeyError:
+        atom_vdw = res_vdw['default']['vrad']
+    return atom_vdw
```
pckroon, line 286:
```suggestion
    res_vdw = PROTEIN_MAP[resname]
    try:
        atom_vdw = res_vdw[atomname]
    except KeyError:
        atom_vdw = res_vdw['default']
    return atom_vdw['vrad']
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+def get_atype(resname, atomname):
+    """
+    get the vdw radius of an atom indexed internally within a serially numbered residue
+    """
+    res_vdw = PROTEIN_MAP[resname]
+    try:
+        atom_vdw = res_vdw[atomname]['atype']
+    except KeyError:
+        atom_vdw = res_vdw['default']['atype']
+    return atom_vdw
```
pckroon, line 298:
```suggestion
    res_vdw = PROTEIN_MAP[resname]
    try:
        atom_vdw = res_vdw[atomname]
    except KeyError:
        atom_vdw = res_vdw['default']
    return atom_vdw['atype']
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+
+
+def make_surface(position, fiba, fibb, vrad):
+    """
+    Generate points on a sphere using Fibonacci points
+
+    position: centre of sphere
+    """
+    x, y, z = position[0], position[1], position[2]
```
pckroon, line 307:
```suggestion
    x, y, z = position
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+        atom_vdw = res_vdw['default']['atype']
+    return atom_vdw
+
+
+def make_surface(position, fiba, fibb, vrad):
+    """
+    Generate points on a sphere using Fibonacci points
+
+    position: centre of sphere
+    """
```
pckroon, line 306: Docstring incomplete

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+    ]
+
+    to get it correct in the 2 dimensional way we actually need.
+
+    At the moment we only need this function once
+    (to get the residue COGs foreach atom)
+    so it's not too much of a limiting factor, but something to optimise
+    better in future
+
+    """
```
pckroon, line 347: That's a comment, not a docstring ;)

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+
+    return out
+
+
+def bondtype(i, j):
+    maxatomtype = 10
+    assert i >= 1
+    assert i <= maxatomtype
+    assert j >= 1
+    assert j <= maxatomtype
```
pckroon, line 384:
```suggestion
    assert 1 <= i <= maxatomtype
    assert 1 <= j <= maxatomtype
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+    # BOND TYPE
+    # Types of contacts:
+    # HB -- 1 -- hydrogen-bond
+    # PH -- 2 -- hydrophobic
+    # AR -- 3 -- aromatic - contacts between aromatic rings
+    # IB -- 4 -- ionic bridge - contacts created by two atoms with different charges
+    # DC -- 5 -- destabilizing contact - contacts which are in general repulsive
+    # OT -- 6 -- denotes negligible other contacts.
+    # 1-HB,2-PH,3-AR,4-IP,5-DC,6-OT
+    btype = np.array([[1, 1, 1, 5, 5, 6, 6, 6, 1, 1],
```
pckroon, line 397: This should probably be a global variable, which also calls the usefulness of this function into question.

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+                      [6, 6, 6, 6, 6, 6, 6, 6, 6, 6],
+                      [1, 1, 5, 5, 5, 6, 6, 6, 5, 4],
+                      [1, 5, 1, 5, 5, 6, 6, 6, 4, 5]])
+    return btype[i][j]
+
+
+def contact_info(system):
+
+    system = MergeAllMolecules().run_system(system)
+    G = make_residue_graph(system.molecules[0])
```
pckroon, line 413: Hmmn. I understand it but I still don't like it. Maybe do the merge in the martinize2 script (or GoPipeLine), and have contact_info run on molecules? I think that'll make life easier later

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+    for node in G.nodes:
+        # we only need these for writing at the end
+        resnames.append(G.nodes[node]['resname'])
+        resids.append(G.nodes[node]['resid'])
+        chains.append(G.nodes[node]['chain'])
+        nodes.append(G.nodes[node]['_res_serial'])
+
+        res_pos = []
+        for subnode in sorted(G.nodes[node]['graph'].nodes):
+            if 'position' in G.nodes[node]['graph'].nodes[subnode]:
```
pckroon, line 434: Maybe also filter for NaN?

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+    nodes = []
+    for node in G.nodes:
+        # we only need these for writing at the end
+        resnames.append(G.nodes[node]['resname'])
+        resids.append(G.nodes[node]['resid'])
+        chains.append(G.nodes[node]['chain'])
+        nodes.append(G.nodes[node]['_res_serial'])
+
+        res_pos = []
+        for subnode in sorted(G.nodes[node]['graph'].nodes):
```
pckroon, line 433: Maybe change node and subnode into residue and atom?

pckroon, line 433: I suggest not sorting, and just building a dict of graph_idx: array_idx

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+
+                vdw_list.append(get_vdw_radius(G.nodes[node]['graph'].nodes[subnode]['resname'],
+                                               G.nodes[node]['graph'].nodes[subnode]['atomname']))
+                atypes.append(get_atype(G.nodes[node]['graph'].nodes[subnode]['resname'],
+                                        G.nodes[node]['graph'].nodes[subnode]['atomname']))
+
+            if G.nodes[node]['graph'].nodes[subnode]['atomname'] == 'CA':
+                ca_pos.append(G.nodes[node]['graph'].nodes[subnode]['position'])
+
+        cogs.append(np.stack(np.tile(np.stack(res_pos).mean(axis=0), (len(res_pos), 1))))
```
pckroon, line 448: This needs a comment

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+    resnames = np.array(resnames)
+    nodes = np.array(nodes)
+
+    # 2) find the number of residues that we have
+    nresidues = len(G)
+
+    return cogs, vdw_list, atypes, coords, res_serial, resids, chains, resnames, nodes, ca_pos, nresidues, G
+
+
+def calculate_contact_map(cogs, vdw_list, atypes, coords, res_serial,
```
pckroon, line 467: docstring

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,610 @@
...
+
+    # 4) set up final bits of information
+    # sums of vdw pairs for each atom we have
+    vdw_sum_array = vdw_list[:, np.newaxis] + vdw_list[np.newaxis, :]
+
+    # distances between all the atoms that we have
+    atomic_distances = cdist(coords, coords)
+
+    # array with 1 on the diagonal, so we can exclude the self atoms
+    diagonal_ones = np.diagflat(np.ones(atomic_distances.shape[0], dtype=int))
```
pckroon, line 488:
```suggestion
    diagonal_ones = np.diagflat(np.full(atomic_distances.shape[0], np.inf))
```
Maybe? Saves a check later on. But I think this goes away with the kdtree anyway

pckroon (changes requested):
I'll review more thoroughly once v2 is there ;)
Love the feature

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,653 @@
...
+                b = np.where(nodes == i2)[0][0]
+                if over == 1 or (over == 0 and rcsu == 1):
+                    # this is a OV or rCSU contact we take it
+                    contacts_list.append((int(G.nodes[a]['resid']), G.nodes[a]['chain'],
+                                          int(G.nodes[b]['resid']), G.nodes[b]['chain']))
+
+    return contacts_list
+
+
+def write_contacts(nresidues, overlaps, contacts, stabilisers, destabilisers, nodes, ca_pos, G):
```
pckroon, line 629: This function repeats a lot of code from get_contacts. Can't you use the contacts list that generates as input here?

vermouth/rcsu/contact_map.py
```diff
@@ -49,5 +692,42 @@ def read_go_map(file_path):
...
+                                                                                coords,
+                                                                                res_serial,
+                                                                                nresidues)
+
+            self.system.go_params["go_map"].append(get_contacts(nresidues,
+                                                                overlaps, contacts,
+                                                                stabilisers,
+                                                                destabilisers,
+                                                                nodes,
+                                                                G))
```
pckroon, line 731: Wrap all this into a single function.

pckroon (changes requested):
It's starting to shape up very nicely! Awesome work so far.
For the next pass, besides the comments, could you also have a look at the functions you create, and decide whether those will have a use beyond the go map? If not, mark them as private (prefix a `_`).
Some functions also need a lot of arguments, and return a lot of different things. Do you think there's any chance we could simplify that?

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,653 @@
...
+    norm: bool
+        if True, then any entry > 0 in the summed array = 1
+    """
+    out = np.array([int(arrin[np.where(residues == i)[0], np.where(residues == j)[0][:, np.newaxis]].sum())
+                    for i in range(nresidues)
+                    for j in range(nresidues)]).reshape((nresidues, nresidues))
+    if norm:
+        out[out > 0] = 1
+
+    return out
```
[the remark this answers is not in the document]
pckroon: I'm also not sure how sparse `out` is here. Maybe that should also be a dictionary.

vermouth/rcsu/contact_map.py
```diff
@@ -12,10 +12,653 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ..processors.processor import Processor
+import numpy as np
+from scipy.spatial.distance import euclidean, cdist
```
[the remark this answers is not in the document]
csbrasnett: I'll remove the cdist import, the euclidean is needed in case you want to write the contact map to file

csbrasnett:
That should be all the more minor points addressed, the main one being how to improve the atom2res function. (had to push this now because my laptop keeps losing the .git in my clone somehow?! feel free to ignore until I've addressed everything) 

We have to use `itertools.product` rather than `itertools.combinations` because we can't assume that the array is symmetric. But the speedup is sufficient that this is insignificant in comparison.

I've had a look at using dictionaries instead for the atomic resolution arrays. I'm still not sure it's possible/advisable with the way I've currently written it? Or at least, I'd rather get all the other jobs above sorted and revisit that later if we think there might be substantial advantage. I agree, the arrays are pretty sparse - but they aren't completely, which is slightly annoying.

csbrasnett:
I think that's as much as we need for now. afaict the priority list is now:

1) Write a header for the contact_map_vermouth.out file
2) deal with #585
3) deal with #642
4) (and if we want to before completing this implementation) change the contact arrays to dictionaries for further speed improvement

bin/martinize2
```diff
@@ -967,6 +989,17 @@ def entry():
...
 
+    if go:
+        system = vermouth.MergeAllMolecules().run_system(system)
+        # need this here because have to get contact map at atomistic resolution
+        if go_file is None:
+            LOGGER.info("Generating Go model contact map.", type="step")
+            GenerateContactMap(write_file=args.go_write_file).run_system(system)
+        else:
+            LOGGER.info("Reading Go model contact map.", type="step")
+            read_go_map(system=system, path=go_file)
```
pckroon, line 1000: So, something like this
```suggestion
    if args.go:
        system = vermouth.MergeAllMolecules().run_system(system)
        # need this here because have to get contact map at atomistic resolution
        if isinstance(args.go, str):
            LOGGER.info("Reading Go model contact map.", type="step")
            read_go_map(system=system, path=args.go)
        else:
            LOGGER.info("Generating Go model contact map.", type="step")
            GenerateContactMap(write_file=args.go_write_file).run_system(system)
```

bin/martinize2
```diff
@@ -988,14 +1021,12 @@ def entry():
         vermouth.ApplyPosres(node_selector, args.posres_fc).run_system(system)
 
     # Generate the Go model if required
-    if args.go:
+
+    if go:
```
pckroon, line 1025: I suggest checking whether the system is annotated

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+        atom_idxs = np.array(atom_map[res_idx])
+        atom_jdxs = np.array(atom_map[res_jdx])
+        value = arrin[atom_idxs,
+        atom_jdxs[:, np.newaxis]].sum()
+        out[res_idx, res_jdx] = value
+
+    if norm:
+        out[out > 0] = 1
+
+    return out
```
pckroon, line 368: Don't make np arrays where you don't need them. The atom_map is a dict of resid: tuple of node indices?
```suggestion
    nresidues = len(atom_map)  # Right?
    out = np.zeros((nresidues, nresidues))
    for res_idx, res_jdx in product(atom_map.keys(), atom_map.keys()):
        atom_idxs = atom_map[res_idx]
        atom_jdxs = atom_map[res_jdx]
        value = arrin[atom_idxs, atom_jdxs].sum()
        out[res_idx, res_jdx] = value

    if norm:
        out[out > 0] = 1

    return out
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+    vdw_list = []
+    atypes = []
+    res_serial = []
+    nodes = []
+    for residue in G.nodes:
+        # we only need these for writing at the end
+        resnames.append(G.nodes[residue]['resname'])
+        resids.append(G.nodes[residue]['resid'])
+        chains.append(G.nodes[residue]['chain'])
+        nodes.append(G.nodes[residue]['_res_serial'])
```
pckroon, line 392: ? nodes is res_serial?

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+        subgraph = G.nodes[residue]['graph']
+
+        for atom in sorted(G.nodes[residue]['graph'].nodes):
```
pckroon, line 395:
```suggestion
        for atom in sorted(subgraph.nodes):
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+
+                positions_all.append(subgraph.nodes[atom]['position'] * 10)
+
+                vdw_list.append(_get_vdw_radius(subgraph.nodes[atom]['resname'],
+                                                subgraph.nodes[atom]['atomname']))
+                atypes.append(_get_atype(subgraph.nodes[atom]['resname'],
+                                         subgraph.nodes[atom]['atomname']))
+
+            if subgraph.nodes[atom]['atomname'] == 'CA':
+                ca_pos.append(subgraph.nodes[atom]['position'])
```
pckroon, line 408:
```suggestion
                if subgraph.nodes[atom]['atomname'] == 'CA':
                    ca_pos.append(subgraph.nodes[atom]['position'])
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+    chains = np.array(chains)
+    resnames = np.array(resnames)
+    nodes = np.array(nodes)
+
+    # 2) find the number of residues that we have
+    nresidues = len(G)
+
+    return vdw_list, atypes, coords, res_serial, resids, chains, resnames, nodes, ca_pos, nresidues, G
+
+def _calculate_overlap(coords_tree, vdw_list, natoms, vdw_max, alpha):
```
pckroon, line 426:
```suggestion
def _calculate_overlap(coords_tree, vdw_list, natoms, vdw_max, alpha=1.24):
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+    """
+    over = np.zeros((natoms, natoms))
+    over_sdm = coords_tree.sparse_distance_matrix(coords_tree, 2 * vdw_max * alpha)
+    for (idx, jdx), distance_between in over_sdm.items():
+        if idx != jdx:
+            if distance_between < (vdw_list[idx] + vdw_list[jdx]) * alpha:
+                over[idx, jdx] = 1
+    return over
+
+def _calculate_csu(coords, vdw_list, fiba, fibb, natoms, coords_tree, vdw_max, water_radius):
```
pckroon, line 449:
```suggestion
def _calculate_csu(coords, vdw_list, fiba, fibb, natoms, coords_tree, vdw_max, water_radius=2.80):
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+
+    # Calculate the CSU contacts of the molecule
+    hit_results = _calculate_csu(coords, vdw_list, fiba, fibb, natoms, coords_tree, vdw_max, water_radius=2.80)
+
+    # find the types of contacts we have
+    contactcounter_1, stabilisercounter_1, destabilisercounter_1 = _contact_types(hit_results, natoms, atypes)
+
+    atom_map = {}
+    for i in range(nresidues):
+        atom_map[i] = np.where(res_serial == i)[0]
```
pckroon, line 584:
```suggestion
    atom_map = defaultdict(list)
    for atom_idx, res_idx in enumerate(res_serial):
        atom_map[res_idx].append(atom_idx)
```

vermouth/rcsu/contact_map.py
```diff
@@ -12,19 +12,672 @@
...
+
+    all_contacts: list
+        list of lists of every contact found
+    ca_pos: list
+        list of (3,) arrays with the position of the CA atom of each residue
+    G: nx.Graph
+        residue graph of the input molecule
+    '''
+    # this to write out the file if needed
+    with DeferredFileWriter.open(Path('contact_map_vermouth.out'), 'w') as f:
```
pckroon: Should the file path we're writing to an argument for this function?

vermouth/tests/rcsu/test_contact_map.py
```diff
@@ -144,3 +153,293 @@ def test_contact_info(test_molecule):
...
+    water_radius = 1
+
+    hits = contact_map._calculate_csu(points,
+                                              vdw_list,
+                                              fiba,
+                                              fibb,
+                                              natoms,
+                                              tree,
+                                              vdw_max,
+                                              water_radius)
```
pckroon, line 237: indentation

vermouth/tests/rcsu/test_contact_map.py
```diff
@@ -144,3 +153,293 @@ def test_contact_info(test_molecule):
...
+        written_lines = infile.readlines()
+
+    # skip the first line here because it's the vermouth version
+    for line, expected_line in zip(written_lines[1:], expected_lines[1:]):
+        assert line == expected_line
+
+
+@pytest.mark.parametrize('write_out',
+                         ((False),
+                          (True))
```
pckroon, line 397:
```suggestion
                         (False, True)
```

pckroon (changes requested):
Excellent work. I think I found a way to clean up the CLI definition a bit, and a few tiny nitpicks otherwise. Please also do a search/replace for Go/Gō in the documentation. I think I got them all, but who knows.

pckroon (approved):
Amazing work <3

PR #647, opened 2025-01-09
Martini3-IDP

csbrasnett:
This PR:

1) Adds the IDP force field of @WangLiguo-kyrie as per [the parameters](https://github.com/Martini-Force-Field-Initiative/Martini3-IDP-parameters)
2) Makes some small edits for Go models to ensure that for multi domain proteins Go bonds applied between folded-disordered pairs are removed.
3) For IDPs in general, we overwrite assigned secondary structure as "C" to ensure that we're not applying IDP parameters to folded regions by accident.

Note that the martini3-IDP/aminoacids.ff file is slightly different here to ensure that it works properly with martinize.

Will need to update citations.bib once the paper is accepted


bin/martinize2
```diff
@@ -977,6 +970,17 @@ def entry():
...
+    ss_sequence = list(
+        itertools.chain(
+            *(
+                SS_CG[i]
+                for i in dssp.sequence_from_residues(molecule, "cgsecstruct")
+                for molecule in system.molecules
+                if selectors.is_protein(molecule) if i is not None
+            )
+        )
+    )
```
pckroon, line 982: This translation to CG is too simplistic, no?

```suggestion
    ss_sequence = list(
        itertools.chain(
            *(
                dssp.convert_dssp_to_martini(dssp.sequence_from_residues(molecule, "cgsecstruct"))
                for molecule in system.molecules
                if selectors.is_protein(molecule) if i is not None
            )
        )
    )
```

Also, how does this interact with the "normal" dssp processing, and people providing a secondary structure on the cli?

vermouth/processors/annotate_idrs.py
```diff
@@ -39,6 +39,10 @@ def annotate_disorder(molecule, id_regions, annotation="cgidr"):
         _old_resid = node['_old_resid']
         if _in_resid_region(_old_resid, id_regions):
             molecule.nodes[key][annotation] = True
+            if "cgsecstruct" in molecule.nodes[key]:
+                molecule.nodes[key]["cgsecstruct"] = "C"
+        else:
+            molecule.nodes[key][annotation] = False
```
pckroon, line 45:
```suggestion
            if "cgsecstruct" in molecule.nodes[key]:
                molecule.nodes[key]["cgsecstruct"] = "C"
            molecule.nodes[key][annotation] = True  # ???
        else:
            molecule.nodes[key][annotation] = False
```

pckroon (changes requested):
Some small questions. Note that I didn't look at the ff files

bin/martinize2
csbrasnett: This won't affect the "normal" processing, the annotation is done for both ways before martinize() is called. The dssp/ss flags are dealt with at that point, and this is only to make sure that the cg secondary structure is determined correctly. The ss_sequence needs to be moved here to align with the overwriting that happens in the idr annotation, otherwise the 'uncorrected' string is written out

csbrasnett:
also, I'm assuming that patch coverage is low because of the ff file addition? I can't otherwise see anything more that needs to have tests added?

bin/martinize2
pckroon: My concern with this is that it becomes ambiguous what's atomistic SS and whats CG SS. The idea of printing the found SS structure is so that users can pass it to the `-ss` flag to reproduce the behaviour. However, if that only works if the SS flavour that's expected is the same as the one printed.

csbrasnett: Point taken. However, for the purposes of the current IDP models, ensuring the ss used when the links are applied doesn't intersect in a nasty way with the -idr flags is fairly important.

Perhaps the way around is if the string gets changed, print it in the itp twice? first what was found by dssp, second what was actually applied in the links?

pckroon: That sounds reasonable

csbrasnett:
checking ff files, hold off on merging

bin/martinize2
```diff
@@ -1037,6 +1036,12 @@ def entry():
         defines = ("GO_VIRT",)
         itp_paths = {"atomtypes": "go_atomtypes.itp",
                      "nonbond_params": "go_nbparams.itp"}
+        if not args.water_bias:
+            # this ensures that disordered-folded go bonds get removed regardless of force field.
+            vermouth.processors.ComputeWaterBias(args.water_bias,
+                                                 {s: float(eps) for s, eps in args.water_bias_eps},
+                                                 [(int(start), int(stop)) for start, stop in args.water_idrs],
+                                                ).run_system(system)
```
pckroon, line 1044: Do `args.water_bias_eps` and `args.water_idrs` have sane defaults?

bin/martinize2
```diff
@@ -1167,6 +1172,25 @@ def entry():
...
+    if any([molecule.meta.get('modified_cgsecstruct', False) for molecule in system.molecules]):
+        LOGGER.info(("Secondary structure assignment changed between dssp and martinize. "
+                     "Check files for details."), type="general")
+        header += [
+            "The assigned secondary structure conflicted with ",
+            "annotated IDRs. The following sequence of Martini secondary ",
+            "structure was actually applied to the system:",
+            "".join([dssp.SS_CG[i] for i in supp_ss_seq])
+        ]
+
```
pckroon, line 1193: Any chance we can fold this into an (existing) processor?

bin/martinize2
```diff
@@ -1167,6 +1172,25 @@ def entry():
...
+        )
+    )
+    if any([molecule.meta.get('modified_cgsecstruct', False) for molecule in system.molecules]):
+        LOGGER.info(("Secondary structure assignment changed between dssp and martinize. "
+                     "Check files for details."), type="general")
+        header += [
+            "The assigned secondary structure conflicted with ",
+            "annotated IDRs. The following sequence of Martini secondary ",
+            "structure was actually applied to the system:",
+            "".join([dssp.SS_CG[i] for i in supp_ss_seq])
```
pckroon, line 1191: Can you explain why the `dssp.SS_CG[i]`, rather than calling `convert_dssp_to_martini`? The latter does a more complex around helices

pckroon (changes requested):
Some small questions, looks good otherwise

bin/martinize2
csbrasnett: False and [] respectively.

bin/martinize2
csbrasnett: I thought about this, I'm not sure where it could go to then be written as an additional header?

bin/martinize2
csbrasnett: The change in the annotation is made to `cgsecstruct` not the `secstruct` one, which is needed for it to be picked up when the links are applied. doing `convert_dssp_to_martini(i) will then return the sequence converting the already converted cg sequence into garbage that's not useful for rerunning martinize

bin/martinize2
pckroon: `s: float(eps) for s, eps in args.water_bias_eps` will cause an error then?

bin/martinize2
pckroon: Interesting point. Should Systems collect headers which the topology writer than picks up? Maybe in a `meta` attribute similar to Molecules?

bin/martinize2
pckroon: Check, clear.
I think I would appreciate some different variables names around though. `supp_ss_seq` doesn't mean too much to me, and maybe `secstruct` should be `aasecstruct` or something along those lines to contrast it to `cgsecstruct`

bin/martinize2
csbrasnett: ah wait, `args.water_bias_eps` and `args.water_idrs`both default to [], it's `args.water_bias` that defaults to []. As the comment says, this is intended to run empty so that cross domain Go bonds are removed. It can be changed to be explicitly empty though

bin/martinize2
pckroon: That's very recognizable, and I agree with splitting it off into a separate PR

vermouth/processors/annotate_idrs.py
```diff
@@ -39,6 +39,12 @@ def annotate_disorder(molecule, id_regions, annotation="cgidr"):
         _old_resid = node['_old_resid']
         if _in_resid_region(_old_resid, id_regions):
             molecule.nodes[key][annotation] = True
+            if "cgsecstruct" in molecule.nodes[key]:
+                if molecule.nodes[key]["cgsecstruct"] != "C":
+                    molecule.nodes[key]["cgsecstruct"] = "C"
+                    molecule.meta['modified_cgsecstruct'] = True
```
pckroon, line 45:
```suggestion
            if "cgsecstruct" in molecule.nodes[key] and molecule.nodes[key]["cgsecstruct"] != "C":
                    molecule.nodes[key]["cgsecstruct"] = "C"
                    molecule.meta['modified_cgsecstruct'] = True
```

pckroon (changes requested):
Looking great! Small nitpick from my side, but fine otherwise. Is this PR finished/complete like this?

csbrasnett:
The only outstanding thing would be a full citation for the force field when it comes, I'll check if we want to merge now or hold off

csbrasnett:
@pckroon ready for merge now

PR #649, opened 2025-01-12
sort before dssp

fgrunewald:


pckroon:
Maybe it's cleaner to extract the functionality in `SortMoleculeAtoms.run_molecule` to a separate function? I'm not sure it makes a big difference tbh

fgrunewald:
@pckroon perhaps but I think for now this current fix works well enough. For the new protein model we may to have to reexamine the DSSP workflow. Perhaps we fix this issue then

pckroon (approved):

PR #655, opened 2025-01-31
Add non-forcefield-specific citations to the root of the data directory

pckroon:
This is particularly useful for *code* that needs citing, such as MDTraj, and vermouth.

The vermouth citation gets formatted as
```
    INFO - general - Please cite: Kroon, P C; Grunewald, F; Barnoud, J; van Tilburg, M; Souza, P C; Wassenaar, T A; Marrink, S J;  elife 2024; 10.7554/elife.90627.2
```

Things got a bit more messy because of Molecules and Systems without forcefield (or, `force_field=None`), so I also did something sane with the defaults there.

### DIFF
--- a/vermouth/gmx/topology.py
+++ b/vermouth/gmx/topology.py
@@ -2,13 +2,14 @@
 I/O of topology parameters that are not molecules.
 """
 import itertools
-from collections import namedtuple
+from collections import namedtuple, ChainMap
 import textwrap
 import vermouth
 from vermouth.file_writer import deferred_open
 from vermouth.citation_parser import citation_formatter
 from ..log_helpers import StyleAdapter, get_logger
 from .itp import _interaction_sorting_key
+from ..data import COMMON_CITATIONS
 
 LOGGER = StyleAdapter(get_logger(__name__))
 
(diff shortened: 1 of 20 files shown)


pckroon:
I'm not going to meet the patch test coverage; it's mostly because there's no tests for the Quote processor (and I don't care about that), and there's no tests for `write_gmx_topology` (and I don't have time to make those)

vermouth/gmx/topology.py
```diff
@@ -179,6 +180,10 @@ def write_gmx_topology(system,
     )
     for moltype, molecules in molecule_groups:
         molecule = next(molecules)
+        if molecule.force_field is not None:
+            citation_map = ChainMap(molecule.force_field.citations, COMMON_CITATIONS)
+        else:
+            citation_map = COMMON_CITATIONS
```
csbrasnett, line 186: Does this mean that the MDTraj citation is now added to everything even if it's not used?

csbrasnett (commented):
lgtm, just one query about how mdtraj citations are now treated, because I think this now has the citation written regardless of whether it was used?

vermouth/gmx/topology.py
pckroon: No, only that you can add the mdtraj citation to molecules of any forcefield, even if that forcefield doesn't specify the citation.

csbrasnett: Got it, thanks

csbrasnett (approved):

PR #658, opened 2025-02-07
move header writing in gmx files to a system level meta attribute

csbrasnett:
instead of collecting the header for gmx files as a list, move it to a system.meta["header"] list that can be more dynamically added to

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -912,19 +912,6 @@ def entry():
         write_canon=args.write_canon,
     )
 
-    # grompp has a limit in the number of character it can read per line
-    # (due to the size limit of a buffer somewhere in its implementation).
-    # The command line can be longer than this limit and therefore
-    # prevent grompp from reading the topology.
-    gromacs_char_limit = 4000  # the limit is actually 4095, but I play safe
-    command = " ".join(sys.argv)
-    if len(command) > gromacs_char_limit:
-        command = command[:gromacs_char_limit] + " ..."
-    system.meta["header"].extend(("This file was generated using the following command:",
-                                  command,
-                                  VERSION,
-                                 ))
-
     LOGGER.info("Read input.", type="step")
     for molecule in system.molecules:
         LOGGER.debug("Read molecule {}.", molecule, type="step")
@@ -1156,7 +1143,9 @@ def entry():
                            args.top_path,
                            itp_paths=itp_paths,
                            C6C12=False,
-                           defines=defines)
+                           defines=defines,
+                           command=sys.argv,
+                           version=VERSION)
 
     # Write a PDB file.
     vermouth.pdb.write_pdb(system, str(args.outpath), omit_charges=True)
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -911,7 +911,10 @@ def entry():
         write_repair=args.write_repair,
         write_canon=args.write_canon,
     )
-
+    system.meta["header"].extend(("This file was generated using the following command:",
+                                  " ".join(sys.argv),
+                                  VERSION,
+                                  ))
     LOGGER.info("Read input.", type="step")
     for molecule in system.molecules:
         LOGGER.debug("Read molecule {}.", molecule, type="step")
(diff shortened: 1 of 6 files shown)


vermouth/gmx/topology.py
```diff
@@ -178,6 +200,9 @@ def write_gmx_topology(system,
     molecule_groups = itertools.groupby(
         system.molecules, key=lambda x: x.meta["moltype"]
     )
+
+    header = system_header(system)
```
pckroon, line 204:
```suggestion
    header = system.meta.get('header', [])
```

pckroon (changes requested):
Nice work.
I do think there's some snakes in the grass where the headers of separate molecules start to affect each other though. https://github.com/marrink-lab/vermouth-martinize/pull/658/files#diff-4071134fd6f80afcd683f455e2558cc2fbd683b801fe6bc78ebc28fb86cdf655R225 modifies the header for the entire system, rather than just the one molecule.
Not sure that's a real issue.

csbrasnett:
Moved things around. Moving the secondary structure part to dssp made a slight mess because it then wrote that ahead of the command used, which IMO is important to keep as the first line in the itp file.

bin/martinize2
```diff
@@ -912,6 +912,19 @@ def entry():
...
+    # prevent grompp from reading the topology.
+    gromacs_char_limit = 4000  # the limit is actually 4095, but I play safe
+    command = " ".join(sys.argv)
+    if len(command) > gromacs_char_limit:
+        command = command[:gromacs_char_limit] + " ..."
+    system.meta["header"].extend(("This file was generated using the following command:",
+                                  command,
+                                  VERSION,
+                                 ))
+
```
pckroon, line 927: The is gromacs specific, and should be moved to the topology writer

vermouth/dssp/dssp.py
```diff
@@ -570,6 +589,9 @@ def run_molecule(molecule):
         convert_dssp_annotation_to_martini(molecule)
         return molecule
 
+    def run_system(self, system):
+        gmx_system_header(system)
+
```
pckroon:
```suggestion
    def run_system(self, system):
        gmx_system_header(system)
        super().run_system(system)
```

pckroon (changes requested):
Small comments

vermouth/tests/test_dssp.py
```diff
@@ -705,3 +706,28 @@ def test_cterm_atomnames():
...
+    system = create_sys_all_attrs(test_molecule,
+                                  moltype="molecule_0",
+                                  secstruc=secstruc,
+                                  defaults={"chain": "A"},
+                                  attrs={"resname": resnames,
+                                         "atype": atypes})
+
+    dssp.AnnotateMartiniSecondaryStructures().run_system(system)
+
+    assert system.meta.get('header')
```
pckroon, line 733: Check at least the length of the header, and maybe even its contents

pckroon (changes requested):

vermouth/dssp/dssp.py
```diff
@@ -554,7 +554,7 @@ def gmx_system_header(system):
         )
     )
 
-    if None not in ss_sequence:
+    if (None not in ss_sequence) and (len(ss_sequence) > 0):
```
pckroon, line 557: nit
```suggestion
    if None not in ss_sequence and ss_sequence:
```

bin/martinize2
```diff
@@ -912,19 +912,6 @@ def entry():
...
-    # The command line can be longer than this limit and therefore
-    # prevent grompp from reading the topology.
-    gromacs_char_limit = 4000  # the limit is actually 4095, but I play safe
-    command = " ".join(sys.argv)
-    if len(command) > gromacs_char_limit:
-        command = command[:gromacs_char_limit] + " ..."
-    system.meta["header"].extend(("This file was generated using the following command:",
-                                  command,
-                                  VERSION,
-                                 ))
```
pckroon: I should've been more clear I think: add the full string to the header list here, but do the length limiting in the topology writer (for *all* headers/comments)

pckroon (changes requested):

pckroon (dismissed):
LGTM!

csbrasnett:
Thanks, working on fixing the test and there's something very weird happening with the line length stuff... will be done eod...

pckroon (approved):
Awesome

PR #659, opened 2025-02-14
change posres_fc format

csbrasnett:
as discussed in #639 

- write global #ifndef at top of itps
- change value of posres_fc to global
- add some tests for the posres processor
- add documentation page for things vermouth writes that are applied via gromacs preprocessor


csbrasnett:
the tests are failing because the itp reader can't handle `#define` statements: 

```
Don't know how to parse pargma #define POSRES_FC 1000 at line 14.
```

so that needs to be implemented somewhere in the file parser if we want to solve this

pckroon:
I think your analysis is spot on. Maybe you can even ignore the fact that its in an ifdef/ifndef block? Implementing that in a generic way gets quite gnarly I think.

csbrasnett:
That seems to work nicely now, and I've double checked that output files run as expected. Have had to hack the tests slightly to ensure they still work which is not ideal, but if we're going to ignore `#define` sections in this way I don't think we have an option to compare properly?

Also, I think the docs page was a good idea, but not sure if it's actually in the right place at the moment so happy to move it wherever. Didn't feel like it sat right alongside the other tutorials though?

vermouth/tests/integration_tests/test_integration.py
```diff
@@ -159,7 +159,12 @@ def _interaction_equal(interaction1, interaction2):
     interaction parameters.
     """
     p1 = list(map(float, interaction1.parameters))
-    p2 = list(map(float, interaction2.parameters))
+    try:
+        p2 = list(map(float, interaction2.parameters))
+    except ValueError:
+        if interaction2.parameters[1] == 'POSRES_FC':
+            p2 = list(map(float, ['1', '1000', '1000', '1000']))
```
pckroon, line 166: Why? Add a comment please

pckroon (changes requested):

doc/source/gromacs_variables.rst
```diff
@@ -0,0 +1,43 @@
...
+`preprocessing <https://manual.gromacs.org/documentation/current/user-guide/mdp-options.html#preprocessing>`_.
+
+Preprocessing statements in gromacs mdp files may be used to customise topologies at the preprocessing stage
+of simulation setup. The statements in the ``define`` statement of the input molecular dynamics parameters
+(.mdp) file are passed to the topology file, where ``#ifdef`` statements are used to modify the exact topology
+of the molecule. For more examples of ``#ifdef`` statements, see the
+`gromacs manual <https://manual.gromacs.org/2024.2/reference-manual/topologies/topology-file-formats.html#ifdef-statements>`_.
+
+Note that the ``define`` statement in the .mdp file is only given once, with arguments delineated by the ``-D``
+prefix on the same line.
```
pckroon, line 14: Maybe add an example for this? `define = -DFLEXIBLE -DPOSRES_FC=1000`

doc/source/index.rst
```diff
@@ -16,6 +16,7 @@ Welcome to VerMoUTH's documentation!
    file_formats
    tutorials/index
    api/modules
+   gromacs_variables
```
pckroon, line 19: Maybe between file formats and the tutorials is a better spot? At least above the api docs

vermouth/gmx/itp_read.py
```diff
@@ -145,13 +145,15 @@ def parse_pragma(self, line, lineno=0):
                 condition, tag = line.split()
                 self.current_meta = {'tag': tag, 'condition': condition.replace("#", "")}
             elif self.current_meta is not None:
-                raise IOError("Your #ifdef/#ifndef section is orderd incorrectly."
+                raise IOError("Your #ifdef/#ifndef section is ordered incorrectly."
                               "At line {} I read {} but there is still"
                               "an open #ifdef/#ifndef section from"
                               "before.".format(lineno, line.split()[0]))
+        elif line.startswith("#define"):
+            pass
```
pckroon, line 153: Add a comment why `pass`, rather than setting the molecule meta

vermouth/processors/apply_posres.py
```diff
@@ -18,13 +18,14 @@
 def apply_posres(molecule, selector, force_constant, functype=1, ifdef='POSRES'):
     for key, node in molecule.nodes.items():
         if selector(node):
-            parameters = [functype, ] + [force_constant, ] * 3
+            parameters = [functype, ] + ["POSRES_FC", ] *3
```
pckroon, line 21:
```suggestion
            parameters = [functype, ] + ["POSRES_FC", ] * 3
```

pckroon (changes requested):
Very nice. Just a few small comments

vermouth/gmx/itp_read.py
csbrasnett: do you mean the current_meta?

pckroon: Probably, yes

pckroon:
Could you try to figure out the failing docs build? You can test it locally before you push it. You can find the exact sphinx invocation in the CI config file (https://github.com/marrink-lab/vermouth-martinize/blob/master/.github/workflows/run_tests.yml#L117). I don't have time available to have a look today :(

pckroon (approved):

PR #661, opened 2025-02-19
updating conf.py for sphinx

csbrasnett:
atm, partial fix until we figure out the problem with `napoleon_type_aliaes`

- remove sphinxcontrib-apidoc now apidoc is inbuilt sphinx.ext
- update conf.py as required

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -51,7 +51,7 @@ extensions = [
     'sphinx.ext.viewcode',
     'sphinx.ext.napoleon',
     'sphinx.ext.autosectionlabel',
-    'sphinxcontrib.apidoc',
+    'sphinx.ext.apidoc',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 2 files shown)


csbrasnett:
@pckroon if we want to fix this in the short term, either we:

1. change `sphinx-build -EnW` to `sphinx-build -En` so that the build doesn't fail, or
2. we just enforce `sphinx < 2.8.0` in requirements-docs.txt, and drop the update to `sphinx.ext.apidocs` for now until sphinx is updated

If we do the former, I've been assured that the documentation should still render correctly for now

pckroon:
Option 3, set napoleon_type_aliases to None. Your call.

csbrasnett:
won't that break the correct references to the `vermouth.molecule.Molecule`?

pckroon:
Seemed to work when I quickly tried it here, but YMMV

pckroon (approved):
Seems to work :+1:

pckroon:
I'll merge this now, could you create an issue so we can remember putting the napoleon aliases back later, referencing the upstream issue?

PR #663, opened 2025-02-24
Rtp patch

fgrunewald:
Patch to make the rtp parser play nice with modifications. Essentially, we need to add updated resids to modification atoms. I also added a fail-safe in molecule resid assignment upon merging.

### DIFF
--- a/vermouth/processors/do_mapping.py
+++ b/vermouth/processors/do_mapping.py
@@ -408,7 +408,6 @@ def apply_mod_mapping(match, molecule, graph_out, mol_to_out, out_to_mol):
     mol_to_mod, modification, references = match
     LOGGER.info('Applying modification mapping {}', modification.name, type='general')
     graph_out.citations.update(modification.citations)
-
     mod_to_mol = defaultdict(dict)
     for mol_idx, mod_idxs in mol_to_mod.items():
         for mod_idx in mod_idxs:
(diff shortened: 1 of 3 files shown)


vermouth/processors/do_mapping.py
```diff
@@ -465,6 +464,21 @@ def apply_mod_mapping(match, molecule, graph_out, mol_to_out, out_to_mol):
...
+            continue
+        neighbors = modification.neighbors(mod_idx)
+        for mod_idx_neigh in neighbors:
+            anchor = mod_to_out[mod_idx_neigh]
+            resid_new = graph_out.nodes[anchor].get('resid', None)
+            resid_old = graph_out.nodes[anchor].get('_old_resid', None)
+            if resid_new:
+                graph_out.nodes[out_idx]['_old_resid'] = resid_old
+                graph_out.nodes[out_idx]['resid'] = resid_new
+                break
```
pckroon, line 480: Does it work if you add "resid" to the `attributes_must` in martinize2? https://github.com/marrink-lab/vermouth-martinize/blob/master/bin/martinize2#L178 IIRC, that means that if the attribute is missing it'll recursively find the nearest node that has, and use that.

fgrunewald: no because you are not guaranteed to pick it up from the actual residue; you might as well get it from the next one

pckroon: True. OTOH it'll usually work, because the closest atom will probably be the same residue. And, the solution here will only look at the direct neighbours, so it'll break if you have a multi-atom modification

fgrunewald: You're right but I don't want to track down this bug in a year from now again.

pckroon: Likewise, so:
```suggestion
    # FIXME we need to assing updated resids to the modification atoms
    # the block mapping does this for the unmodified atoms
    for mod_idx, out_idx in mod_to_out.items():
        if mod_idx in anchors:
            continue
        neighbors = modification.neighbors(mod_idx)
        for mod_idx_neigh in neighbors:
            anchor = mod_to_out[mod_idx_neigh]
            resid_new = graph_out.nodes[anchor].get('resid', None)
            resid_old = graph_out.nodes[anchor].get('_old_resid', None)
            if resid_new:
                graph_out.nodes[out_idx]['_old_resid'] = resid_old
                graph_out.nodes[out_idx]['resid'] = resid_new
                break
        if not resid_new:  # Last found resid is *still* None....
            raise ValueError(f"No resid found for {format_atom_string(graph_out.nodes[out_idx])}")
```

Issue #664, opened 2025-02-24
docs napoleon aliases


csbrasnett:
In #661 we changed the napoleon aliases to None to fix an issue with the docs. In future this will be fixed in Sphinx, so we can put it back.

see:

https://github.com/sphinx-doc/sphinx/issues/13366
https://github.com/marrink-lab/vermouth-martinize/pull/661

PR #665, opened 2025-02-24
restore napolean_type_aliases with sphinx 8.2.0 -> 8.2.1

csbrasnett:


csbrasnett:
@pckroon this will already resolve #664 🤦

pckroon (changes requested):
Could you also change the required sphinx version to >=8.2.1?

PR #671, opened 2025-03-04
improved backbone atom selection

csbrasnett:
With this PR, write the `[ macros ]` directive into the force field meta. This can then propagate useful things like atomnames for selectors. In this case, we use it to select backbones per force field when applying position restraints


csbrasnett:
There's a bit of a mess with the go stuff now in that we have a flag, `-go-backbone`, which could be made redundant if the backbone name is defined elsewhere. But it might be good to retain for flexibility. Just a thought.

vermouth/selectors.py
```diff
@@ -176,5 +175,9 @@ def filter_minimal(molecule, selector):
...
     """
     for name, atom in molecule.nodes.items():
-        if selector(atom):
-            yield name
+        if kwargs:
+            if selector(atom, kwargs['bb_atomname']):
+                yield name
+        else:
+            if selector(atom):
+                yield name
```
csbrasnett, line 183: this is really messy and not ideal, can't think of a better solution atm

vermouth/rcsu/go_structure_bias.py
```diff
@@ -181,8 +186,19 @@ def contact_selector(self, molecule):
...
+                                                        select_backbone,
+                                                        bb_atomname=self.backbone))
+                        bb_node_B = next(filter_minimal(self.res_graph.nodes[resB]['graph'],
+                                                        select_backbone,
+                                                        bb_atomname=self.backbone))
+                    except StopIteration:
+                        LOGGER.warning(f'No backbone atoms with name "{self.backbone}" found in molecule. '
+                                       'Check -go-backbone argument if your forcefield does not use this name for '
+                                       'backbone bead atoms. Go model cannot be generated. Will exit now.')
+                        sys.exit()
```
pckroon, line 200:
```suggestion
                        sys.exit(1)
```

pckroon (changes requested):
I really like the functionality, but I suggest an alternative implementation. Macros are not (intended to be) ff-wide, but rather file bound.
Instead, use the forcefield variables (`general.ff`), and allow using those in the ff files. For example:
```
[ macros ]
protein_resnames $variables[protein_resname]
```
Or `variables.protein_resname`. Or a `:` separator. Or something.
Note: $variables should be resolvable anywhere in a ff file, not just in the macros section as far as I'm concerned. This does mean that the order in which the files get parsed becomes important --- the variables need to be parsed first...

csbrasnett:
I think that satisfies what you want now? Thanks for the clarification, I should have dug around the parsing more before adding another parser!

some small notes:
- I wrote tests for ApplyPosres in #659. Currently they don't exist in main, but we can add to that here as appropriate once that's merged.
- I haven't moved other stuff from the `[ macros ]` around, and tbh I'm not sure I see the motivation for it?
- If this is all good at this point I'll add the definition to other force fields too

pckroon (changes requested):
I like it this way :)
I still think being able to use ff variable in ff files would be a nice feature, but that's a different PR.
I also just merged the POSRES_FC PR. Could you resolve the merge conflict with this one?

csbrasnett:
Nothing more to add atm thanks!

PR #675, opened 2025-03-10
Protect angles in helices when adjacent to prolines (2nd try)

### LINKED ISSUE #674: Pro-adjacent residues in helices have weak, unprotected angles.
The current implementation weakens helical BB-BB-BB angles whenever a Pro is present, which makes sense.

The helical dihedrals, however, are never switched off (shouldn't they be?). Helix BB-BB-BB angles centered on Pro are, therefore, protected by type 10 angles from reaching 180 deg.

Angles involving adjacent residues, however, are problematic because they are also weakened, but not protected by type 10 potentials. Because they can more easily reach 180 deg, and due to the imposed helical dihedral, the system can become unstable.

We observed this only when in combination with GōMartini elastic potentials, although the vulnerability should exist even in plain Martini.

PR #673 was opened with changes to protect Pro-adjacent helical angles with type 10 potentials, just like the Pro-centered angles are protected.

mnmelo:
Re-submitting this PR because its original incarnation (#673) was mistakenly closed too early. See the discussion there.

The current implementation weakens helical BB-BB-BB angles whenever a Pro is present, which makes sense.

The helical dihedrals, however, are never switched off (shouldn't they be?). Helix BB-BB-BB angles centered on Pro are, therefore, protected by type 10 angles from reaching 180 deg.

Angles involving adjacent residues, however, are problematic because they are also weakened, but not protected by type 10 potentials. This PR fixes that.

Fixes #674.


mnmelo:
Should be good to go,  @pckroon.

pckroon (approved):
Cheers! Looks good to me!

PR #672, opened 2025-03-05
bump image for tests from ubuntu-20.04 to ubuntu-latest 

csbrasnett:
bump image from ubuntu-20.04 to ubuntu-latest per https://github.com/actions/runner-images/issues/11101

PR #676, opened 2025-03-11
pause tests for dssp executable

csbrasnett:
With reference to #672 update the tests to run on `ubuntu-latest`, and accordingly remove support for the dssp executable.

This requires moving MDTraj from an optional dependency to a hard dependency.

Any code referring to a dssp executable has been removed. Possibly some test files can also now be removed.


pckroon (commented):
That cleans up nicely. Does @fgrunewald have an opinion about removing dssp support in favour of mdtraj?

csbrasnett:
That's fair enough. The problem ran into in #672 is that upgrading the tests to run on `ubuntu-latest` instead of `ubuntu-20.04` is the significant change in the version of dssp. I guess if you're happy to have some tests failing/remove the tests then we can live with it, while someone finds time to check a new version of dssp?

fgrunewald:
@csbrasnett you can use mdtraj for testing instead of dssp and comment out the dssp tests. This is just about making a "hrad" dependence i.e. putting it into setup.cfg

csbrasnett:
there're tests in there that can be changed to use mdtraj/will run fine without the executable. Maybe for now we leave them commented out and sort the whole thing once a newer version of dssp is supported

fgrunewald (approved):

Issue #654, opened 2025-01-31
Martinizing large proteins/protein complexes


TommyBiomodelling:
hello there vermouth-martinize people!

I am currently having trouble martinizing a fairly large protein complex (~1MDa). The problem is simply due to the number limit of pdb files (99999 atoms), is there any way to fix this or circumvent this problem when coarse graining with martinize?

cheers,
Tommy

csbrasnett:
Hi Tommy,

What command are you using? Are you processing your input file to a clean pdb using grep beforehand? I don't know if that will definitely solve the problem, but it might help.

Thanks for the suggestion of parsing .cif files. We can definitely add it to the list of ongoing improvements

Chris

pckroon:
I've had a (very) brief look at a cif reader/writer a couple of years ago already, but never had the time to really sink my teeth into it. 
A cif reader/writer would also relax the dssp version requirements quite a bit actually.

pckroon:
Let me drop this here: https://pypi.org/project/PyCifRW/

shastry2358:
Came here looking for a solution to the same problem. Are there any suggestions from the developers on martini-zing systems (protein + lipid) that exceed the traditional PDB formats capacity for index numbers?

PR #679, opened 2025-03-17
add parsing for CIF files

csbrasnett:
Per #654 read CIF files using the PyCifRW library. 

Have tested on a number of proteins with both their pdb and cif files, and got identical results. 

Two possible issues:

~1) The file reader library can be quite slow~ solved with prefiltering in https://github.com/marrink-lab/vermouth-martinize/pull/679/commits/da9d22b6a9be5076b94c9488288d20f1c1a5cc64 
~2) When reading a solution NMR cif file, all the models were loaded. We can probably write an exception for this and insist on using only a single structure in the file. This is likely to be a minor issue because AFAIU one of the main reasons for including cif file parsing would be so users can martinize AF3 structures~ solved with addition of model reading


vermouth/pdb/cif.py
```diff
@@ -0,0 +1,133 @@
...
+    model: int
+        If the PDB file contains multiple models, which one to select.
+
+    Returns
+    -------
+    list[vermouth.molecule.Molecule]
+        The parsed molecules. Will only contain edges if the PDB file has
+        CONECT records. Either way, the molecules might be disconnected. Entries
+        separated by TER, ENDMDL, and END records will result in separate
+        molecules.
```
pckroon, line 64: ??

vermouth/pdb/cif.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+
+    # list of categories from the read cif file to read into a molecule
+    cif_categories = [
+    '_atom_site.id', # atomid
+    '_atom_site.auth_atom_id', # atomname
+    '_atom_site.auth_comp_id', # resname
+    '_atom_site.auth_asym_id', # chain
+    '_atom_site.auth_seq_id', # resid
+    #'_atom_site.pdbx_pdb_ins_code', # insertion_code
```
pckroon, line 74: Shouldn't this be read?

vermouth/pdb/cif.py
```diff
@@ -0,0 +1,133 @@
...
+    '_atom_site.cartn_x', # x
+    '_atom_site.cartn_y', # y
+    '_atom_site.cartn_z', # z
+    #'_atom_site.occupancy', # occupancy
+    #'_atom_site.b_iso_or_equiv', # temp_factor
```
pckroon, line 79: Likewise?

vermouth/pdb/cif.py
```diff
@@ -0,0 +1,133 @@
...
+
+    cf = ReadCif(str(file_name))
+
+    molecules = []
+
+    for model in cf.keys():
+
+        molecule = Molecule()
+        # annoyingly _refine_hist.number_atoms_total is not a mandatory entry
+        number_atoms = len(cf[model]['_atom_site.cartn_x'])
```
pckroon, line 98:
```suggestion
        number_atoms = len(cf[model]['_atom_site.id'])
```
Is it worth checking whether _refine_hist.number_atoms_total is the same as number_atoms? (if given)

vermouth/pdb/cif.py
```diff
@@ -0,0 +1,133 @@
...
+            for attr, attr_type, cif_key in zip(cif_category_names, cif_category_types, cif_categories):
+                properties[attr] = attr_type(cf[model][cif_key][idx])
+
+            if properties['resname'] in exclude or (ignh and properties['element'] == 'H'):
+                continue
+
+            properties['position'] = np.array([properties['x'],
+                                               properties['y'],
+                                               properties['z'],
+                                               ], dtype=np.float32) / 10
```
pckroon, line 112:
```suggestion
                                               ], dtype=np.float) / 10
```

vermouth/pdb/cif.py
```diff
@@ -0,0 +1,133 @@
...
+
+            molecule.add_node(idx, **properties)
+
+        molecule.box = _cell(cf, model)
+        molecules.append(molecule)
+
+    return molecules
+
+class CIFReader():
+    def __init__(self, file, exclude, ignh):
```
pckroon, line 122: Make sure that this signature is the same as for the pdb reader

vermouth/processors/cif_reader.py
```diff
@@ -0,0 +1,57 @@
...
+        from the final molecule(s)
+    ignh: bool
+        If True, hydrogens will be discarded from the input structure.
+
+    See also
+    --------
+    :func:`~vermouth.pdb.cif.read_cif_file`
+
+    """
+    def __init__(self, filename, exclude=(), ignh=False):
```
pckroon, line 48: Also check signature here

pckroon (changes requested):
Some small points. I'll have a look at the tests later

vermouth/pdb/cif.py
csbrasnett: I left them out because they don't actually get read in the pdb reader either, as far as I can see?

pckroon: Fair enough. Doesn't cost anything to do read them though.

csbrasnett:
tested of 610 pairs of pdbs/cifs, cif only fractionally slower than pdb, and no errors found in comparing sets of output itps

image

vermouth/tests/pdb/test_cif.py
```diff
@@ -0,0 +1,127 @@
...
+        anum = tokens[1]
+        aname = tokens[-2]
+        resname = tokens[-4]
+        chain = tokens[-3]
+        resid = tokens[-5]
+        insert_code = tokens[9]
+        x = tokens[10]
+        y = tokens[11]
+        z = tokens[12]
+        position = np.array([tokens[10], tokens[11], tokens[12]], dtype=np.float32) / 10
```
pckroon, line 78:
```suggestion
        position = np.array([tokens[10], tokens[11], tokens[12]], dtype=float) / 10
```

vermouth/tests/pdb/test_cif.py
```diff
@@ -0,0 +1,127 @@
...
+    return data
+
+
+def test_read_cif_file():
+
+    molecule = cif.read_cif_file(CIF_PROTEIN)[0]
+
+    reference_data = cif_lines()
+
+    for i in list(molecule.nodes)[None:17:None]:
```
pckroon, line 109:
```suggestion
    for i in sorted(molecule.nodes)[:len(reference_data)]:
```

vermouth/tests/pdb/test_cif.py
```diff
@@ -0,0 +1,127 @@
...
+        assert all([all([j in molecule.nodes[i].keys() for j in reference_data[i].keys()])])
```
pckroon, line 110:
```suggestion
        assert all(j in molecule.nodes[i].keys() for j in reference_data[i].keys())
```

vermouth/tests/pdb/test_cif.py
```diff
@@ -0,0 +1,127 @@
...
+    processor = cif.CIFReader(file=CIF_PROTEIN, exclude=('HOH'), ignh=False)
+
+    molecules = processor.reader()
+
+    assert len(molecules) == 1
+
+
+
+
+
```
pckroon, line 127:
```suggestion

```

pckroon (changes requested):
Is it an option to also test if 1ubq.pdb and 1ubq.cif result in the same molecules?

vermouth/pdb/cif.py
```diff
@@ -0,0 +1,166 @@
...
+    molecules = []
+
+    molecule = Molecule()
+
+    # find the indices in the data which are actually from the model that we're after.
+    model_atom_indices = np.where(np.array(cf[fname]['_atom_site.pdbx_PDB_model_num'], dtype=int) == modelidx)[0]
+
+    idx = 0  # add nodes by separate index in case we skip some atoms
+    for i in model_atom_indices:
+        properties = properties_dict_list[i]
```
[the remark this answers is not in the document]
csbrasnett: I would let it break earlier during the filtering loop, checking that the key is present in the data that's been read in. So we have something like:

```python
if category in cf[fname]:
    # cast the data to its correct type
     data = [casting(i, cattype) for i in cf[fname][category]]
     all_data.append(data)
     names.append(name)
elif category in [ essential_properties ]:
     LOGGER.warning(f"you must have {name} in your input file")
```

where `essential_properties` is defined from a subset of `cif_categories_all`. MY suggestion is it contains [ resid, resname, atomname, x, y, z ] ?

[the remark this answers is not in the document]
pckroon: A warning sounds good.
[resname, resid, (atomname or element)] I'd say. And if there's no atomname, but there is an element you/we can derive a unique atomname based on the element. Anything else should get filled in later automatically I think

vermouth/pdb/cif.py
```diff
@@ -128,13 +142,14 @@ def read_cif_file(file_name, exclude=('SOL', 'HOH'), ignh=False, modelidx=1):
...
     # find the indices in the data which are actually from the model that we're after.
-    model_atom_indices = np.where(np.array(cf[fname]['_atom_site.pdbx_PDB_model_num'], dtype=int) == modelidx)[0]
+    model_atom_indices = [i for i, x in enumerate([int(j) == modelidx for
+                                                   j in cf[fname]['_atom_site.pdbx_PDB_model_num']]) if x]
 
     idx = 0  # add nodes by separate index in case we skip some atoms
-    for i in model_atom_indices:
-        properties = properties_dict_list[i]
+    for _ in model_atom_indices:
+        properties = properties_dict_list[_]
```
pckroon, line 150:
```suggestion
    for idx in model_atom_indices:
        properties = properties_dict_list[idx]
```

vermouth/pdb/cif.py
```diff
@@ -128,13 +142,14 @@ def read_cif_file(file_name, exclude=('SOL', 'HOH'), ignh=False, modelidx=1):
     molecule = Molecule()
 
     # find the indices in the data which are actually from the model that we're after.
-    model_atom_indices = np.where(np.array(cf[fname]['_atom_site.pdbx_PDB_model_num'], dtype=int) == modelidx)[0]
+    model_atom_indices = [i for i, x in enumerate([int(j) == modelidx for
+                                                   j in cf[fname]['_atom_site.pdbx_PDB_model_num']]) if x]
```
pckroon, line 146: Are there better variable names than i, j, and x?

vermouth/pdb/cif.py
```diff
@@ -104,8 +104,13 @@ def read_cif_file(file_name, exclude=('SOL', 'HOH'), ignh=False, modelidx=1):
                           float, float, float,
                           float, float,
                           str, float, int]
+    essential_properties = ['resname', 'resid']
 
     cf = ReadCif(str(file_name))
+    # PyCifRW seems to store everything from the file it reads in a top level key from _entry.id
+    # which ~ corresponds to the pdb code. From a single file we should only have one key.
+    if len(cf.keys()) > 1:
+        LOGGER.warning("This cif file contains multiple entries. Will parse the first one.")
```
pckroon, line 113: Is there a way to give this warning only if the user doesn't give a modelidx?

pckroon (changes requested):
Very nice work so far. A few more small things

vermouth/pdb/cif.py
csbrasnett: This isn't to do with the model index contained in the entry, it's if there are multiple entries in the same cif file. If you e.g. concatenated 1UBQ and 1DPX into the same file, then (afaiu) that's a valid .cif file with 2 `_entry.id` lines. Then we get:

```
cf = ReadCif(str(file_name))
>>> cf.keys()
['1UBQ', '1DPX']
```

each of which could contain multiple models in their `_atom_site` loops.

pckroon: Aaaah... Don't we then just create multiple molecules?

csbrasnett: possibly... I'll investigate whether having multiple entries per file is particularly common. I'd guess that you'd get into trouble later with multiple _cell entries for a single system, so it might be easiest to leave as is

pckroon: I think I understand why that would be an issue, but why would that be an issue? We don't use box information anyway do we?

csbrasnett:
Had to remove the multiple entries test, I don't understand why `assert any([rec.levelname == 'INFO' for rec in caplog.records])` doesn't find anything when I can see that the log is written?

pckroon (changes requested):
Great work. 2 final nitpicks, then I think I'll be happy with it
The caplog test for the info message probably does not work because of loglevels, i.e. only warning and higher gets logged by default

pckroon (approved):
Awesome. Looks good to me!

Issue #689, opened 2025-04-17
dssp error


csbrasnett:
currently from a source installation there's something wrong with the dssp editing that went on in #676 - just using `-dssp` seems to raise the missing installation error even when mdtraj is definitely installed.

PR #691, opened 2025-04-21
Secstruc gmx header fix

csbrasnett:
Per #689 I thought I had something weird going on with an mdtraj installation not being picked up and used, but it turned out that was a local environment problem. While tracking it down to make sure it was all ok, I did notice that with the recent changes 

1) the dssp string wasn't actually being written to the header because we were still looking for a `secstruct` attribute not `aasecstruct`
2) the business with changing the effective string wrt IDRs wasn't actually being used because of where the dssp gmx header annotation comes in the pipeline.

So, this PR fixes 1) easily and 2) by moving some code around

### DIFF
--- a/vermouth/dssp/dssp.py
+++ b/vermouth/dssp/dssp.py
@@ -548,7 +548,7 @@ def gmx_system_header(system):
     ss_sequence = list(
         itertools.chain(
             *(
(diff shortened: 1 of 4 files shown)


pckroon (approved):
Excellent find!

PR #697, opened 2025-05-13
Idr annotation fix

csbrasnett:
Currently the `-id-regions` flag only accepts resid regions, with no chain specifier. This PR fixes that by slightly changing the input to be more like the modifications style, `<chain>-<resid_start>:<resid_end>`.

### DIFF
--- a/vermouth/processors/annotate_idrs.py
+++ b/vermouth/processors/annotate_idrs.py
@@ -52,6 +52,8 @@ def parse_residues(resspec):
         out['resids'] = [(int(res_start), int(res_end))]
     if chain:
         out['chain'] = chain[0]
+    else:
+        out['chain'] = None
     return out
 
 def annotate_disorder(molecule, id_regions, annotation="cgidr"):
--- a/vermouth/processors/annotate_idrs.py
+++ b/vermouth/processors/annotate_idrs.py
@@ -69,8 +69,8 @@ def annotate_disorder(molecule, id_regions, annotation="cgidr"):
         for key, node in molecule.nodes.items():
             _old_resid = node['_old_resid']
             chain = node['chain']
-            # make sure we have the correct chain and are in the right region. If no chain in resspec assume True.
-            if (chain == region.get('chain', True)) and _in_resid_region(_old_resid, region.get('resids')):
+            # make sure we have the correct chain and are in the right region. If no chain in region assume single chain system.
+            if (region.get('chain') is None or region.get('chain') == chain) and _in_resid_region(_old_resid, region.get('resids')):
                 molecule.nodes[key][annotation] = True
                 if "cgsecstruct" in molecule.nodes[key] and molecule.nodes[key]["cgsecstruct"] != "C":
                         molecule.nodes[key]["cgsecstruct"] = "C"
--- a/vermouth/processors/annotate_idrs.py
+++ b/vermouth/processors/annotate_idrs.py
@@ -23,8 +23,36 @@ from ..selectors import is_protein
 from .processor import Processor
 from ..rcsu.go_utils import _in_resid_region
 from ..log_helpers import StyleAdapter, get_logger
+
 LOGGER = StyleAdapter(get_logger(__name__))
 
+
+def parse_residues(resspec):
+    """
+    Parse a residue specification: [<chain>-][[#]<resid>]:[[#]<resid>] where
+    resid is /[0-9]+/.
+    Returns a dictionary with keys 'chain', 'resid_start', and 'resid_end' for the
+    fields that are specified. Resids will be ints.
+
+    Parameters
+    ----------
+    resspec: str
+
+    Returns
+    -------
+    dict
+
+    """
+    # <chain>-<resid>
+    *chain, resids = resspec.split('-', 1)
+    res_start, res_end = resids.split(':', 1)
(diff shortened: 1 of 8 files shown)


vermouth/processors/annotate_idrs.py
```diff
@@ -23,9 +23,37 @@
...
+    res_start, res_end = resids.split(':', 1)
+
+    out = {}
+    if resids:
+        out['resids'] = [(int(res_start), int(res_end))]
+    if chain:
+        out['chain'] = chain[0]
+    return out
+
 def annotate_disorder(molecule, id_regions, annotation="cgidr"):
```
pckroon, line 57: The meaning of id_regions changed, update the docstring

vermouth/processors/water_bias.py
```diff
@@ -87,16 +90,20 @@ def assign_residue_water_bias(self, molecule, res_graph):
...
+                    if (region.get('chain') is None or region.get('chain') == chain) and _in_resid_region(_old_resid,
+                                                                                                          region.get(
+                                                                                                                  'resids')):
+                        eps = self.water_bias.get('idr', 0.0)
+                        sec_struc = res_graph.nodes[res_node]['cgsecstruct']
             elif self.auto_bias:
                 sec_struc = res_graph.nodes[res_node]['cgsecstruct']
                 eps = self.water_bias.get(sec_struc, 0.0)
             else:
                 continue
```
pckroon, line 106:
```suggestion
            if self.auto_bias:
                sec_struc = res_graph.nodes[res_node]['cgsecstruct']
                eps = self.water_bias.get(sec_struc, 0.0)
            else:
                for region in self.idr_regions:
                    if (region.get('chain') is None or region.get('chain') == chain) and _in_resid_region(_old_resid,  region.get('resids')):
                        eps = self.water_bias.get('idr', 0.0)
                        sec_struc = res_graph.nodes[res_node]['cgsecstruct']
```
I think this should do the exact same

vermouth/processors/water_bias.py
```diff
@@ -135,17 +142,19 @@ def remove_cross_nb_interactions(self, molecule, res_graph):
...
-                                                              prefix=molecule.meta.get('moltype')))
-                all_cross_pairs.append(np.where(all_go_pairs == vs_go_node)[0]) #just need the first one
+        for region in self.idr_regions:
+            for res_node in res_graph.nodes:
+                resid = res_graph.nodes[res_node]['resid']
+                _old_resid = res_graph.nodes[res_node]['_old_resid']
+                chain = res_graph.nodes[res_node]['chain']
+                if (region.get('chain') is None or region.get('chain') == chain) and _in_resid_region(_old_resid,
+                                                                                                      region.get(
+                                                                                                              'resids')):
```
pckroon, line 152:
```suggestion
                if (region.get('chain') is None or region.get('chain') == chain) and _in_resid_region(_old_resid, region.get('resids')):
```

pckroon: Maybe also time to abstract this to a function

vermouth/rcsu/go_utils.py
```diff
@@ -79,3 +79,29 @@ def _get_bead_size(atype):
...
+    -------
+    bool
+    """
+    condition0 = (region.get('chain') is None or region.get('chain') == chain)
+    condition1 = _in_resid_region(resid, region.get('resids'))
+
+    if condition0 and condition1:
+        return True
+    else:
+        return False
```
pckroon, line 106:
```suggestion
    return condition0 and condition1
```

vermouth/processors/water_bias.py
```diff
@@ -92,18 +92,13 @@ def assign_residue_water_bias(self, molecule, res_graph):
...
-            elif self.auto_bias:
+            if self.auto_bias:
                 sec_struc = res_graph.nodes[res_node]['cgsecstruct']
                 eps = self.water_bias.get(sec_struc, 0.0)
-            else:
-                continue
+            for region in self.idr_regions:
+                if _in_chain_and_resid_region(region, _old_resid, chain):
+                    eps = self.water_bias.get('idr', 0.0)
+                    sec_struc = res_graph.nodes[res_node]['cgsecstruct']
```
pckroon: Here the logic changed now? Please double check this is working as intended

pckroon (changes requested):

vermouth/processors/water_bias.py
csbrasnett: yes, the logic has changed, but if you look at the expected behaviour in `test_water_bias.py::test_assign_residue_water_bias[secstruc4-water_bias4-idr_regions4-expected4]` (which otherwise fails) then this is actually the expected behaviour: a residue with an IDR annotation has the IDR take over from the auto biasing of the secondary structure. The original snippet is:

```
       if _in_resid_region(_old_resid, self.idr_regions):
           eps = self.water_bias.get('idr', 0.0)
           sec_struc = res_graph.nodes[res_node]['cgsecstruct']
       elif self.auto_bias:
           sec_struc = res_graph.nodes[res_node]['cgsecstruct']
           eps = self.water_bias.get(sec_struc, 0.0)
       else:
           continue
```

which works out this same way

pckroon (approved):

PR #704, opened 2025-06-17
Add molecule sorting function to pdb_to_universal

csbrasnett:
From a user bug error, some missing links were found by gromacs (atoms unconnected to anything else). Investigation found that this was the result of a scrambling of node order somehow applied by the CanonicalizeModifications processor. I didn't investigate any further cause. Adding the extra processor in the pipeline appears to resolve the issue.

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -164,6 +164,7 @@ def pdb_to_universal(
             defer_writing=False,
         )
     vermouth.AttachMass(attribute="mass").run_system(canonicalized)
+    vermouth.SortMoleculeAtoms().run_system(canonicalized)
     return canonicalized
 
 


pckroon:
I need more background on this. Links do nothing with node and atom indices, so I don't see how sorting would affect those.

csbrasnett:
We have an input pdb with 2 chains. Running:
```
$ martinize2 -f input.pdb -o topol.top -x 6ZH9_CG.pdb -dssp -go contacts.map -ignh -name molecule_0 -merge E,F -noscfix -resid input 
```

reveals that for the resids, chain A runs 1-195 as expected. Chain B has 128 residues, but somewhere (apparently in CanonicalizeModifications) the resid ordering gets shuffled to run 68-128 then 1-67. This somehow later causes a bond to be missing. IE, there should be a bond 67BB-68BB, but this is missing, which causes a fatal error in gromacs.

I really can't tell why that processor is shuffling the order of the nodes, or why it should make the bond missing later, but this definitely fixes it

pckroon (approved):

fgrunewald:
If the chain IDs are missing martinize assigns them no ID, which is the same as all residues in the same chain. Since we require unique resids they get labeled and later sorted but that might lead to unpredictable resids in the output. It might have to do something with the ` -merge E,F `. Do you have the pbd ?

PR #711, opened 2025-07-10
remove -idr-tune flag.

csbrasnett:
If `-id-regions` is given, use id features automatically. This cleans things up a bit and makes sure that a small missing flag doesn't generate wrong parameters.


pckroon (dismissed):
Excellent. Do you feel like fixing the beauty mark? It's optional as far as I'm concerned.

pckroon (approved):

Issue #720, opened 2025-08-14
sanitise ff files


csbrasnett:
some things from workshop this week:

1) Remove martini3 beta force fields 
2) Merge martini3-IDP into martini3
     - and document this somewhere
3) Maybe split force fields into `aminoacids.ff` and `links.ff` files for sanitisation? 

Not so bothered about the one, but might be healthy for ongoing optimisation/maintenance. It also might help make clearer in a merged martini force field how IDP interactions are split out from folded stuff, and it's intended that they can be mixed.

csbrasnett:
in doing this, should integration tests that use force fields to be removed be removed themselves, or just updated? @pckroon

PR #739, opened 2025-10-24
ff file sanitising

csbrasnett:
addresses #720 with the hope that combining various martini3 force fields into a single one will help user confusion.

This PR firstly merges the Martini3-IDP force field into `martini3001`, and deletes the `martini30b32` and `martini30dev` force fields.

Secondly, we split out the links and idr-specific links from `aminoacids.ff` in `martini3001` so that these can be managed more easily in future.

2 integration tests needed updating. tier-1: lysozyme used the removed `martini30b32` force field, and tier-1: hst5 used the old idp parameters from GoMartini.

the final thing to check here is the readthedocs in case there is now outdated information about idr parameters, and a blog post for the website


pckroon:
IIRC the last link to apply has priority, so simply make sure the IDR link is below the regular one.

pckroon (changes requested):
Very nice. Some minor comments.
I assume you checked the resulting changes in the test outputs?

PR #744, opened 2025-11-20
Add 4 virtual sites to Gō models for intra- and inter-molecular bond distinction

niek-fraaije:
By using 4 virtual sites in the go models, intra molecular and inter molecular go bonds can be distinguished. The main advantage is that 'general' itp files can be generated for homologous molecules in you system.

In total, seven files were modified. The first was the main activation script (martinize2), where arguments are parsed and pipelines are initialized. Five changes were made within the G¯o-bonds pipeline itself, and one in the processor responsible for constructing molecular objects, specifically in the bond-creation module (make_bonds.py).

In martinize2, the argument parsing was extended to allow more flexibility in defining G¯o-bond parameters. Instead of specifying a single ε value, users can now provide two values: one for intramolecular G¯o bonds (-go-eps-intra) and one for intermolecular G¯o-bonds (-go-eps-inter). Similarly, cutoff distance ranges can be set independently for both types of interactions (-go-low-intra,-go-up-intra,-go-low-inter,-go-up-inter). An additional required argument (-go-ff) specifies the Martini force field file (e.g. martiniv3.0.0.itp), which is needed to retrieve the standard bead–bead nonbonded interactions. These interactions are canceled through the introduction of virtual sites when G¯o-bonds are applied. 

Several modifications ensured that molecules were handled individually rather than as a merged system. In make_bonds.py, the assignment of the mol_idx attribute was changed so that unique indices are assigned after the system is separated into connected molecules by the MakeBonds processor. This step is essential because the generation of G¯o-bonds now depends on distinguishing molecules individually, rather than assuming a single global topology.

Changes were also required in contact_map.py. Previously, the function MergeAllMolecules() combined all molecules into one object before building the contact map, which is incompatible with the new approach. The code was modified to operate on a list of molecule graphs, one for each molecule, with mol_idx used to maintain the mapping between graphs and their atoms. This ensures that the contact map is generated on a per-molecule basis, while still retaining compatibility with the downstream pipeline. Now instead of reading every atom of the system, every atom of each of the molecule graphs are read in the function, so that the intermolecular contacts are still detected.

With these changes in place, the system can be martinized as usual. The GoPipeline then generates the appropriate G¯o-bonds for each molecule based on the updated contact maps. As a first step, the pipeline groups structurally identical molecules and assigns each group a common reference molecule. To facilitate efficient lookups, two mapping dictionaries are constructed: unique_mols_map, which stores, for each molecule, the set of equivalent molecules of the same type, and reference_mol_map, which records the designated reference molecule for each molecule in the system.

The second step introduces the four virtual sites, labeled a, b, c, and d, to every backbone bead of the protein.
• a--a: negative interactions reproducing the baseline Martini 3 nonbonded terms.
• b--b: positive interactions defining the intermolecular G¯o-bonds.
• c--c: positive interactions defining the intramolecular G¯o-bonds (which, without correction, would also
contribute intermolecularly).
• d--d: negative interactions that cancel the unwanted intermolecular component of the c--c interactions, ensuring they act strictly intramolecularly.

Changes in go_vs_includes.py include the introduction of four virtual sites per backbone bead, instead of a single virtual site. The code now also checks whether a virtual site has already been added, preventing duplicates that can arise when handling homologous molecules. 

The final step in the G¯o-bond pipeline is the addition of the appropriate contacts, defined in the contact map, to the coarse-grained system. This is implemented in go_structure_bias.py. Extensive modifications were made to allow for per-molecule identification and addition of G¯o-bonds, as well as to integrate the four virtual sites into this process. Instead of treating the system as a single merged molecule, each molecule is separately converted into a graph and stored in a dictionary (molecule_graphs) keyed by mol_idx. Contacts satisfying the G¯o-bond criteria based on user-specified distance ranges are selected using contact_selector().

Within contact_selector(), each molecule is evaluated independently, and all intermolecular contacts are obtained via the initial mapping of homologous molecules. Necessary exclusions for virtual sites b and d are identified and added; these exclusions are required only for intra- and intermolecular G¯o-bonds between homologues. All backbone bead combinations are recorded for easy lookup of Lennard–Jones parameters from the Martini forcefield. All valid contacts are compiled into symmetrical matrix, which records the bond type (intra-, intermolecular with a homologue, or intermolecular with a non-homologue), the virtual sites involved, and the backbone bead types. Identical contacts among homologous molecules are checked to not differ significantly. Minimum and maximum measured distance should not differ more than 20%, else there will be a warning indicating where the contact is that is too variable.

Using symmetrical_matrix, the G¯o-bonds are generated via compute_go_interaction(). A dictionary mapping bead–bead pairs to their normal Martini3 Lennard–Jones parameters is first constructed from the provided forcefield. For each contact, the σ and ε values of the virtual sites are then set according to this dictionary and the user-specified-go-eps-inter and-go-eps-intra arguments.

Finally, since the system is no longer treated as merged, martinize2 writes separate topology files for each unique molecule. All virtual sites are defined in go_nbparams.itp and go_interaction parameters and atom types, respectively.

Note:
If assigning mol_idx at the end of make_bonds becomes problematic, this identifier can be replaced by another attribute.

A tutorial on using this go bond implementation is found at another github page: 
https://github.com/niek-fraaije/Tutorial-go-bonds

A tutorial to visualize the these go bonds dynamically with vmd is found at:
https://github.com/niek-fraaije/Go-bond-visualization

### DIFF
--- a/bin/martinize2
+++ b/bin/martinize2
@@ -17,7 +17,6 @@
 """
 High level API for Martinize2
 """
-import copy
 import argparse
 import functools
 import logging
@@ -562,13 +561,6 @@ def entry():
         help=("Path to martini_v3.0.0.itp"),
     )
     go_group.add_argument(
-        "-go-eps",
-        dest="go_eps",
-        type=float,
-        default=9.414,
-        help=("The strength of the Go model structural bias in kJ/mol."),
-    )
-    go_group.add_argument(
         "-go-eps-inter",
         dest="go_eps_inter",
         type=float,
@@ -1024,7 +1016,7 @@ def entry():
     if args.go:
         if isinstance(args.go, Path):
             LOGGER.info("Reading Go model contact map.", type="step")
-            read_go_map(system=system, file_path=args.go) ##### this needs to be changed so that it also gives the mol_idx as output into the 'go_params'
+            read_go_map(system=system, file_path=args.go)
 
         else:
             LOGGER.info("Generating Go model contact map.", type="step")
--- a/vermouth/graph_utils.py
+++ b/vermouth/graph_utils.py
@@ -227,7 +227,7 @@ def partition_graph(graph, partitions):
                 new_graph.add_edge(new_idx, new_jdx, **edge_attrs)
     return new_graph
 
-def make_residue_graph(graph, attrs=('chain', 'resid', 'resname', 'insertion_code'), go=False):
+def make_residue_graph(graph, attrs=('chain', 'resid', 'resname', 'insertion_code')):
     """
     Create a new graph based on `graph`, where nodes with identical attribute
     values for the attribute names in `attrs` will be contracted into a single,
@@ -250,8 +250,7 @@ def make_residue_graph(graph, attrs=('chain', 'resid', 'resname', 'insertion_cod
         The resulting coarser graph, where equivalent nodes are contracted to a
         single node.
     """
-    if go:
-        attrs=('atomid', 'chain', 'resid', 'resname', 'insertion_code', 'resid_before_merge')
+
     # Create partitions. These will contain all nodes, even those without e.g.
     # a resname, since those will get resname None
     residue_idxs = collect_residues(graph, attrs)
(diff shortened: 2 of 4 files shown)


niek-fraaije:
Hi, I’ve rebased the changes onto the go-implementation branch of my fork and just realized that the pull request was originally created from my master branch. Should I close this PR and open a new one from the rebased branch?

Hope you enjoy the holiday season!

pckroon:
Yes please :)

PR #762, opened 2026-03-30
Add group and others read permissions to written files

csbrasnett:
files are currently written with only user rw permissions. This addition adds read rights to group and others, which makes sharing and viewing files written easier

### DIFF
--- a/vermouth/file_writer.py
+++ b/vermouth/file_writer.py
@@ -25,6 +25,7 @@ import pathlib
 import shutil
 import threading
 import tempfile
+import stat
 
 from .log_helpers import StyleAdapter, get_logger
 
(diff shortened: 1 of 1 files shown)


pckroon:
I think this is a terrible idea. I believe this is (on Linux) governed by `umask`, set by the user's shell. I see no reason why martinize2 needs to meddle with this.
I'm also not sure how this affects Windows portability.

csbrasnett:
fair enough, just been caught out by this too many times & thought it may be vaguely useful!

pckroon:
Understandable, but I don't think martinize2 is the right place to fix this (for all users, in all environments) when IMO the objectively better solution is the user setting their umask themselves.
In your case, maybe in your .bashrc?

Issue #721, opened 2025-08-15
remove folded-disordered elastic bonds


csbrasnett:
elastic network bonds applied between folded and disordered regions are not automatically removed. We do this for the Go network already so should reproduce

PR #764, opened 2026-04-17
IDR cross interaction removal

csbrasnett:
In order to ensure that we removed Go potentials that crossed folded/IDR domain boundaries, the ComputeWaterBias processor was called twice in the executable, because that function was hidden there. This PR moves that functionality to its own processor, and extends the functionality to elastic networks, addressing #721.

### DIFF
--- a/vermouth/processors/idr_cross_domain_interaction_removal.py
+++ b/vermouth/processors/idr_cross_domain_interaction_removal.py
@@ -68,7 +68,7 @@ class IDRCrossDomainInteractionRemoval(Processor):
                 chain = res_graph.nodes[res_node]['chain']
                 # if the region is in the original resid and chain
                 if _in_chain_and_resid_region(region, _old_resid, chain):
-                    # get the 
+                    # get the matching virtual site from the residue attrs
                     vs_go_node = next(get_go_type_from_attributes(res_graph.nodes[res_node]['graph'],
                                                                   resid=resid,
                                                                   chain=chain,
@@ -93,7 +93,6 @@ class IDRCrossDomainInteractionRemoval(Processor):
         """
         # find elastic bonds by meta
         elastic_bonds = [list(i.atoms) for i in molecule.interactions['bonds'] if i.meta.get('group') == "Rubber band"]
-        print(elastic_bonds)
         # list to record which items we don't want. cross = elastic bond between folded and disordered domain.
         all_cross_pairs = []
         # make a map between bond indicies and their residues. index starts from 1 so -1 for idx        
--- a/vermouth/processors/idr_cross_domain_interaction_removal.py
+++ b/vermouth/processors/idr_cross_domain_interaction_removal.py
@@ -18,6 +18,9 @@ from .annotate_idrs import parse_residues
 from ..graph_utils import make_residue_graph
 from ..rcsu.go_utils import get_go_type_from_attributes, _in_chain_and_resid_region
 import numpy as np
+from ..log_helpers import StyleAdapter, get_logger
+
+LOGGER = StyleAdapter(get_logger(__name__))
 
 class IDRCrossDomainInteractionRemoval(Processor):
     def __init__(self, 
@@ -42,9 +45,8 @@ class IDRCrossDomainInteractionRemoval(Processor):
         self.id_regions = []
         for region in id_regions:
             self.id_regions.append(parse_residues(region))
-        self.system = None
 
-    def remove_cross_nb_interactions(self, molecule, res_graph):
+    def remove_cross_nb_interactions(self, molecule, res_graph, nbparams):
         """
         Remove Go interactions between folded and disordered regions of a molecule
 
@@ -56,7 +58,7 @@ class IDRCrossDomainInteractionRemoval(Processor):
             the residue graph of the molecule
         """
         #list of all the Go pairs in the molecule
-        all_go_pairs = np.array([list(i.atoms) for i in self.system.gmx_topology_params["nonbond_params"] if 'W' not in list(i.atoms)])
+        all_go_pairs = np.array([list(i.atoms) for i in nbparams if 'W' not in list(i.atoms)])
         # list to record which items we don't want. cross = go potential between folded and disordered domain.
         all_cross_pairs = []
         # for each IDR that we have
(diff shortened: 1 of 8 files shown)


vermouth/processors/idr_interaction_optimising.py
```diff
@@ -0,0 +1,171 @@
...
+# limitations under the License.
+
+from .processor import Processor
+from .annotate_idrs import parse_residues
+from ..graph_utils import make_residue_graph
+from ..rcsu.go_utils import get_go_type_from_attributes, _in_chain_and_resid_region
+import numpy as np
+
+class IDRInteractionOptimising(Processor):
+
```
pckroon, line 24:
```suggestion
```

vermouth/processors/idr_cross_domain_interaction_removal.py
```diff
@@ -0,0 +1,171 @@
...
+        ----------
+        go: bool
+            this system contains a go model
+        elastic: bool
+            this system contains an elastic network
+        id_regions: list
+            list of tuples of residue regions defining the IDRs
+        system: vermouth.system.System
+            the system of the molecules is used for
+            storing the nonbonded parameters
```
pckroon: Doesn't match signature.

vermouth/processors/idr_interaction_optimising.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        self.go = go
+        self.elastic = elastic
+        self.id_regions = []
+        for region in id_regions:
+            self.id_regions.append(parse_residues(region))
+        self.system = None
+        self.elastic_res_distance = elastic_res_distance
+
+
```
pckroon:
```suggestion

```

vermouth/processors/idr_interaction_optimising.py
```diff
@@ -0,0 +1,171 @@
...
+        Parameters
+        ----------
+        molecule: :class:`vermouth.molecule.Molecule`
+            the molecule
+        res_graph: :class:`networkx.Graph`
+            the residue graph of the molecule
+        """
+        #list of all the Go pairs in the molecule
+        all_go_pairs = np.array([list(i.atoms) for i in self.system.gmx_topology_params["nonbond_params"] if 'W' not in list(i.atoms)])
+        # print(all_go_pairs)
```
pckroon:
```suggestion
```

vermouth/processors/idr_interaction_optimising.py
```diff
@@ -0,0 +1,171 @@
...
+        Parameters
+        ----------
+        molecule: :class:`vermouth.molecule.Molecule`
+            the molecule
+        res_graph: :class:`networkx.Graph`
+            the residue graph of the molecule
+        """
+
+        # list of all atom pairs with bonds between them in the molecule
+        all_bonds = np.array([i.atoms for i in molecule.get_interaction('bonds')])
```
pckroon: Don't elastic network bonds have a meta attribute we can use for selection?

vermouth/processors/idr_cross_domain_interaction_removal.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # delete the folded-disordered elastic network bonds from the list
+        # go backwards otherwise list order gets messed up.
+        for i in reversed(all_cross_pairs):
+            molecule.remove_interaction('bonds', tuple(i))
+
+
+    def run_molecule(self, molecule):
+        if not self.system:
+            raise IOError('This processor requires a system.')
```
pckroon: For the Go interactions I guess. Why set self.system rather than passing the system as argument to the cross-go-interaction method?

pckroon (changes requested):
I like it! Just a few small comments.
I'm not the greatest fan of the processor getting a system attribute though. Maybe there's a way around that? Also, the name of the processor could be more descriptive. Optimising is very vague. Maybe something more explicit about removing cross domain interactions?
Lastly, you do a lot of numpy to find the relevant interactions; my personal go-to would have been set logic. Did you have a specific reason to go the numpy way?

csbrasnett:
That's most of the comments. I'll chew the fat on the system issue a bit more. Looking at https://github.com/marrink-lab/vermouth-martinize/blob/master/vermouth/processors/water_bias.py where this is inherited from, I think the use of the system attribute in the processor is defensive against systems without go parameters or the need for biases? I don't know if @fgrunewald has any thoughts?

pckroon:
I guess you need to modify the system's nonbonded parameters, hence you need a system. My suggestion would probably be to override the `run_system` method, and have that do a `for molecule in system.molecules: ...`. 

Or, better idea (but harder?) is to have run_molecule somehow do some accounting on what non bonded parameters need changing, and add an appropriate function call at the end of run_system to deal with those.

vermouth/processors/idr_cross_domain_interaction_removal.py
```diff
@@ -95,45 +95,34 @@ def remove_cross_elastic(self, molecule, elastic_bond_residue_distance):
...
+                         "chain": molecule.nodes[j].get('chain')                                
+                         }                    
+                         for i,j in enumerate(molecule.nodes, 1)}
+                
         # list of bonds, but which residues they link not nodes
-        residue_bonds = np.array([[nodes_map[i].get('resid') for i in j] for j in all_bonds])
+        residue_bonds = [[nodes_map[i].get('resid') for i in j] for j in elastic_bonds]
         # same but for chains
-        chains = np.array([[nodes_map[i].get('chain') for i in j] for j in all_bonds])
+        chains = [[nodes_map[i].get('chain') for i in j] for j in elastic_bonds]
```
pckroon, line 108: lists or (frozen)sets?

csbrasnett: needs to be lists

vermouth/processors/idr_cross_domain_interaction_removal.py
```diff
@@ -97,7 +93,7 @@ def remove_cross_elastic(self, molecule):
         """
         # find elastic bonds by meta
         elastic_bonds = [list(i.atoms) for i in molecule.interactions['bonds'] if i.meta.get('group') == "Rubber band"]
-
+        print(elastic_bonds)
```
pckroon: Leftover

vermouth/processors/idr_cross_domain_interaction_removal.py
```diff
@@ -131,26 +127,33 @@ def remove_cross_elastic(self, molecule):
...
+                else:
+                    res_graph = make_residue_graph(molecule)
+
+                interactions_for_removal = self.remove_cross_nb_interactions(molecule=molecule, 
+                                                                             res_graph=res_graph,
+                                                                             nbparams=system.gmx_topology_params["nonbond_params"])
+                self._remove_nb_interactions(system, interactions_for_removal)
+
+            elif self.elastic:
+                self.remove_cross_elastic(molecule=molecule)
```
pckroon, line 158: Nice, good solution <3

pckroon (changes requested):
Excellent. Just an errant leftover print statement.

pckroon (approved):

pckroon:
No idea what's wrong with codecov today.

PR #772, opened 2026-05-22
Migrate to pyproject.toml

pckroon:
Time doesn't pause. Migrate the build and install config to the more modern pyproject.toml. Also exclude tests (and some other development files) from the distributed code.
Part of this is also moving bin/martinize to inside the vermouth package.

- [x] Fix fetch depth in CI actions
- [x] Fix no-local-version-strict -> fail-on-uncommitted-changes
- [x] Fix RTD config


pckroon:
Modernized docs can be found here: https://vermouth-martinize--772.org.readthedocs.build/en/772/index.html

pyproject.toml
```diff
@@ -1,10 +1,71 @@
...
+
+[project.optional-dependencies]
+mdtraj = ["mdtraj"]
+PyCifRW = ["PyCifRW"]
+
+[project.scripts]
+martinize2 = "vermouth.bin.martinize2:entry"
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
```
csbrasnett: What's the advantage of listing everything in a requirements.txt vs. listing dependencies directly here? Also if we're using this, might it make sense to list the optional dependencies similarly?

csbrasnett (commented):
⬆️ things we love to see

Just one small question about how we're organising dependencies. I don't have particularly strong feelings, just would like to clarify please!

pyproject.toml
pckroon: Excellent question. I personally like a separate requirements.txt (because `pip install -r requirements.txt`), but I'm not sure it makes sense. That's why I went for this middle ground.
Yesterday I also looked at including the extra dependencies (`vermouth[tests, docs]`) from their respective files, but I ran into the issue that either *all* extra dependencies come from files (including cifrw and mdtraj), *or* none.
I'll have a look around whether a separate requirements.txt still makes sense.

csbrasnett (approved):
nice wrapping up, merge away

PR #773, opened 2026-05-26
Files created with martinize should follow umask

### LINKED ISSUE #771: Files created with Martinize do not follow shell umask
I realize this was already discussed in #762, I'm opening a new issue because it is already closed, and because I'm commenting on the issue, rather than the submitted patch.

I also see files created by martinize have only owner read+write permissions, despite every other file in the same folder typically having owner read+write and group read, everyone read permissions.

I would like to add to the previous discussion, that methods in Python's tempfile, including `tempfile.mkstemp` seem to create files with more restrictive permissions than the umask of the shell environment (my umask is 0022, but it created a file with `-rw-------` permissions), and I believe this is intentional behavior in tempfile. `shutil.move` then preserves permissions when it moves the temp file to the final path.

While I am not entirely familiar with the codebase and I have not tested it in depth, maybe _open_tmp_file could be changed to delete the temporary file, and create a new one from scratch within the same path, but with open()?

aster31:
Make a temporary file using open() to create it with default permissions. Copy permissions for all files from this source.

Based on discussion in #771, fixes #771 and #762

### DIFF
--- a/vermouth/file_writer.py
+++ b/vermouth/file_writer.py
@@ -171,8 +171,17 @@ class DeferredFileWriter(metaclass=Singleton):
                 LOGGER.info('Backing up {} to {}.', final_path, free_path, type='general')
                 shutil.move(str(final_path), str(free_path))
             LOGGER.debug('Writing output to {}.', final_path, type='general')
-            shutil.copyfile(tmp_path, str(final_path))
-            os.remove(tmp_path)
+            try:
+                success = True
+                shutil.copyfile(tmp_path, str(final_path))
+            except Exception as error:
+                success = False
+                LOGGER.exception("An error occurred when moving your output from {tmp_path} to {final_path}: {error}", tmp_path=tmp_path, final_path=final_path, error=error)
+            if success:
+                try:
+                    os.remove(tmp_path)
+                except OSError as error:
+                    LOGGER.warning("An error occurred removing temporary file {tmp_path}: {error}", tmp_path=tmp_path, error=error)
 
     @staticmethod
     def _append_file(tmp_path, final_path, mode='a'):
--- a/vermouth/file_writer.py
+++ b/vermouth/file_writer.py
@@ -150,10 +150,17 @@ class DeferredFileWriter(metaclass=Singleton):
         Existing file destinations will be backed up according to the Gromacs
         scheme.
         """
+        # get a temp path
+        handle, perm_source = tempfile.mkstemp()
+        os.close(handle)
+        os.remove(perm_source)
+        # create an empty file with default permissions based on env
+        open(perm_source, "x").close()
+
         while self.open_files:
             tmp_path, final_path, mode = self.open_files.popleft()
             if 'w' in mode or '+' in mode:  # write
-                self._write_file(tmp_path, final_path)
+                self._write_file(tmp_path, final_path, perm_source)
             elif 'r' in mode:  # read = error
                 raise AssertionError("Files opened with mode 'r' should not be "
                                      "treated special")
@@ -162,7 +169,9 @@ class DeferredFileWriter(metaclass=Singleton):
             else:
                 raise KeyError('Unknown file mode')
 
-    def _write_file(self, tmp_path, final_path):
+        os.remove(perm_source)
+
+    def _write_file(self, tmp_path, final_path, perm_source):
         # There is no way to move a file and make it error if the destination
         # already exists, so use a lock instead.
         with lock:
@@ -172,6 +181,7 @@ class DeferredFileWriter(metaclass=Singleton):
                 shutil.move(str(final_path), str(free_path))
             LOGGER.debug('Writing output to {}.', final_path, type='general')
             shutil.move(tmp_path, str(final_path))
+            shutil.copymode(perm_source, str(final_path))
 
     @staticmethod
     def _append_file(tmp_path, final_path, mode='a'):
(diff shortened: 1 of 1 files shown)


vermouth/file_writer.py
```diff
@@ -150,10 +150,17 @@ def write(self):
         Existing file destinations will be backed up according to the Gromacs
         scheme.
         """
+        # get a temp path
+        handle, perm_source = tempfile.mkstemp()
+        os.close(handle)
+        os.remove(perm_source)
```
pckroon:
```suggestion
    perm_source = '.VERMOUTH_TEMPFILE'
```

I honestly like your solution a little better, *but* maybe it's important to create the file in the same directory?
Using `tempfile.gettempprefix` may still be worth considering.

vermouth/file_writer.py
```diff
@@ -162,7 +169,9 @@ def write(self):
             else:
                 raise KeyError('Unknown file mode')
 
-    def _write_file(self, tmp_path, final_path):
+        os.remove(perm_source)
```
pckroon: This needs to be wrapped in a try/except. Or does `os.remove` not error if the file cannot be removed?

vermouth/file_writer.py
```diff
@@ -172,6 +181,7 @@ def _write_file(self, tmp_path, final_path):
                 shutil.move(str(final_path), str(free_path))
             LOGGER.debug('Writing output to {}.', final_path, type='general')
             shutil.move(tmp_path, str(final_path))
+            shutil.copymode(perm_source, str(final_path))
```
pckroon: And this needs a try/except as well. Just in case.

pckroon (changes requested):
Very nice work. Some small comments to capture weird edge cases

vermouth/file_writer.py
aster31: I'm not sure if it's guaranteed that the working directory is writable, alternatively perhaps it's possible to create it in the directory of one of the output files, using the directory argument of mkstemp?

vermouth/file_writer.py
aster31: It does error if the file does not exist, but I don't think silently failing is a good idea. It would only fail if there is a second thread/process deleting the file in the short time window when write() happens, and crashing seems preferable to me in that scenario.

Of course I can add a try/expect if you insist.

vermouth/file_writer.py
pckroon: True... 
New terrible idea: for each file-to-write, do this trick in the directory the file will be written. This is getting a bit silly though. If those folders are not writable you're going to have an issue anyway. Why all this effort: I have no idea whether file attributes can be statically inherited from folders. 

Alternative approach: Instead of creating a temp file for each file, create one tempdir per run in which we write regular files, and move everything in place at the end from there. It would need some checking whether those files would then respect the umask.

vermouth/file_writer.py
pckroon: Or if the file doesn't exist on disk yet and therefor cannot be unlinked? I agree that it's super rare. It's a somewhat silly reason to crash though, especially if the rest of the process worked fine. Maybe issue a warning that we're littering on the file system if that happens

vermouth/file_writer.py
aster31: In my testing, only tempfile.mkstemp and similar methods create files not following umask. If files are created using open() they follow umask, and with os.open, the umask is applied on top of the provided mode.

If opening temporary files with open() is acceptable, it could happen right at the target directory for each file, which might be the simplest (no tempdir management), and would mean that shutil.move is guaranteed to move rather than copy+delete (which happens when files are on different filesystems).

pckroon: > If opening temporary files with open() is acceptable

What do you mean? I'm listening :)

The main reason I implemented this in the first place was to prevent users from running simulations with topologies that caused warnings, similar to how grompp works.
Another option would be to always do copy+delete. Use `open` to create an empty file, and copy file contents from the temporary file using something from shutil (iirc)

aster31:
> Another option would be to always do copy+delete. Use `open` to create an empty file, and copy file contents from the temporary file using something from shutil (iirc)

You're right, this makes it possible to simplify the change into a much nicer/simpler one.

If you'd like I can try squashing the two commits on my fork into one, as it then creates a tiny three line diff

vermouth/file_writer.py
```diff
@@ -171,7 +171,8 @@ def _write_file(self, tmp_path, final_path):
                 LOGGER.info('Backing up {} to {}.', final_path, free_path, type='general')
                 shutil.move(str(final_path), str(free_path))
             LOGGER.debug('Writing output to {}.', final_path, type='general')
-            shutil.move(tmp_path, str(final_path))
+            shutil.copyfile(tmp_path, str(final_path))
+            os.remove(tmp_path)
```
pckroon: Very clean! Since `copyfile` is not atomic I do suggest the following addition:


```suggestion
            try:
                success = True
                shutil.copyfile(tmp_path, str(final_path))
            except Exception as error:
                success = False
                LOGGER.exception("An error occurred when moving your output from {tmp_path} to {final_path}: {error}", tmp_path=tmp_path, final_path=final_path, error=error)
            if success:
                try:
                    os.remove(tmp_path)
                except OSError as error:
                    LOGGER.warning("An error occurred removing temporary file {tmp_path}: {error}", tmp_path=tmp_path, error=error)
```
Something along those lines anyway...

pckroon (changes requested):

pckroon (approved):
Love it.
Many thanks for investigating and fixing this!

PR #774, opened 2026-05-29
Warning for id-regions

carolinasarto:
According to #770, I'm including a warning that will tell users when they are trying to use parameters for an IDP/IDR region that it's not present in the input pdb.

### DIFF
--- a/vermouth/bin/martinize2.py
+++ b/vermouth/bin/martinize2.py
@@ -1026,20 +1026,50 @@ def entry():
         disordered_regions=args.id_regions
     )
 
-    print("DEBUG id_regions:", args.id_regions)
+    # Check that all requested residues to be treated as IDR are actually present in the structure, and warn if not.
     if any(args.id_regions):
-        annotated = any(
-            mol.nodes[n].get("cgidr")
+
+        # Rebuild requested_resids from args.id_regions
+        requested_resids = set()
+        for region in args.id_regions:
+            if ":" in region:
+                a, b = region.split(":")
+                start, end = int(a), int(b)
+            else:
+                start = end = int(region)
+            lo, hi = sorted((start, end))
+            requested_resids.update(range(lo, hi + 1))
+
+        # Collect annotated residues using TRUE PDB numbering
+        annotated_resids = {
+            mol.nodes[n]['stash']['resid']
             for mol in system.molecules
             for n in mol.nodes
-        )
-        if not annotated:
+            if mol.nodes[n].get("cgidr")
+        }
+
+        # Collect ALL PDB residues present in the structure
+        present_resids = {
+            mol.nodes[n]['stash']['resid']
+            for mol in system.molecules
+            for n in mol.nodes
+        }
+        min_pdb = min(present_resids)
+        max_pdb = max(present_resids)
+
+        # Find missing requested residues
+        missing = sorted(requested_resids - annotated_resids)
+
+        if missing:
+            # show only first 10 missing residues
+            missing_preview = missing[:10]
+
             LOGGER.warning(
-                "No residues were annotated as IDR. "
-                "Check that the resids in -id-regions match your input structure.",
+                f"Requested residues in -id-regions do not exist in the input structure. "
+                f"Some missing residues: {missing_preview} ... "
+                f"Please select residues within the PDB range {min_pdb}–{max_pdb}.",
                 type="missing-flag",
             )
-        
 
     if 'bondedtypes' in known_force_fields[args.to_ff].variables:
         LOGGER.info("Generating implicit interactions for RTP force field", type='step')
--- a/vermouth/bin/martinize2.py
+++ b/vermouth/bin/martinize2.py
@@ -1027,21 +1027,37 @@ def entry():
     )
 
     # Check if any residues were annotated as IDR, if IDR regions were specified.
+    # if any(args.id_regions):
+    #     for region in args.id_regions:
+    #         # check if any molecule node was actually annotated
+    #         annotated = any(
+    #             mol.nodes[n].get("idr")
+    #             for mol in system.molecules
+    #             for n in mol.nodes
+    #         )
+    #         if not annotated:
+    #             LOGGER.warning(
+    #                 "No residues were annotated as IDR. "
+    #                 "Check that the resids in -id-regions match your input structure.",
+    #                 type="missing-feature",
+    #             )
+    #             break
+    
+
+
     if any(args.id_regions):
-        for region in args.id_regions:
-            # check if any molecule node was actually annotated
-            annotated = any(
-                mol.nodes[n].get("idr")
-                for mol in system.molecules
-                for n in mol.nodes
+        annotated = any(
+            mol.nodes[n].get("idr")
+            for mol in system.molecules
+            for n in mol.nodes
+        )
+        if not annotated:
+            LOGGER.warning(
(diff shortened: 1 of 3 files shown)


vermouth/bin/martinize2.py
```diff
@@ -1025,6 +1025,52 @@ def entry():
...
+        if missing:
+            # show only first 10 missing residues
+            missing_preview = missing[:10]
+
+            LOGGER.warning(
+                f"Requested residues in -id-regions do not exist in the input structure. "
+                f"Some missing residues: {missing_preview} ... "
+                f"Please select residues within the PDB range {min_pdb}–{max_pdb}.",
+                type="missing-flag",
+            )
```
pckroon: Could you also log how many are missing?
Plus, I think it's better to let the logger deal with the string formattring rather than using f-strings. That way the (expensive) formatting is only done if actually needed.

```suggestion
            LOGGER.warning(
                "Requested residues in -id-regions do not exist in the input structure. "
                "Some missing residues: {missing_preview} ... "
                "Please select residues within the PDB range {min_pdb}–{max_pdb}.",
                type="missing-flag", missing_preview=missing_preview, min_pdb=min_pdb, max_pdb=max_pdb
            )
```

pckroon (changes requested):
Great functionality, thanks for implementing!
However, I think it makes more sense to move this to the annotateIDR processor, rather than adding it to bin/martinize2.
Other than that, small comment about the logging

csbrasnett:
Just to jump in, I agree it would be better if this lived in the processor directly. @carolinasarto, we spent a while getting a similar error in the AnnotateMutMod processor nicely done, it might be useful to look there as a reference. Thanks for putting the PR together, will be super useful!

csbrasnett:
@carolinasarto I think this is heading in the right direction, thanks for pushing forward with this feature. The latest commit looks good, but I think needs to be fully moved into the annotate_idrs processor rather than live in the executable please. If you could then add some tests to make sure we're catching the right things. [This](https://github.com/marrink-lab/vermouth-martinize/blob/d0ec4df926026108b1fa8c479f7f646a8aaa3bc7/vermouth/tests/test_annotate_mut_mod.py#L457) test for the annotate_mut_mod processor might be a good place to start as a comparison

carolinasarto:
Thanks for the feedback! I've moved the warning to the annotate_idrs processor, and I'm now logging how many residues are missing. I also switched to the same formatting style used in the AnnotateMutMod processor, without f-strings. I added tests following the test_mod_resid_not_correct pattern you suggested, to make sure we're catching the right things.

Let me know if this works or if more changes are needed.

vermouth/tests/test_annotate_idrs.py
```diff
@@ -172,3 +178,56 @@ def test_parse_disorder_resspec(resspec, expected):
...
+    Tests that a warning is logged when -id-regions requests residues
+    that are not present in the input structure.
+
+    The structure only contains resids 1-4, so requesting 9:11 should
+    trigger exactly one WARNING that mentions how many residues are
+    missing (3).
+    """
+    system = _make_idr_system()
+
+    caplog.clear()
```
pckroon: This should not be necessary

vermouth/tests/test_annotate_idrs.py
```diff
@@ -172,3 +178,56 @@ def test_parse_disorder_resspec(resspec, expected):
...
+    """
+    Build a minimal system with resids 1-4 in chain A.
+    """
+    system = System(force_field=ForceField(FF_UNIVERSAL_TEST))
+    mol = Molecule(force_field=ForceField(FF_UNIVERSAL_TEST))
+    nodes = [
+        {'chain': 'A', 'resid': 1, 'stash': {'resid': 1}},
+        {'chain': 'A', 'resid': 2, 'stash': {'resid': 2}},
+        {'chain': 'A', 'resid': 3, 'stash': {'resid': 3}},
+        {'chain': 'A', 'resid': 4, 'stash': {'resid': 4}},
```
pckroon: I think the test would become stronger if (for some?) the stashed resid != resid

pckroon (dismissed):
Very nice.
The tests could be strengthened a bit I think, but optional IMO. It may also be worth testing what happens if you provide multiple id-regions (`['1:3`, '4']`) maybe.

Optional as far as I'm concerned though, let me know if you're up for that

carolinasarto:
Thanks for your insights! It made me realize I wasn't including chains or the empty-region case, so I made more tests with multiregions, gaps, and chains. 

I hope this is strong enough, but let me know if you can think of other cases. 

I also included a printed message for the case of an empty region, and another for when you select no chain, saying that it will get all chains. 

As a note, I'm not sure what you meant by [‘1’:’3’,’4’] because that would get an error, it should be 4:4 to select only one resid. But I included other multiregion cases. 

By thinking about this and checking how mutations have their own tests and warning, I was wondering if it would be better to have a global warning whenever a user selects any type of region, because there might be other functions that are silently not recognizing what the user is selecting, PDBs have gaps or different numberings. And instead of adding warnings and tests in each module, maybe it's worth it at some point to define a global checking.

vermouth/processors/annotate_idrs.py
```diff
@@ -130,7 +138,15 @@ def _check_missing_idr_residues(system, id_regions):
...
-            type="missing-flag",
+        )
+
+    # Inform the user when a region has no chain specifier, since it will
+    # be applied to all chains in the structure.
+    if any(region.get('chain') is None for region in id_regions):
+        LOGGER.info(
+            "No chain specified in -id-regions. The regions will be applied "
+            "to all chains in the structure.",
+            type="general",
```
pckroon, line 149: Maybe add a check whether there are multiple chains in the system to begin with?

pckroon (approved):
Fantastic <3

We don't usually test whether INFO messages get logged, but also doesn't hurt :)
As far as I'm concerned tests are definitely thorough enough. 
As for unifying the residue/atom finding: sounds like great idea to me. It may also be nice to see if the CLI formatting for both is similar enough. Separate PR though.

vermouth/tests/test_annotate_idrs.py
```diff
@@ -172,3 +178,138 @@ def test_parse_disorder_resspec(resspec, expected):
...
+        {'chain': 'A', 'resid': 1, 'stash': {'resid': 10}},
+        {'chain': 'A', 'resid': 2, 'stash': {'resid': 11}},
+        {'chain': 'A', 'resid': 3, 'stash': {'resid': 14}},
+        {'chain': 'A', 'resid': 4, 'stash': {'resid': 15}},
+        {'chain': 'B', 'resid': 1, 'stash': {'resid': 20}},
+        {'chain': 'B', 'resid': 2, 'stash': {'resid': 21}},
+    ]
+    mol.add_nodes_from(enumerate(nodes))
+    system.add_molecule(mol)
+    return system
```
csbrasnett, line 204: @pckroon I think this is fine for now but at some point might be useful to think about how/where we're constructing molecules/systems for tests, given we already have these functions (and import them here): 

https://github.com/marrink-lab/vermouth-martinize/blob/cfcad6ecc2316cc40180252523fc11b51e987447/vermouth/tests/test_annotate_idrs.py#L21

pckroon: Good idea, but very much low prio as far as I'm concerned

Issue #776, opened 2026-06-08
Parser does not report which file error happens in


aster31:
Bug description: parser_utils.py class **SectionLineParser** method **parse_section** error messages contain a line number, but it does not specify which file it is.

Relevant source code excerpt (parser_utils.py line 241):

```
        except Exception as error:
            raise IOError("Problems parsing line {}. I think it should be a "
                          "'{}' line, but I can't parse it as such."
                          "".format(lineno, self.section)) from error
```

pckroon:
The reason is because `LineParser.parse`  is given a file handle, rather than a file path. If we did our work correctly the filename is somewhere in the traceback, but looking at some of the code that doesn't seem to be the case... For example: https://github.com/marrink-lab/vermouth-martinize/blob/8c7f8f4abe462112fa3945c96152eddb7d061cbd/vermouth/map_parser.py#L908

I don't immediately see a clean way to implement this, but I'm all ears :)

aster31:
Indeed, I ran into this exception while parsing a mapping file.

Unfortunately I can't see a way to implement it cleanly and without changing the function signatures, the simple solution I think would be to change function signatures to pass file paths alongside file handles?

As an additional note, would it be also possible to include the content of the offending line itself in the command line output, and the original exception message as well?

pckroon:
> The simple solution I think would be to change function signatures to pass file paths alongside file handles?

I don't love this option, since not all file handles will have a corresponding file(name). Take for example network streams or compressed archives.
I may be an option to add relevant output to some of the read/parse file functions. For example:
```py
def parse_mapping_file(filepath, force_fields):
    with open(filepath, encoding='utf-8') as map_in:
        director = MappingDirector(force_fields)
        try:
            mappings = list(director.parse(map_in))
        except IOError as error:
            raise IOError(f'Error reading {filepath}') from error
    return mappings
```
Messy and annoying to go through the whole repo though...


> As an additional note, would it be also possible to include the content of the offending line itself in the command line output, and the original exception message as well?

Absolutely. PR welcome :)

aster31:
> I don't love this option, since not all file handles will have a corresponding file(name). Take for example network streams or compressed archives.

When those are created, could an acceptable placeholder string might be passed in place of the filename? E.g. whoever creates a network stream could pass "NetworkStream" or the ip address or something, I'd imagine it's not hard to come up with something in those cases, which could still provide useful insights for error messages during parsing those streams.

[...]

PR #777, opened 2026-06-12
Parser Utils extra exception info

aster31:
Also see #776

This current PR so far was made with the intention to avoid breaking changes, it so far adds whatever info is readily available to parse error messages (I figured this already partly fixes the issue without the breaking change of adding paths to function signatures).

I also changed two regexes to raw strings, since there were warnings printed while using the martinize2 script (I hope this is ok)

### DIFF
--- a/vermouth/citation_parser.py
+++ b/vermouth/citation_parser.py
@@ -140,7 +140,7 @@ class BibTexDirector():
         str, str
             the field type, the field content
         """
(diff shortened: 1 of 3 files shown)


pckroon (approved):
Love it, many thanks.

And yes, I'm very happy with the drive-by mini-fixes!

Issue #782, opened 2026-07-21
martinize2 -ss flag does not work in protein with modified residue CYSP


Aleksandr-biochem:
Hello,

Hope you are well.

I ran into this issue while setting up a martini3001 simulation with a protein that contains CYSP cysteine palmitoylation. I have atomistic coordinates of the protein with a CYSP residue. When running `martinize2` I tried providing secondary structure string with the `-ss` flag, but I get an error `ValueError: There is no molecule to which to apply the sequence.`, so seemingly CYSP residue prevents recongnion of the molecule as a protein. However, it works fine when providing the DSSP executable.

```
Python 3.11.5
System: Ubuntu 24.04.4 LTS
DSSP 2.0.4
martinize with vermouth 0.15.0
```

I attach a small PDB for Trp-Cage protein (PDB 1L2Y) with CYSP residue to reproduce the issue (see file lines at the bottom).  

Here are the commands to reproduce the issue.

Fails:
```
$ martinize2 -f 1L2Y_CYSP.pdb -x protein_cg.pdb -o topol.top -ff martini3001 -elastic -ef 700 -eu 0.9 -ss CHHHHHHHTTGGGGTCCCCC
    INFO - general - Read 1 molecules from PDB file 1L2Y_CYSP.pdb
    INFO - step - Guessing the bonds.
    INFO - general - 1 molecules after guessing bonds
    INFO - step - Repairing the graph.
    INFO - general - Applying modification N-ter to residue P-ASN1
    INFO - general - Applying modification + to residue P-ASN1
    INFO - general - Applying modification C-ter to residue P-SER20
    INFO - general - Applying modification + to residue P-SER20
    INFO - step - Dealing with modifications.
    INFO - general - Identified the modifications ['N-ter'] on residues ['ASN1', 'ASN1', 'ASN1', 'ASN1']
    INFO - general - Identified the modifications ['LYS-HZ3'] on residues ['LYS8']
    INFO - general - Identified the modifications ['C-ter'] on residues ['SER20', 'SER20', 'SER20']
    INFO - step - Read input.
Traceback (most recent call last):
  File ".../Documents/tools/envs/mdenv/bin/martinize2", line 1205, in <module>
    entry()
  File ".../sann7918/Documents/tools/envs/mdenv/bin/martinize2", line 949, in entry
    ).run_system(system)
      ^^^^^^^^^^^^^^^^^^
[...]

PR #783, opened 2026-07-22
Add [meta] category protein to charmm lipidations

Aleksandr-biochem:
As per #782  `charmm/lipidations.rtp` lack `meta` field as in protein residues from `charmm/aminoacids.rtp`. 

This leads to `has_category(molecule.force_field.blocks.get(...), 'protein')` in `is_protein` evaluating to `False`, when running `martinize2`  on proteins with lipidations. This prevents `-ss` flag from functioning properly.

Added `[meta] category protein` to all lipidations in `charmm/lipidations.rtp`  to allow their use with `-ss` flag.


pckroon (approved):
Good find, thanks!

PR #786, opened 2026-08-28
Reduce memory and runtime of Go contact map calculation

fercofas:
The RCSU Go contact map calculation (vermouth/rcsu/contact_map.py) could consume large amounts of RAM and take a long time on sizable protein complexes. Three changes address this without altering the resulting contact map:

- Compute the pairwise atom distance matrix once per molecule and share it between the OV and CSU contact searches, instead of running the same KDTree-based distance query twice at different cutoff radii.

- Rewrite _classify_contact_types to deduplicate and count (i, k) atom contact pairs per atom row with vectorized NumPy/SciPy operations instead of accumulating into Python dicts keyed by tuples, which carried large per-entry overhead for systems with many contacts.

- Build the Fibonacci sphere KDTree used in _calculate_csu_contacts once per molecule instead of once per atom by querying it in each atom's local (centered, unit-radius) frame rather than rebuilding an equivalent tree translated and scaled to that atom.

Validated against three real structures of increasing size, all producing identical Go contact counts to the unmodified code:
  - 6.4k atoms: wall time 17.2s → 10.9s (-36%)
  - 30k atoms: wall time 79.1s → 50.0s (-37%), peak Python memory 614.3MB → 521.5MB (-15%)
  - 36k atoms: wall time 65.8s → 41.2s (-37%)

### DIFF
--- a/vermouth/rcsu/contact_map.py
+++ b/vermouth/rcsu/contact_map.py
@@ -439,12 +439,16 @@ def _extract_contact_inputs(molecule):
     LOGGER.debug("Extracted {} atoms from {} residues", len(positions_all), nresidues)
     return vdw_list, atypes, coords, res_serial, resids, chains, resnames, res_idx, ca_pos, nresidues, G
 
-def _calculate_ov_contacts(coords_tree, vdw_list, natoms, vdw_max, alpha=1.24):
+def _calculate_ov_contacts(coo, vdw_list, natoms, vdw_max, alpha=1.24):
     """
     Find enlarged (OV) overlap contacts
 
-    coords_tree: KDTree
-        KDTree of the input coordinates
+    coo: scipy.sparse.coo_matrix
+        precomputed sparse pairwise-distance matrix (COO format) of the
+        input coordinates, out to at least a radius of 2 * vdw_max * alpha.
+        This is shared with _calculate_csu_contacts so that the expensive
+        KDTree-based distance query is only performed once per molecule
+        instead of once per contact type.
     vdw_list: list
         list of vdw radii of the input coordinates
     natoms: int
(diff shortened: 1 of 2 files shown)


pckroon:
Hi, thanks! I don't have a lot of time available at the moment, with the start of the academic year and all. I'll try to get to this not too far in the future though. 

@csbrasnett Do you have time for a review?

csbrasnett:
@pckroon will have a look at this before end of week!

vermouth/rcsu/contact_map.py
```diff
@@ -627,11 +697,28 @@ def _compute_residue_contacts(vdw_list, atypes, coords, res_serial, nresidues):
...
     coords_tree = KDTree(coords)
 
+    # OV and CSU contacts both start from a pairwise distance query on the
+    # same coordinate set; the only difference is the cutoff radius used
+    # to prune candidate pairs. Computing this once (at the larger of the
+    # two cutoffs) and reusing it avoids running the expensive KDTree
+    # distance query twice, which previously roughly doubled both the
+    # runtime and peak memory of this step for large systems.
+    alpha = 1.24
+    water_radius = 2.80
```
csbrasnett: I think some refactoring of these variables to be instance variables instead would be good. Could be useful if anyone ever wanted to explore in particular modifying the alpha value

vermouth/rcsu/contact_map.py
```diff
@@ -627,11 +697,28 @@ def _compute_residue_contacts(vdw_list, atypes, coords, res_serial, nresidues):
...
+    cutoff_ov = 2 * vdw_max * alpha
+    cutoff_csu = (2 * vdw_max) + water_radius
+    max_cutoff = max(cutoff_ov, cutoff_csu)
```
csbrasnett, line 710: I think some refactoring of these variables to be instance variables instead would be good. Could be useful if anyone ever wanted to explore in particular modifying the alpha value. It's small, but atm cutoff_ov is being calculated both here and in the _calculate_ov_contacts function, and I think for readability it'd be better to define this as a variable accessible across the class than repeatedly calculating it.

vermouth/rcsu/contact_map.py
```diff
@@ -455,12 +459,17 @@ def _calculate_ov_contacts(coords_tree, vdw_list, natoms, vdw_max, alpha=1.24):
...
-    cols = over_coo.col[keep]
+    vdw_sum = alpha * (vdw_list[coo.row] + vdw_list[coo.col])
+    # cutoff_ov is OV's own cutoff radius (2 * vdw_max * alpha). In normal
+    # usage every per-atom vdw radius is <= vdw_max, so vdw_sum can never
+    # exceed cutoff_ov and this comparison is redundant with the one
+    # above. It is kept explicit so that behaviour is independent of
+    # whichever (possibly larger) radius was used to build the shared
+    # `coo` distance matrix passed in from _compute_residue_contacts.
+    cutoff_ov = 2 * vdw_max * alpha
+    keep = (coo.row < coo.col) & (coo.data < vdw_sum) & (coo.data < cutoff_ov)
```
csbrasnett, line 470: I'm not sure the last condition here is necessary if the sdm is already constructed using the < cutoff_ov condition (714-5) ?

csbrasnett (changes requested):
Two comments about helping readability/future testing/optimisation.

My other comment is that Claude has done its favourite thing of adding very long comments about what's changed. I'd strongly prefer to focus on what the code's _actually_ doing rather than what it used to look like. see e.g. 533-544. Where these long comments have been written please could you succinctly write in your own words about what calculation is being done & how, it'd be much better for maintainence. 

Otherwise looks good, thanks for the contribution!

vermouth/rcsu/contact_map.py
fercofas: Fixed. cutoff_ov is now computed once in _compute_residue_contacts and passed into _calculate_ov_contacts as a parameter, so it's no longer recomputed there.

vermouth/rcsu/contact_map.py
fercofas: It's needed in general: coo is built at max(cutoff_ov, cutoff_csu) since it's shared with the CSU calculation, and for typical vdW radii cutoff_csu > cutoff_ov, so coo does extend past cutoff_ov.

vermouth/rcsu/contact_map.py
csbrasnett: ok, this is fixed enough for now. Ideally I think we want this processor refactored so that `alpha` and `water_radius` are exposed (with the current hard coded values as defaults), but IMO that's outside the scope of this PR. I'll open an issue an address it later.

csbrasnett (approved):

pckroon:
Fantastic. Thanks so much! <3

